PR #301, opened 2014-12-30
Introduce IdentityError

HaoK:
As we discussed today, to keep the logical error along with the msg, introducing a new IdentityError class

Code is the logical status code for the error
Description is the friendly error


HaoK:
Merged c9d27e27e6dd6c15b4228de6e5de3c1a22c15a6d

Issue #17, opened 2014-03-20
Need to rehash password if custom PasswordHasher returns PasswordVerificationResult.SuccessRehashNeeded


suhasj:
If the user implements a custom PasswordHasher and returns the PasswordVerificationResult.SuccessRehashNeeded in the 'VerifyHashedPassword' method, we need to rehash the password and store it back into the database. 

 This is not implemented currently, for example. in the Find method for usermanager we verify if the verification code is not 'Failed'. If not we return true. We can check if the code is PasswordVerificationResult.SuccessRehashNeeded and rehash the password

henninga:
Will this only be part of the vNext?

PR #306, opened 2015-01-02
Rehash passwords when needed

HaoK:
Fixes https://github.com/aspnet/Identity/issues/17

cc @divega @GrabYourPitchforks @blowdart @suhasj

### DIFF
--- a/src/Microsoft.AspNet.Identity/UserManager.cs
+++ b/src/Microsoft.AspNet.Identity/UserManager.cs
@@ -455,7 +455,7 @@ namespace Microsoft.AspNet.Identity
             {
                 throw new ArgumentNullException("password");
             }
-            var result = await UpdatePasswordInternal(passwordStore, user, password, cancellationToken);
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Identity/UserManager.cs
```diff
@@ -557,7 +557,12 @@ public virtual string NormalizeUserName(string userName)
             {
                 return false;
             }
-            return await VerifyPasswordAsync(passwordStore, user, password, cancellationToken);
+            var result = await VerifyPasswordAsync(passwordStore, user, password, cancellationToken);
```
suhasj: Is there a way to hook convert this to IdentityResult. That way I insert logging here which can trace that the password was rehashed. Else we can look at logging to log for results other than IdentityResult as well

HaoK:
Merged 69ac9abcc5bdd831026bb31983b9e4f2979915e5

PR #315, opened 2015-01-07
Add concurrency support for EF Store

HaoK:


HaoK:
cc @divega @suhasj

src/Microsoft.AspNet.Identity/IdentityErrorDescriber.cs
```diff
@@ -15,6 +15,15 @@ public virtual IdentityError DefaultError()
                 Description = Resources.DefaultError
             };
         }
+        public virtual IdentityError ConcurrencyFailure()
+        {
+            return new IdentityError
+            {
+                Code = nameof(ConcurrencyFailure),
```
divega: Not closely related to this checkin, but would CallerMemberNameAttribute in a parameter constructor of IdentityError be nicer way to achieve this?

divega:
Are we doing this only for User? Other than that, :shipit:

divega:
Actually i missed that it was on Role as well. Just :shipit:

src/Microsoft.AspNet.Identity/IdentityErrorDescriber.cs
HaoK: Not exactly, they are meant to be logical strings, as opposed to caller, i.e. Logging could print out the error code, instead of the actual error string if desired.  Without the code developers would have to compare against the error strings itself which was a big complaint in the 2.0 stack with identityResult only having an IEnumerable of error strings

HaoK: Oh I see what you are saying, you just meant as a code cleanup to use the attribute instead of nameof?  I don't particularly like that bit of magic as it doesn't even save characters in this situation

divega: I meant that, but only if we end up doing this nameof() thing in most places where we create an IdentityError, in which case I think it would be less error prone.

HaoK:
Merged a8c872aba97e641d38ba2cc098a4ac191a666b6a

Issue #187, opened 2014-08-15
Role and Email normalization


HaoK:

HaoK:
Going to rename IUserNameNormalizer to be IIdentityKeyNormalizer since we will be using this for role name, user name, and email normaliation

HaoK:
We settled on ILookupNormalizer

PR #320, opened 2015-01-09
Normalize all keys (user/role name, email)

HaoK:
Fixes https://github.com/aspnet/Identity/issues/187

cc @divega @suhasj

### DIFF
--- a/src/Microsoft.AspNet.Identity.EntityFramework/UserStore.cs
+++ b/src/Microsoft.AspNet.Identity.EntityFramework/UserStore.cs
@@ -157,7 +157,7 @@ namespace Microsoft.AspNet.Identity.EntityFramework
                 throw new ArgumentNullException("user");
             }
 
-            //Context.Attach(user);
+            Context.Attach(user);
             user.ConcurrencyStamp = Guid.NewGuid().ToString();
             Context.Update(user);
             try
--- a/src/Microsoft.AspNet.Identity.EntityFramework/UserStore.cs
+++ b/src/Microsoft.AspNet.Identity.EntityFramework/UserStore.cs
@@ -961,7 +961,7 @@ namespace Microsoft.AspNet.Identity.EntityFramework
                         && userclaims.ClaimType == claim.Type
                         select user;
 
-            return (IList<TUser>)await query.ToListAsync(cancellationToken);
+            return await query.ToListAsync(cancellationToken);
         }
 
         /// <summary>
@@ -988,7 +988,7 @@ namespace Microsoft.AspNet.Identity.EntityFramework
                             where userrole.RoleId.Equals(role.Id)
                             select user;
 
-                return (IList<TUser>) await query.ToListAsync(cancellationToken);
+                return await query.ToListAsync(cancellationToken);
             }
             return new List<TUser>();
         }
--- a/src/Microsoft.AspNet.Identity.EntityFramework/UserStore.cs
+++ b/src/Microsoft.AspNet.Identity.EntityFramework/UserStore.cs
@@ -78,7 +78,7 @@ namespace Microsoft.AspNet.Identity.EntityFramework
             return AutoSaveChanges ? Context.SaveChangesAsync(cancellationToken) : Task.FromResult(0);
         }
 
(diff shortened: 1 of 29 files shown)


src/Microsoft.AspNet.Identity.EntityFramework/UserStore.cs
```diff
@@ -940,7 +961,7 @@ public virtual Task<bool> GetTwoFactorEnabledAsync(TUser user, CancellationToken
                         && userclaims.ClaimType == claim.Type
                         select user;
 
-            return (IList<TUser>)await query.ToListAsync(cancellationToken);
+            return await query.ToListAsync(cancellationToken);
```
suhasj: Wasn't there a conversion issue here ? I mean from List<T> to IList<T>

divega: AFAIK the conversion issue is only between Task<List<T>> and Task<IList<T>>. Once the List<T> is unwrapped there shouldn't be a problem.

HaoK: yeah I just reflexively clean up grey'd out VS code that is redundant.  Compiler is perfectly happy without the cast

src/Microsoft.AspNet.Identity/IIdentityKeyNormalizer.cs
```diff
@@ -4,15 +4,15 @@
 namespace Microsoft.AspNet.Identity
 {
     /// <summary>
-    ///     Used to normalize a user name
+    ///     Used to normalize keys for consistent lookups
     /// </summary>
-    public interface IUserNameNormalizer
+    public interface IIdentityKeyNormalizer
```
divega: Using just "key" to describe this concept seems to introduce some ambiguity, because there are already primary keys and foreign keys in the entities in addition to the properties we are normalizing. I like "lookup key" more, e.g. `IIdentityLookupKeyNormalizer`.

src/Microsoft.AspNet.Identity/RoleManager.cs
```diff
@@ -32,6 +33,7 @@ public class RoleManager<TRole> : IDisposable where TRole : class
                 throw new ArgumentNullException("store");
             }
             Store = store;
+            KeyNormalizer = keyNormalizer ?? new UpperInvariantKeyNormalizer();
```
divega: Can you please ping the engineering team about this pattern? Not blocking on it.

divega:
:shipit:

src/Microsoft.AspNet.Identity/IIdentityKeyNormalizer.cs
HaoK: How about even more generic, IStringNormalizer since we are basically just using it to normalize strings when we want?

divega: Not bad, but we do it with a specific purpose... What about `IStringLookupNormalizer`? :smile:

HaoK: How about we just drop string and key then, and just do IIdentityLookupNormalizer?

HaoK: Or just ILookupNormalizer

divega: Sounds good (either, I am inclined to keep Identity in the name though, assuming we do it for other interfaces).

HaoK: We only have one other interface with identity and mostly do not, I will file a bug to rationalize, I think we should drop it from IIdentityMessageProvider.

HaoK:
Merged b59440d95ff01994a077be806046151c64197d6c

src/Microsoft.AspNet.Identity.EntityFramework/UserStore.cs
suhasj: ahh true. thanks for clearing that up

PR #327, opened 2015-01-14
Add logging

suhasj:
@HaoK @divega 
PR for adding logging

### DIFF
--- a/src/Microsoft.AspNet.Identity/IdentityErrorDescriber.cs
+++ b/src/Microsoft.AspNet.Identity/IdentityErrorDescriber.cs
@@ -185,5 +185,14 @@ namespace Microsoft.AspNet.Identity
                 Description = Resources.PasswordRequiresUpper
             };
         }
+
+        public virtual IdentityError IncorrectPassword()
+        {
+            return new IdentityError
+            {
+                Code = nameof(IncorrectPassword),
+                Description = Resources.IncorrectPassword
+            };
+        }
     }
 }
--- a/src/Microsoft.AspNet.Identity/IdentityErrorDescriber.cs
+++ b/src/Microsoft.AspNet.Identity/IdentityErrorDescriber.cs
@@ -185,14 +185,5 @@ namespace Microsoft.AspNet.Identity
                 Description = Resources.PasswordRequiresUpper
             };
         }
-
-        public virtual IdentityError IncorrectPassword()
-        {
-            return new IdentityError
-            {
-                Code = nameof(IncorrectPassword),
-                Description = Resources.IncorrectPassword
-            };
-        }
     }
 }
--- a/src/Microsoft.AspNet.Identity/Resources.resx
+++ b/src/Microsoft.AspNet.Identity/Resources.resx
@@ -309,15 +309,15 @@
     <value>{0} : Success</value>
     <comment>Logging method execution success</comment>
   </data>
-  <data name="LoggingSigninStatus" xml:space="preserve">
-    <value>{0} for user: {1} : Result : {2}</value>
+  <data name="LoggingSigninResult" xml:space="preserve">
+    <value>{0} : Result : {1}</value>
     <comment>Logging statement for SignInManager</comment>
   </data>
-  <data name="LoggingIdentityResultMessage" xml:space="preserve">
+  <data name="LoggingResultMessage" xml:space="preserve">
     <value>{0} for user: {1}</value>
     <comment>Message prefix for Identity result</comment>
   </data>
-  <data name="LoggingIdentityResultMessageForRole" xml:space="preserve">
+  <data name="LoggingResultMessageForRole" xml:space="preserve">
     <value>{0} for role: {1}</value>
     <comment>Message prefix for Identity result for role operation</comment>
   </data>
--- a/src/Microsoft.AspNet.Identity/Resources.resx
+++ b/src/Microsoft.AspNet.Identity/Resources.resx
@@ -321,8 +321,4 @@
     <value>{0} for role: {1}</value>
     <comment>Message prefix for Identity result for role operation</comment>
   </data>
-  <data name="IncorrectPassword" xml:space="preserve">
-    <value>Given password cannot be verified</value>
-    <comment>Error when password cannot be verified for user</comment>
-  </data>
 </root>
--- a/src/Microsoft.AspNet.Identity/Resources.resx
+++ b/src/Microsoft.AspNet.Identity/Resources.resx
@@ -130,7 +130,7 @@
     <comment>Default name for the email token provider</comment>
   </data>
   <data name="DefaultEmailTokenProviderSubject" xml:space="preserve">
-    <value>Security Code</value>
+    <value>Security code</value>
     <comment>Default subject for the email</comment>
   </data>
   <data name="DefaultError" xml:space="preserve">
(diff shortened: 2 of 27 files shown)


src/Microsoft.AspNet.Identity/SignInManager.cs
```diff
@@ -403,6 +411,59 @@ public AuthenticationProperties ConfigureExternalAuthenticationProperties(string
...
+        /// <summary>
+        ///     Log SignInStatus for user and return SignInStatus
+        /// </summary>
+        /// <param name="status"></param>
+        /// <param name="user"></param>
+        /// <param name="methodName"></param>
+        /// <returns></returns>
+        protected async virtual Task<SignInResult> LogResultAsync(SignInResult status, TUser user, [System.Runtime.CompilerServices.CallerMemberName] string methodName = "")
+        {
+            if (status.IsLockedOut)
```
divega: Would it be better to move some of this functionality to a Log method in SignInResult similar to how we do it with IdentityResult?

divega:
@suhasj Could you paste here some samples of log output so that we can get feedback?

Also, it seems we are now doing some extra formatting strings and making lots of "fake async" calls to prepare mesages for the log write methods even if nobody is listening. It is not clear to me how to avoid that given the current Logging API. It seems like something we need to follow up on. Can you help with that?

Besides these comments, looks :shipit: to me.

src/Microsoft.AspNet.Identity/Resources.resx
```diff
@@ -291,14 +291,38 @@
...
+  <data name="LoggingIdentityResultMessage" xml:space="preserve">
+    <value>{0} for user: {1}</value>
+    <comment>Message prefix for Identity result</comment>
+  </data>
+  <data name="LoggingIdentityResultMessageForRole" xml:space="preserve">
+    <value>{0} for role: {1}</value>
+    <comment>Message prefix for Identity result for role operation</comment>
+  </data>
+  <data name="IncorrectPassword" xml:space="preserve">
+    <value>Given password cannot be verified</value>
```
HaoK: nit: Maybe this should be "Password is incorrect." to match the code?

src/Microsoft.AspNet.Identity/SignInManager.cs
```diff
@@ -166,10 +173,11 @@ private Task ResetLockout(TUser user, CancellationToken token)
                 await UserManager.AccessFailedAsync(user, cancellationToken);
```
HaoK: So in general it looks like we are ignoring any potential mid method failures, like this AccessFailedAsync call actually could fail when updating the user (especially now that we have concurrency built in).  SignInManager currently ignores the failure (this behavior is debateable), but we probably should at least consider logging these results if they are failures, even if we don't fail the higher level sign in method...

src/Microsoft.AspNet.Identity/SignInManager.cs
HaoK: +1, follow same pattern as IdentityResult

src/Microsoft.AspNet.Identity/UserManager.cs
```diff
@@ -568,9 +593,18 @@ public virtual string NormalizeKey(string key)
...
-                await UpdateAsync(user, cancellationToken);
+                await UpdateUserAsync(user, cancellationToken);
+            }
+            if (result != PasswordVerificationResult.Failed)
+            {
+                await LogResultAsync(IdentityResult.Success, user);
+            }
+            else
+            {
+                await LogResultAsync(IdentityResult.Failed(ErrorDescriber.IncorrectPassword()), user);
```
HaoK: I think you should have an overload of LogResult that takes a bool and just prints out, instead of introducing a new resource that isn't used except for logging

CheckPasswordAsync for user: whatever : True | False

src/Microsoft.AspNet.Identity/UserManager.cs
```diff
@@ -759,7 +793,10 @@ private IUserSecurityStampStore<TUser> GetSecurityStore()
             CancellationToken cancellationToken = default(CancellationToken))
         {
             ThrowIfDisposed();
-            return await GenerateUserTokenAsync(user, Options.PasswordResetTokenProvider, "ResetPassword", cancellationToken);
+            var token = await GenerateUserTokenAsync(user, Options.PasswordResetTokenProvider, "ResetPassword", cancellationToken);
+            await LogResultAsync(IdentityResult.Success, user);
```
HaoK: Perhaps another oveload of LogResult which takes no success code/bool, and just log:

Invoked MethodName with user blah

HaoK: Actually looks like there should be the following log overloads potentially

Log(bool, ...)
Log(...)
Log(IdentityResult, ...)

rather than LogResult

src/Microsoft.AspNet.Identity/SignInManager.cs
suhasj: sure I'll move it

src/Microsoft.AspNet.Identity/SignInManager.cs
suhasj: you mean log intermediate method calls as well. Sure I think it is a good approach if the method call returns IdentityResult. I can do a pass

src/Microsoft.AspNet.Identity/UserManager.cs
suhasj: I thought this one where we trace entering and existing a method was to be done as a separate thing across the framework components. Do we want to have that separately here ?

src/Microsoft.AspNet.Identity/UserManager.cs
suhasj: Fair enough

src/Microsoft.AspNet.Identity/Resources.resx
suhasj: referring to your feedback below, I might remove this

suhasj:
@divega The output format of the log depends on the ILogger objects that are registered via DI. Each of them format the output log as expected. I have tested this with Console logger and will verify with SeriLog before I check in.

Example, For console log
`[INFORMATION:UserManager] UpdateAsync for user df34f-fe334... : Success`

We are formatting the messages as needed which was what we decided. We can always have the check `Logger.IsEnabled` and then only log the message. This prevents the strings to be given to the logger if not needed. Let me know if I can include it. I'll sync up with you offline too

src/Microsoft.AspNet.Identity/UserManager.cs
suhasj: https://github.com/aspnet/Identity/issues/328 tracking this

divega:
> We can always have the check on  Logger.IsEnabled  and then only log the message. 

I  think we could do that and/or if possible also check if the level of the log entry we are about to write is being asked for. But it feels to me like it would be much better if we could push that to the Logging API. It looks like doing structured logging allows deferring the creation of the message, i.e. we pass an ILoggerStructure to the write method and logging will invoke the ILoggerStructure.Message later. I wonder whether we should use that or if there should be more ad-hoc way of doing it with a `Func<string>`.

In my opinion we can check-in without that and follow up later with an improvements.

suhasj:
@divega @HaoK updated PR with review comments

suhasj:
@divega I'll create a bug to follow up on the things you have mentioned

src/Microsoft.AspNet.Identity/IdentityErrorDescriber.cs
```diff
@@ -185,5 +185,14 @@ public virtual IdentityError PasswordRequiresUpper()
                 Description = Resources.PasswordRequiresUpper
             };
         }
+
+        public virtual IdentityError IncorrectPassword()
```
suhasj: I'll remove this

suhasj:
Sample output with console logger hooked in 

error

Look for `UserManager` and `SignInManager` for Identity logs

suhasj:
Merged and passing on CI

PR #330, opened 2015-01-16
Making file lock static

suhasj:
This is to fix the CI which is failing apparently because of tests running in multiple threads trying to access the same log files.


divega:
:shipit:

suhasj:
Thank you @divega

suhasj:
merged

PR #331, opened 2015-01-16
Using file lock when reading file

suhasj:
@divega I had missed out on using file lock when reading from the file


suhasj:
@HaoK

suhasj:
Description: The functional tests have a file logger that logs messages to a file and reads from it in the assert statements. I had used `lock` while writing but not while reading causing the tests to fail. I verified that tests run fine after these changes

pranavkm:
:shipit:

suhasj:
Merge thanks @pranavkm

PR #333, opened 2015-01-17
Creating a file for each test

suhasj:
@divega as you suggested creating individual log files for each test so that it removes the flakiness. Lock around file read and write protects additionally


HaoK:
Ugh, so this was probably my bad for not looking at the test code closer on the PR.  Can we not do an in memory logger instead of writing to files?  Rather than writing the log to disk, just store it in a string.  We definitely should not be writing to disk for these tests unless we really have to.  I'll file an issue

PR #335, opened 2015-01-18
Swtiched logging from file to inmemory

suhasj:
@HaoK @divega 

I switched from logging the messages in memory from writing it to a file since the this was causing the tests to be flaky

### DIFF
--- a/test/Shared/IdentityResultAssert.cs
+++ b/test/Shared/IdentityResultAssert.cs
@@ -59,12 +59,11 @@ namespace Microsoft.AspNet.Identity.Test
         }
         private static void VerifySuccessLog(ILogger logger, string className, string methodName, string id, string userOrRole = "user")
         {
-            if (logger is TestLogger)
+            TestLogger testlogger;
+            if ((testlogger = logger as TestLogger) != null)
             {
-                var fileLogger = logger as TestLogger;
                 string expected = string.Format("{0} for {1}: {2} : Success", methodName, userOrRole, id);
-
-                Assert.True(fileLogger.LogMessages.Contains(expected));
+                Assert.True(testlogger.LogMessages.Contains(expected));
             }
             else
             {
@@ -74,10 +73,10 @@ namespace Microsoft.AspNet.Identity.Test
 
         public static void VerifyLogMessage(ILogger logger, string expectedLog)
         {
-            if (logger is TestLogger)
+            TestLogger testlogger;
+            if ((testlogger = logger as TestLogger) != null)
             {
-                var fileLogger = logger as TestLogger;
-                Assert.True(fileLogger.LogMessages.Contains(expectedLog));
+                Assert.True(testlogger.LogMessages.Contains(expectedLog));
             }
             else
             {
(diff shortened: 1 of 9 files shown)


HaoK:
You can get rid of the file lock entirely now no?  Each test should be getting it's own logger so there is no contention

suhasj:
I read that List is not thread safe. Would it happen that during the tests there might be read and write simultaneously. This was the reason why it was failing for files.

HaoK:
So each test manager should be getting it's own logger Instance if that is not the case already you should fix that

On Jan 18, 2015, at 2:47 PM, "Suhas Joshi" <notifications@github.com<mailto:notifications@github.com>> wrote:


I read that List is not thread safe. Would it happen that during the tests there might be read and write simultaneously. This was the reason why it was failing for files.

-

suhasj:
oh yes it is getting it's own ILogger instance. I'll remove the lock and check the result. Hopefully it works fine with the in memory implementation

suhasj:
@HaoK  Removed the lock property and it tests are passing on my machine. Tried multiple times

test/Shared/IdentityResultAssert.cs
```diff
@@ -60,15 +59,12 @@ public static void VerifyRoleManagerSuccessLog(ILogger logger, string methodName
         }
         private static void VerifySuccessLog(ILogger logger, string className, string methodName, string id, string userOrRole = "user")
         {
-            if (logger is TestFileLogger)
+            if (logger is TestLogger)
```
HaoK: nit: combine as and is check so you can just test against null

test/Shared/IdentityResultAssert.cs
```diff
@@ -78,13 +74,10 @@ private static void VerifySuccessLog(ILogger logger, string className, string me
 
         public static void VerifyLogMessage(ILogger logger, string expectedLog)
         {
-            if (logger is TestFileLogger)
+            if (logger is TestLogger)
```
HaoK: same thing combine perhaps rename var to testLogger too

test/Shared/TestLogger.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Collections.Generic;
+using Microsoft.Framework.Logging;
+
+namespace Microsoft.AspNet.Identity.Test
+{
+    public class TestLogger : ILogger
+    {
+        public IList<string> LogMessages { get; private set; } = new List<string>();
+
+        public TestLogger(string name)
```
HaoK: name is unused.  remove

suhasj:
@HaoK updated with feedback

test/Shared/IdentityResultAssert.cs
```diff
@@ -60,15 +59,11 @@ public static void VerifyRoleManagerSuccessLog(ILogger logger, string methodName
         }
         private static void VerifySuccessLog(ILogger logger, string className, string methodName, string id, string userOrRole = "user")
         {
-            if (logger is TestFileLogger)
+            TestLogger testlogger;
```
HaoK: its more readable putting the as in the declaration rather than the if

HaoK:
:shipit: after you move the as's out of the ifs

suhasj:
sure thanks @HaoK

Issue #349, opened 2015-02-05
ExternalLoginSignInAsync be virtual method


jleung-adxstudio:
Can SignInManger.ExternalLoginSignInAsync be made virtual like the other SignIn methods? Even better would be able to override the PreSignInCheck to do custom user lockout.

PR #353, opened 2015-02-06
Add missing virtuals

HaoK:
Enable unit test verifying methods are virtual

Fixes https://github.com/aspnet/Identity/issues/349

cc @divega @suhasj


divega:
:shipit:

PR #354, opened 2015-02-09
Defered logging

suhasj:
@HaoK @divega 

Changes
- Fixed logging to pass a func instead of a string. This func is called in the logger only if the loglevel is enabled
- Fixed tests to pass func instead of string values as intended


src/Microsoft.AspNet.Identity/RoleManager.cs
```diff
@@ -404,12 +404,12 @@ private IRoleClaimStore<TRole> GetClaimStore()
         /// <param name="user"></param>
         /// <param name="methodName"></param>
         /// <returns></returns>
-        protected async Task<IdentityResult> LogResultAsync(IdentityResult result,
+        protected Task<IdentityResult> LogResultAsync(IdentityResult result,
             TRole role, [System.Runtime.CompilerServices.CallerMemberName] string methodName = "")
         {
-            result.Log(Logger, Resources.FormatLoggingResultMessageForRole(methodName, await GetRoleIdAsync(role)));
+            result.Log(Logger, () => { return Resources.FormatLoggingResultMessageForRole(methodName, GetRoleIdAsync(role).Result); });
```
divega: You shouldn't need `return` here.

divega: Hmmm. Sprinkling code that can potentially block everywhere isn't good. Let's chat about this more. We should probably hold of making this change until there is a version of `Log()` that takes an async lambda.

suhasj: This is called in the IdentityResult.Log method https://github.com/aspnet/Identity/blob/DeferedLogging/src/Microsoft.AspNet.Identity/IdentityResult.cs#L65 so it has to return some result and since the log methods are not async I do a block 

These are called only if the loglevel is enabled though

divega: Yo are missing my point. Try `() => Resources.FormatLoggingResultMessageForRole(...)`. Single line lambdas don't need `return` to return something.

suhasj: ah i see. Thanks I got that. 

But the `.Result` call has to still be there

HaoK: .Result is gross, we should just have it be LogAsync for now, and take a Func<Task<string>> so you can await it properly

suhasj:
@divega @HaoK 

Updated with async extension methods for logger and logging via fluent api. Some questions
- `WithLoggingAsync(Task<IdentityResult>)` calls `WithLoggingAsync(IdentityResult)` after calling an await in the later. can there be a way to beautify that.
- Should the method `WithLoggingAsync` be renamed to something else ? Some options I had
  * PostLoggingAsync
  * LogAndReturnAsync

suhasj:
@AndriySvyryd including you since @divega mentioned you could have thoughts on reducing excessive `async/await` in this case

src/Microsoft.AspNet.Identity/SignInManager.cs
```diff
@@ -64,13 +64,13 @@ public class SignInManager<TUser> where TUser : class
         {
             if (Options.SignIn.RequireConfirmedEmail && !(await UserManager.IsEmailConfirmedAsync(user)))
             {
-                return await LogResultAsync(false, user);
+                return await false.WithLoggingAsync(this, user);
             }
             if (Options.SignIn.RequireConfirmedPhoneNumber && !(await UserManager.IsPhoneNumberConfirmedAsync(user)))
             {
-                return await LogResultAsync(false, user);
+                return await false.WithLoggingAsync(this, user);
```
HaoK: Personally I think this looks kind of ugly compared to the earlier iteration, extension off of false just seems really strange, also now we are passing this everywhere, which makes me think the methods really belong on the appropriate Manager

HaoK: return await false.WithLoggingAsync just looks ugly to me, we are awaiting a bool?

src/Microsoft.AspNet.Identity/TaskExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.Identity
+{
+    public static class TaskExtensions
+    {
+        public static async Task<IdentityResult> WithLoggingAsync<TUser>(this Task<IdentityResult> identityResult, UserManager<TUser> userManager,TUser user,
+            [System.Runtime.CompilerServices.CallerMemberName] string methodName = "") where TUser :class
+        {
+            return await (await identityResult).WithLoggingAsync(userManager, user, methodName);
```
HaoK: Ick await (await ...)  Yeah this definitely makes me think the extension method on the result types are the wrong way to go

src/Microsoft.AspNet.Identity/SignInManager.cs
suhasj: I agree with that. We can a split where logging after Task<IdentityResult> can be an extension method while for bool results it can be in the Manager class itself.

Having an extension method makes it easy to have a fluent call that's all

src/Microsoft.AspNet.Identity/TaskExtensions.cs
suhasj: sure that is one place I need feedback on. Definitely felt wrong adding that. Alternatively it would be `var result = await task` and then call the method on it which is the same thing.

One thing I looked into was `Task.ContinueWith` but the parameter was `Task<IdentityResult>` and not `IdentityResult` so the await had to be there anyway

src/Microsoft.AspNet.Identity/SignInManager.cs
HaoK: There's a definite smell if you  are basically passing this to every log method.

src/Microsoft.AspNet.Identity/TaskExtensions.cs
HaoK: My main feedback is that I think the switch to extension methods overall makes the code look  worse, and its not worth it, what you had before was better IMO, @divega what are your thoughts?

src/Microsoft.AspNet.Identity/LoggerExtensions.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// <param name="logger"></param>
+        /// <param name="state"></param>
+        /// <param name="exception"></param>
+        /// <param name="formatter"></param>
+        /// <returns></returns>
+        public static Task WriteInformationAsync(this ILogger logger, object state, Exception exception, Func<object, Exception, Task<string>> formatter)
+        {
+            if (logger.IsEnabled(LogLevel.Information))
+            {
+                return logger.WriteAsync(LogLevel.Information, 0, state, exception, formatter);
```
divega: This is invoking the `WriteAsync()` method defined above with a specific `LogLevel`, right? Why check if `logger.IsEnabled()` twice?

src/Microsoft.AspNet.Identity/LoggerExtensions.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// <param name="logger"></param>
+        /// <param name="state"></param>
+        /// <param name="exception"></param>
+        /// <param name="formatter"></param>
+        /// <returns></returns>
+        public static Task WriteWarningAsync(this ILogger logger, object state, Exception exception, Func<object, Exception, Task<string>> formatter)
+        {
+            if (logger.IsEnabled(LogLevel.Warning))
+            {
+                return logger.WriteAsync(LogLevel.Warning, 0, state, exception, formatter);
```
divega: Same as above, why check for `logger.IsEnabled()` twice?

davidfowl:
I haven't really looked by why is there async logging?

HaoK:
There is no way to synchronously get the user name from identity. It's async only.  So the helper methods that generate log messages that include the user name have to be async. Logging itself doesn't need to be.

divega:
@davidfowl In the particular case of Identity, producing a helpful message to be sent to the log can involve async work, e.g. getting the username for a user is done invoking an async API (note that the default implementation isn't really async, but the design allows for it so se make no assumptions). This may apply or not to other frameworks.

src/Microsoft.AspNet.Identity/SignInManager.cs
divega: Agreed that this one looks weird. Also it seems we could refactor the method to capture the outcome and three places and only log and return it at the end.

src/Microsoft.AspNet.Identity/SignInManager.cs
```diff
@@ -214,7 +215,7 @@ private static ClaimsIdentity CreateIdentity(TwoFactorAuthenticationInfo info)
             }
             var token = await UserManager.GenerateTwoFactorTokenAsync(user, provider);
             await UserManager.NotifyTwoFactorTokenAsync(user, provider, token);
-            return await LogResultAsync(true, user);
+            return await true.WithLoggingAsync(this, user);
```
divega: Are we missing logging for the `return false` outcomes above?

src/Microsoft.AspNet.Identity/RoleManager.cs
```diff
@@ -286,7 +286,7 @@ public virtual string NormalizeKey(string key)
             ThrowIfDisposed();
             await Store.SetRoleNameAsync(role, name, CancellationToken);
             await UpdateNormalizedRoleNameAsync(role);
-            return await LogResultAsync(IdentityResult.Success, role);
+            return await Task.FromResult(IdentityResult.Success).WithLoggingAsync(this, role);
```
divega: Why isn't this using the version of `WithLoggingAsync()` that takes an `IdentityResult` as opposed to `Task<IdentityResult>`?

src/Microsoft.AspNet.Identity/TaskExtensions.cs
```diff
@@ -0,0 +1,44 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.AspNet.Identity
+{
+    public static class TaskExtensions
```
divega: I would put these alongside the ones in `ResultExtensions`. Results just happen to sometimes be async.

src/Microsoft.AspNet.Identity/TaskExtensions.cs
divega: Well, await (await ...) just happens to be what you need sometimes although this is the one you could probably improve a bit with `ContinueWith()` (@suhasj note that the task in the delegate argument of `ContinueWith()` is already completed, or failed).

I don't completely hate the `WithLogging()` experiment. It has a good vibe in that it makes Logging a separate "aspect" that only applies to the results of operations. The one thing that makes me think it is isn't the right way to go is that on every single call we are passing the "this" argument, e.g. the UserManager, RoleManager, SignInManager instance. It is a clear smell that an instance method on those (the way we had it before) would make the most sense.

src/Microsoft.AspNet.Identity/TaskExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        public static async Task<IdentityResult> WithLoggingAsync<TUser>(this Task<IdentityResult> identityResult, UserManager<TUser> userManager,TUser user,
+            [System.Runtime.CompilerServices.CallerMemberName] string methodName = "") where TUser :class
+        {
+            return await (await identityResult).WithLoggingAsync(userManager, user, methodName);
+        }
+
+        public static async Task<IdentityResult> WithLoggingAsync<TRole>(this Task<IdentityResult> identityResultDelegate, RoleManager<TRole> roleManager, TRole role,
```
divega: Why the name `identityResultDelegate`? What makes it a delegate?

src/Microsoft.AspNet.Identity/TaskExtensions.cs
HaoK: Yeah I said the same thing about the "this" smell. Also result extensions is no longer extensible unlike the virtual method. If they add their own derived sign in result with a new status code there is no way to update the logging extension method.

src/Microsoft.AspNet.Identity/RoleManager.cs
divega: Ah, because such version doesn't exist yet for RoleManager... I think it should exist.

src/Microsoft.AspNet.Identity/TaskExtensions.cs
divega: > Also result extensions is no longer extensible unlike the virtual method

Good point too.

src/Microsoft.AspNet.Identity/LoggerExtensions.cs
```diff
@@ -0,0 +1,74 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Threading.Tasks;
+using Microsoft.Framework.Logging;
+
+namespace Microsoft.AspNet.Identity
+{
+    public static class LoggerExtensions
```
divega: @suhasj When you go back to instance methods on UserManager, RoleManager and SignInManager, take a look at getting rid of this level of helpers. Seem largely unnecessary. You can do the check on the other ones and I don't think it is that much duplication.

src/Microsoft.AspNet.Identity/RoleManager.cs
divega: Of course as a instance virtual method on `RoleManager`.

src/Microsoft.AspNet.Identity/SignInManager.cs
suhasj: The token is always generated so there is not false result as such hence omitted that

src/Microsoft.AspNet.Identity/TaskExtensions.cs
suhasj: I thought we grouped extension methods based on the classes on which they hung on. I club all these log extension methods in one place ( if we end up doing them )

src/Microsoft.AspNet.Identity/TaskExtensions.cs
suhasj: Because we were awaiting on it, I thought it was good to put the delegate in the name. Now I feel delegate makes sense only if it is a `Func` or an `Action`. I'll change it

src/Microsoft.AspNet.Identity/LoggerExtensions.cs
suhasj: So this method is public and users can call it separately, so this additional layer of check. I can have public wrapper and make a private method without the check and remove the duplication

src/Microsoft.AspNet.Identity/RoleManager.cs
suhasj: There is just one instance where we return IdentityResult instead of Task in RoleManager so I did not add that. I'll add that change

src/Microsoft.AspNet.Identity/LoggerExtensions.cs
```diff
@@ -0,0 +1,74 @@
...
+        /// <param name="formatter"></param>
+        /// <returns></returns>
+        public static async Task WriteAsync(this ILogger logger, LogLevel logLevel, int eventId, object state, Exception exception,
+            Func<object, Exception, Task<string>> formatter)
+        {
+            if (logger.IsEnabled(logLevel))
+            {
+                if (formatter != null)
+                {
+                    var message = await formatter(state, exception);
```
AndriySvyryd: Can't the formatter be sync?

src/Microsoft.AspNet.Identity/ResultExtensions.cs
```diff
@@ -0,0 +1,102 @@
...
+
+            return result;
+        }
+
+        public static async Task<bool> WithLoggingAsync<TUser>(this bool result, UserManager<TUser> userManager, TUser user,
+            [System.Runtime.CompilerServices.CallerMemberName] string methodName = "") where TUser : class
+        {
+            var logLevel = result ? LogLevel.Information : LogLevel.Warning;
+
+            await userManager.Logger.WriteAsync(logLevel, 0, methodName, null, async (callingMethod, exception) =>
```
AndriySvyryd: Assuming that waiting for the write operation to finish isn't required to proceed with execution then it's not necessary to await on it.

suhasj: so we call the method and continue. There won't be scenario when the execution fails because the calling thread is no longer executing right ?

AndriySvyryd: If the call is not awaited on it will have nothing to do with the calling thread.
Also if an exception is thrown during execution of the called method it will not be observed unless there's a handler registered on TaskScheduler.UnobservedTaskException

suhasj:
@divega @HaoK 

Updated with review comments and this looks pretty good to me. I have corrected some doc comments as well

One possible stopper for checkin might be, right now the logging is done as 
`var baseMessage = Resources.Format(...)
Logger.WriteX(Resource.Format(baseMessage, args...)`

let me know if you want it as 
`Logger.WriteX(Resource.Format(Resources.Format(..,..), args...)`. I choose the first one just so it is more readable.

src/Microsoft.AspNet.Identity/UserManager.cs
```diff
@@ -1915,10 +1919,26 @@ public virtual Task<IList<TUser>> GetUsersInRoleAsync(string roleName)
...
+            {
+                if (Logger.IsEnabled(LogLevel.Warning))
+                {
+                    var baseMessage = Resources.FormatLoggingResultMessage(methodName, await GetUserIdAsync(user));
+                    Logger.WriteWarning(Resources.FormatLogIdentityResultFailure(baseMessage,
+                        string.Join(",", result.Errors.Select(x => x.Code).ToList())));
+                }
+            }
             return result;
         }
```
HaoK: I don't really like how we have two LogResult methods here, one for result, and one for bool.  Seems pretty clunky, this might be a sideeffect of removing the virtual Log methods from the various result objects, remind me why we removed the result virtual Log methods again?

suhasj: The virtual Log method on the IdentityResult was not called by the bool overload anyway so they were present in the last checkin as well. The decision came because we constructed half message in UserManager/RoleManager and rest in the IdentityResult. This make convenient where do the whole construction of the message in one place after checking whether log level is enabled.

src/Microsoft.AspNet.Identity/UserManager.cs
```diff
@@ -1929,17 +1949,23 @@ public virtual Task<IList<TUser>> GetUsersInRoleAsync(string roleName)
...
+        protected virtual async Task<bool> LogResultAsync(bool result,
             TUser user, [System.Runtime.CompilerServices.CallerMemberName] string methodName = "")
         {
-            var baseMessage = Resources.FormatLoggingResultMessage(methodName, await GetUserIdAsync(user));
-            if (result)
-            {
-                Logger.WriteInformation(string.Format("{0} : {1}", baseMessage, result.ToString()));
-            }
-            else
+            var logLevel = result ? LogLevel.Information : LogLevel.Warning;
```
divega: This pattern looks a bit better than the one we have in the other methods, i.e. decide the log level, put it in a variable and checking if it is enabled instead of having nested ifs.

src/Microsoft.AspNet.Identity/UserManager.cs
```diff
@@ -1929,17 +1949,23 @@ public virtual Task<IList<TUser>> GetUsersInRoleAsync(string roleName)
...
+
+            if (Logger.IsEnabled(logLevel))
             {
-                Logger.WriteWarning(string.Format("{0} : {1}", baseMessage, result.ToString()));
+                var baseMessage = Resources.FormatLoggingResultMessage(methodName, await GetUserIdAsync(user));
+
+                if(logLevel == LogLevel.Warning)
```
divega: Is there a version of `Write` we can use from logging that would just take `logLevel` as a parameter?

src/Microsoft.AspNet.Identity/SignInManager.cs
```diff
@@ -415,15 +418,43 @@ public virtual AuthenticationProperties ConfigureExternalAuthenticationPropertie
...
-        protected async virtual Task<SignInResult> LogResultAsync(SignInResult status, TUser user, [System.Runtime.CompilerServices.CallerMemberName] string methodName = "")
+        protected virtual async Task<SignInResult> LogResultAsync(SignInResult result, TUser user,
+        [System.Runtime.CompilerServices.CallerMemberName] string methodName = "")
         {
-            status.Log(Logger, Resources.FormatLoggingResultMessage(methodName, await UserManager.GetUserIdAsync(user)));
+            if (Logger.IsEnabled(LogLevel.Information))
+            {
+                var status = "";
+
+                if (result.IsLockedOut)
```
divega: Consider moving this to a method in result that returns the right status string.

src/Microsoft.AspNet.Identity/UserManager.cs
divega: @HaoK Why would the result object need to know how to log? I prefer that it just knows how to return relevant data to construct the log message.

src/Microsoft.AspNet.Identity/UserManager.cs
suhasj: you mean like `Write(LogLevel level, string message)` ? no there isn't. The direct method off course takes eventid, state, exception and func

src/Microsoft.AspNet.Identity/UserManager.cs
HaoK: @divega Sure, I just mean the Result itself should have a say in how to format its log message, whether it does the final log, or generates a string for someone else to log doesn't matter, but the result object should have a say is my main point

divega: Agreed, I just suggested that the logic to decide the status string (e.g. if (result.IsLockedOut) status = "Lockedout", etc) probably belongs in one of the result types. Maybe there are larger chunks that can be moved there.

src/Microsoft.AspNet.Identity/UserManager.cs
divega: And accepts those to be null?

src/Microsoft.AspNet.Identity/UserManager.cs
suhasj: Sure. But we still need an additional check inside to call `LogInformation` or `LogWarning` appropriately based on the result but yes this reduces one line for sure

src/Microsoft.AspNet.Identity/SignInManager.cs
suhasj: sure

src/Microsoft.AspNet.Identity/RoleManager.cs
```diff
@@ -387,14 +391,29 @@ private IRoleClaimStore<TRole> GetClaimStore()
         ///     Logs the current Identity Result and returns result object
         /// </summary>
         /// <param name="result"></param>
-        /// <param name="user"></param>
+        /// <param name="role"></param>
         /// <param name="methodName"></param>
         /// <returns></returns>
-        protected async Task<IdentityResult> LogResultAsync(IdentityResult result,
+        protected virtual async Task<IdentityResult> LogResultAsync(IdentityResult result,
```
divega: Can you refactor this method to also use something like `var logLevel = result.Succeeded ? LogLevel.Information : LogLevel.Warning` etc. as you did in a different version of the method?

src/Microsoft.AspNet.Identity/UserManager.cs
suhasj: sure Logger.Write(LogLevel.Information, eventId, message, null, null) should work the same but yes not easily super concise. On the flip side, if we make the eventId configurable, then users can set specific eventId for Identity messages something not possible right now. Console logger doesn't much care about the eventId but other loggers might

src/Microsoft.AspNet.Identity/RoleManager.cs
suhasj: sure I am doing it in all places

src/Microsoft.AspNet.Identity/UserManager.cs
divega: We can chat about allowing the eventId configurable. The main reason I am asking right now bout this though is because you can replace the whole if with a single line of code.

suhasj:
Updated with result classes to have a two string implementation.

One question pending is that the `LogResultAsync` (methods)[https://github.com/aspnet/Identity/pull/354/files#diff-bfbbcc7825aec8772885e70b17f60a84L409] in SignInManager  seems to have the same logic but operate on `bool` and `SignInResult` objects. They can be merged into a single method but the signature would `LogResultAsync(object result, ..)` which I feel is not recommended. Pointers on how to make this better.

Apart from that, added did some refactoring to static properties as suggested by @divega

src/Microsoft.AspNet.Identity/SignInResult.cs
```diff
@@ -40,74 +32,62 @@ public class SignInResult
...
 
         /// <summary>
         ///     Static failure result
         /// </summary>
         /// <returns></returns>
-        public static SignInResult Failed
-        {
-            get { return _failed; }
-        }
+        public static SignInResult Failed => new SignInResult();
```
HaoK: This change now makes it so we are returning new instances every time the static property is called no?

src/Microsoft.AspNet.Identity/SignInResult.cs
```diff
@@ -40,74 +32,62 @@ public class SignInResult
...
 
         /// <summary>
-        ///     Log result based on properties
+        ///     Returns string implemenation of the result. Provides string value for the assigned property
         /// </summary>
-        /// <param name="logger"></param>
-        /// <param name="message"></param>
-        public virtual void Log(ILogger logger, string message)
+        /// <returns></returns>
+        public override string ToString()
```
HaoK: You can make this read nicer with something like this
```
return (IsLockedOut ? "LockedOut" :
           IsNotAllowed ? "NotAllowed" :
           ""

Comment (HaoK) [inline src/Microsoft.AspNet.Identity/SignInResult.cs]: Also the default should be Unknown as opposed to empty string.  Also shouldn't these go in a resource for loc?

File: src/Microsoft.AspNet.Identity/SignInResult.cs (hunk 18 above)
Comment (divega) [inline src/Microsoft.AspNet.Identity/SignInResult.cs]: Good point. My fault. I suggested @suhasj to look into something like this just to make this look simpler.

Comment (suhasj) [inline src/Microsoft.AspNet.Identity/SignInResult.cs]: I had the same question but this change was suggested by Resharper as well. I can revert this back

File: src/Microsoft.AspNet.Identity/SignInResult.cs (hunk 19 above)
Comment (suhasj) [inline src/Microsoft.AspNet.Identity/SignInResult.cs]: Me and @divega had this chat and the consensus is that this is a behavior similar to enum so in that scenario it makes sense to hard code it.

File: src/Microsoft.AspNet.Identity/SignInResult.cs (hunk 18 above)
Comment (divega) [inline src/Microsoft.AspNet.Identity/SignInResult.cs]: We can just bring back the field and do `public static SignInResult Failed => _failed`

Comment (suhasj) [inline src/Microsoft.AspNet.Identity/SignInResult.cs]: sure I was making that change :)

**Comment by suhasj** [comment]
Updated

File: src/Microsoft.AspNet.Identity/SignInResult.cs (hunk 19 above)
Comment (divega) [inline src/Microsoft.AspNet.Identity/SignInResult.cs]: I actually wasn't sure about it when we finished the conversation, so I think it is more of a question: These currently look like the implementation of `ToString()` for an enum, e.g. the strings represent members of the type in PascalCase. Should we return a descriptive name instead? If these are members of the type I don't think they should be localized. If they are descriptive strings then they should be. If they are not localized then I don't know if there is a good reason to still stick them in resources. @HaoK what do you think?

File: src/Microsoft.AspNet.Identity/SignInResult.cs [hunk 20]
```diff
@@ -40,74 +38,43 @@ public class SignInResult
...
         /// <returns></returns>
-        public static SignInResult TwoFactorRequired
-        {
-            get { return _twoFactorRequired; }
-        }
+        public static SignInResult TwoFactorRequired => _twoFactorRequired;
 
         /// <summary>
-        ///     Log result based on properties
+        ///     Returns string implemenation of the result. Provides string value for the assigned property
```
divega: Implementation or representation?

divega: The second sentence is redundant.

suhasj: sure

src/Microsoft.AspNet.Identity/SignInResult.cs
```diff
@@ -40,74 +38,43 @@ public class SignInResult
...
-            }
-            else
-            {
-                logger.WriteInformation(Resources.FormatLoggingSigninResult(message, "Failed"));
-            }
+            return IsLockedOut
+                ? "Lockedout"
+                : IsNotAllowed
+                    ? "NotAllowed"
+                    : RequiresTwoFactor ? "RequiresTwoFactor" : Succeeded ? "Succeeded" : "Failed";
```
divega: Nit: I like the indentation and new lines to be the way it appeared in @HaoK's suggestion.

src/Microsoft.AspNet.Identity/SignInManager.cs
```diff
@@ -415,15 +418,20 @@ public virtual AuthenticationProperties ConfigureExternalAuthenticationPropertie
...
         ///     Log SignInStatus for user and return SignInStatus
         /// </summary>
-        /// <param name="status"></param>
+        /// <param name="result"></param>
         /// <param name="user"></param>
         /// <param name="methodName"></param>
         /// <returns></returns>
-        protected async virtual Task<SignInResult> LogResultAsync(SignInResult status, TUser user, [System.Runtime.CompilerServices.CallerMemberName] string methodName = "")
+        protected virtual async Task<SignInResult> LogResultAsync(SignInResult result, TUser user,
+        [System.Runtime.CompilerServices.CallerMemberName] string methodName = "")
```
divega: Nit: Add a using with the namespace of the attribute everywhere and make this shorter.

src/Microsoft.AspNet.Identity/SignInManager.cs
```diff
@@ -406,8 +405,12 @@ public virtual AuthenticationProperties ConfigureExternalAuthenticationPropertie
         /// <returns></returns>
         protected async virtual Task<bool> LogResultAsync(bool result, TUser user, [System.Runtime.CompilerServices.CallerMemberName] string methodName = "")
         {
-            Logger.WriteInformation(Resources.FormatLoggingSigninResult(Resources.FormatLoggingResultMessage(methodName,
-                await UserManager.GetUserIdAsync(user)), result));
+            // Check if log level is enabled before creating the message.
```
divega: In UserManager we treat false as warning, e.g. `var logLevel = result ? LogLevel.Information : LogLevel.Warning`. Is it intentional that we treat true and false the same here?

suhasj: I aligned the experience with the logging guidelines given at https://github.com/aspnet/Logging/wiki/Guidelines which says `Warning` is for unexpected flow in the application. In SignInManager, the result is bool for operations that are mostly read-only and most of the time it is an expected value. You might argue that locking out user might be an unexpected error but it is a known result. Hence I choose to log all flows of SignInManager. In contrast, adding an existing role to the database or creating a user with existing username is an error in the happy flow hence I chose it as Warning

I am open to this being changed since this is subjective. From a uniformity perspective yes, we can the same in SignInManager as well.

divega: The guidelines are good information. I guess we need to decide when we apply them to Identity whether it is more helpful to log something like a user locked out as a warning. @HaoK what is your take?

src/Microsoft.AspNet.Identity/SignInResult.cs
HaoK: I don't feel super strongly about this one, we can go with them being more like an Enum and they should just match the boolean properties exactly then, since that should be easier

src/Microsoft.AspNet.Identity/SignInResult.cs
```diff
@@ -40,74 +38,42 @@ public class SignInResult
...
-            }
-            else if (Succeeded)
-            {
-                logger.WriteInformation(Resources.FormatLoggingSigninResult(message, "Succeeded"));
-            }
-            else
-            {
-                logger.WriteInformation(Resources.FormatLoggingSigninResult(message, "Failed"));
-            }
+            return IsLockedOut ? "Lockedout" :
```
HaoK: nit: I'm not sure if this is the PR messed up or not, but can we make the indentation for all of the checks on the following line on the same indent level with IsLockedOut?

suhasj: ah format document messed it up

src/Microsoft.AspNet.Identity/SignInManager.cs
HaoK: So I would say, the two 'success' results for SignIn manager are Success and TwoFactorRequired, all of the other results should be considered failures and logged like user manager failures.

I would try True/False logging in the same way as UserManager

HaoK: This sort of goes back to my point about the result object knowing how to render its log message, as today you'd have to put a switch inside SignInManager's Log to do this, while I think this decision should be taken care of by the result object (whether its currently a success or a failure).  That was why I origionally suggested for the Log to happen inside the result, while the managers method was mostly just to format the outer string and pass it into XXXResult.Log(outerformat), and the result call the appropriate Log method with warning/information since only the result knows truly whether its a failure or success

HaoK: This is even more clear in the derived result case, when they extend SignInResult with an additional failure, they'd have to override both SignInResult.ToString and SignInManager.LogResult today, instead of just overriding SignInResult.ToString to add logic for any new status codes, and then deferring to the parent behavior

HaoK: So I think we should add a test for each managers demonstrating a derived result class which adds a new status and updates the logging messages to handle them, that should make it easier to review what the extensibility story actually looks like, @suhasj can you add these tests as part of this PR?

HaoK:
Replaced by https://github.com/aspnet/Identity/pull/410 i  think

Issue #355, opened 2015-02-10
Consider getting rid of CancelationToken parameter on SignInManager methods


divega:
Top level objects in Identity such as SignInManager depend on an HttpContext which already carries a CancelationToken. Unless there are other valid sources of cancelation than the cancelation of the request, it should be ok to remove the additional CancelationToken parameters in the async methods and use the CancelationToken from the HttpContext instead. 

Currently the identity stores and manager classes do not depend on HttpContext so their methods still need the CancelationToken to be obtained and passed.

PR #356, opened 2015-02-10
Remove CancellationToken from Manager APIs

HaoK:
Only the Stores continue to take CancellationToken, but they are no longer optional 

Fixes https://github.com/aspnet/Identity/issues/355

cc @divega @suhasj


suhasj:
I thought we were removing the cancellation token from SigninManager only ? Isnt that the case ?

HaoK:
Our managers are scoped anyways, so it seems safe to just have them depend on HttpContextAccessor in the ctor as well, @divega thoughts?

divega:
@HaoK @suhasj Adding the HttpContextAccessor dependency seems fine. It was my thought was well when I realized that only SignInManager had it. A couple of points:

1. I believe the removal of CancellationToken from methods of some of the interfaces adds an implicit assumption that the implementaiton of the interface will have access to the CancelationToken from the HttpContext. This might be ok. I just want us to think it through.

2. @yishaigalatzer mentioned that there were potentially multiple sources of cancellation. If you haven't already, could you follow up and make sure that listening to the CancellationToken in HttpContext is enough?

Otherwise it looks :shipit: to me.

HaoK:
@divega All of the interface implementations can be activated from DI, and as such they can all use HttpContextAccessor to get access to the cancellation token if they desire (we don't do this today, but custom implementations could if they want)

divega:
> we don't do this today
@HaoK is there anything inherently cancellable that we are missing? BTW, I noticed that you left the CancellationToken in place for the messaging API, which seems the best choice.

HaoK:
I'm not really sure, to be honest, we never got around to writing any tests that were exercising the cancellation tokens anyways...  So I'm glad we got rid of them for that reason alone :)

divega:
Ok. I took a look at some of the interfaces, e.g. validation. If none of the default implementations of the interfaces use databases or need to potentially make calls over the network, I think it is fine to remove the CancelationToken from the methods. As you said, customer implementations can get the IHttpContextAccessor if needed. 

What about the second concern? Can you follow up with @yishaigalatzer?

HaoK:
Sure I'll can follow up with Yishai

PR #360, opened 2015-02-12
API Review cleanup #1

HaoK:
- Remove constructor defaults
- Make service properties internal
- Add Logging/HttpContextAccessor services which are required by
identity

Fixes https://github.com/aspnet/Identity/issues/358
Fixes https://github.com/aspnet/Identity/issues/357

cc @divega @suhasj


divega:
:shipit:

Issue #359, opened 2015-02-12
Remove IdentityMessage/interfaces


HaoK:
Since the Email/SMS services already live in the app, we can probably remove the messaging interfaces from core entirely, and just leave enough hooks for the app to directly plug in its messaging services

PR #362, opened 2015-02-12
Remove messaging providers/interfaces

HaoK:
Mostly just a lot of deleting code...

Fixes https://github.com/aspnet/Identity/issues/359

cc @divega @suhasj


samples/IdentitySample.Mvc/Startup.cs
```diff
@@ -86,39 +80,5 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerFactory)
             SampleData.InitializeIdentityDatabaseAsync(app.ApplicationServices).Wait();
         }
 
-        public class EmailMessageProvider : IIdentityMessageProvider
```
suhasj: nit: should these be is separate classes ?

HaoK: I'm not sure what you are asking here, this code is deleted

suhasj: ah my bad... one of those brain drain days

HaoK:
Merged c2e96fa570593688ccd159372562f6bbe545b640

PR #377, opened 2015-02-27
Update to new version of Linq.Expressions

suhasj:
@divega @HaoK 
Updating version since Identity is failing on CI because of recent changes


suhasj:
Merged to unblock CI. Let me know if i need to revert changes

PR #378, opened 2015-02-27
Moved POCOs to EF and fixed tests

suhasj:
@HaoK I moved the POCOs to Identity.EF. The Identity.Test and Identity.InMemory.Test broke due to this change since they did not reference Identity.EF. Now we could either duplicate the code in the tests or reference the assembly and include the namespace. In addition the UserManagerTestBase is shared so that has to updated also. I choose to reference the assembly as this seems more cleaner. I made sure we reference only the POCOs and nothing else. Let me know if there is a better way of doing this.

/cc @divega 

You can skip the nitpick feedback for this PR

### DIFF
--- a/src/Microsoft.AspNet.Identity.EntityFramework/IdentityUser.cs
+++ b/src/Microsoft.AspNet.Identity.EntityFramework/IdentityUser.cs
@@ -2,8 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 
-namespace Microsoft.AspNet.Identity
+namespace Microsoft.AspNet.Identity.EntityFramework
 {
     public class IdentityUser : IdentityUser<string>
     {
@@ -87,5 +88,20 @@ namespace Microsoft.AspNet.Identity
         ///     Used to record failures for the purposes of lockout
         /// </summary>
         public virtual int AccessFailedCount { get; set; }
+
+        /// <summary>
+        ///     Navigation property for users in the role
+        /// </summary>
+        public virtual ICollection<IdentityUserRole<TKey>> Roles { get; set; } = new List<IdentityUserRole<TKey>>();
+
+        /// <summary>
+        ///     Navigation property for users claims
+        /// </summary>
+        public virtual ICollection<IdentityUserClaim<TKey>> Claims { get; set; } = new List<IdentityUserClaim<TKey>>();
+
+        /// <summary>
+        ///     Navigation property for users logins
+        /// </summary>
+        public virtual ICollection<IdentityUserLogin<TKey>> Logins { get; set; } = new List<IdentityUserLogin<TKey>>();
     }
 }
--- a/src/Microsoft.AspNet.Identity.EntityFramework/IdentityUser.cs
+++ b/src/Microsoft.AspNet.Identity.EntityFramework/IdentityUser.cs
@@ -92,16 +92,16 @@ namespace Microsoft.AspNet.Identity.EntityFramework
         /// <summary>
         ///     Navigation property for users in the role
         /// </summary>
-        public virtual ICollection<IdentityUserRole<TKey>> Roles { get; set; } = new List<IdentityUserRole<TKey>>();
+        public virtual ICollection<IdentityUserRole<TKey>> Roles { get; } = new List<IdentityUserRole<TKey>>();
 
         /// <summary>
         ///     Navigation property for users claims
         /// </summary>
-        public virtual ICollection<IdentityUserClaim<TKey>> Claims { get; set; } = new List<IdentityUserClaim<TKey>>();
+        public virtual ICollection<IdentityUserClaim<TKey>> Claims { get; } = new List<IdentityUserClaim<TKey>>();
 
         /// <summary>
         ///     Navigation property for users logins
         /// </summary>
-        public virtual ICollection<IdentityUserLogin<TKey>> Logins { get; set; } = new List<IdentityUserLogin<TKey>>();
+        public virtual ICollection<IdentityUserLogin<TKey>> Logins { get; } = new List<IdentityUserLogin<TKey>>();
     }
 }
--- a/test/Microsoft.AspNet.Identity.EntityFramework.Test/DefaultPocoTest.cs
+++ b/test/Microsoft.AspNet.Identity.EntityFramework.Test/DefaultPocoTest.cs
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Linq;
+using System.Security.Claims;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Builder;
 using Microsoft.AspNet.Identity.Test;
@@ -14,7 +16,7 @@ namespace Microsoft.AspNet.Identity.EntityFramework.Test
     [TestCaseOrderer("Microsoft.AspNet.Identity.Test.PriorityOrderer", "Microsoft.AspNet.Identity.EntityFramework.Test")]
     public class DefaultPocoTest
     {
-        private readonly string ConnectionString = @"Server=(localdb)\mssqllocaldb;Database=DefaultSchemaTest" + DateTime.Now.Month + "-" + DateTime.Now.Day + "-" + DateTime.Now.Year + ";Trusted_Connection=True;";
+        private readonly string ConnectionString = @"Server=(localdb)\mssqllocaldb;Database=DefaultSchemaTest" + DateTime.Now.Month + "-" + DateTime.Now.Day + "-" + DateTime.Now.Year + ";Trusted_Connection=True;MultipleActiveResultSets=true";
         public IdentityDbContext CreateContext(bool ensureCreated = false)
         {
             var db = DbUtil.Create(ConnectionString);
@@ -60,6 +62,148 @@ namespace Microsoft.AspNet.Identity.EntityFramework.Test
             var user = new IdentityUser { UserName = userName };
             IdentityResultAssert.IsSuccess(await userManager.CreateAsync(user, password));
             IdentityResultAssert.IsSuccess(await userManager.DeleteAsync(user));
+        }
+
+        [Fact]
+        public async Task CanIncludeUserClaimsTest()
+        {
+            // Arrange
+            CreateContext(true);
+            var builder = new ApplicationBuilder(CallContextServiceLocator.Locator.ServiceProvider);
+
+            var services = new ServiceCollection();
+            DbUtil.ConfigureDbServices<IdentityDbContext>(ConnectionString, services);
+            services.AddIdentity<IdentityUser, IdentityRole>().AddEntityFrameworkStores<IdentityDbContext>();
+            builder.ApplicationServices = services.BuildServiceProvider();
+
+            var userManager = builder.ApplicationServices.GetRequiredService<UserManager<IdentityUser>>();
+            var dbContext = builder.ApplicationServices.GetRequiredService<IdentityDbContext>();
+
+            var username = "user" + new Random().Next();
+            var user = new IdentityUser() { UserName = username };
+            IdentityResultAssert.IsSuccess(await userManager.CreateAsync(user));
+
+            for (var i = 0; i < 10; i++)
+            {
+                IdentityResultAssert.IsSuccess(await userManager.AddClaimAsync(user, new Claim(i.ToString(), "foo")));
+            }
+
+            user = dbContext.Users.Include(x => x.Claims).FirstOrDefault(x => x.UserName == username);
+
+            // Assert
+            Assert.NotNull(user);
+            Assert.NotNull(user.Claims);
+            Assert.Equal(10, user.Claims.Count());
+        }
+
+        [Fact]
+        public async Task CanIncludeUserLoginsTest()
+        {
+            // Arrange
+            CreateContext(true);
+            var builder = new ApplicationBuilder(CallContextServiceLocator.Locator.ServiceProvider);
+
+            var services = new ServiceCollection();
+            DbUtil.ConfigureDbServices<IdentityDbContext>(ConnectionString, services);
+            services.AddIdentity<IdentityUser, IdentityRole>().AddEntityFrameworkStores<IdentityDbContext>();
+            builder.ApplicationServices = services.BuildServiceProvider();
+
+            var userManager = builder.ApplicationServices.GetRequiredService<UserManager<IdentityUser>>();
+            var dbContext = builder.ApplicationServices.GetRequiredService<IdentityDbContext>();
+
+            var username = "user" + new Random().Next();
+            var user = new IdentityUser() { UserName = username };
+            IdentityResultAssert.IsSuccess(await userManager.CreateAsync(user));
+
+            for (var i = 0; i < 10; i++)
+            {
+                IdentityResultAssert.IsSuccess(await userManager.AddLoginAsync(user, new UserLoginInfo("foo" + i, "bar" + i, "foo")));
+            }
+
+            user = dbContext.Users.Include(x => x.Logins).FirstOrDefault(x => x.UserName == username);
+
+            // Assert
+            Assert.NotNull(user);
+            Assert.NotNull(user.Logins);
+            Assert.Equal(10, user.Logins.Count());
+        }
+
+        [Fact]
+        public async Task CanIncludeUserRolesTest()
+        {
+            // Arrange
+            CreateContext(true);
+            var builder = new ApplicationBuilder(CallContextServiceLocator.Locator.ServiceProvider);
+
+            var services = new ServiceCollection();
+            DbUtil.ConfigureDbServices<IdentityDbContext>(ConnectionString, services);
+            services.AddIdentity<IdentityUser, IdentityRole>().AddEntityFrameworkStores<IdentityDbContext>();
+            builder.ApplicationServices = services.BuildServiceProvider();
+
+            var userManager = builder.ApplicationServices.GetRequiredService<UserManager<IdentityUser>>();
+            var roleManager = builder.ApplicationServices.GetRequiredService<RoleManager<IdentityRole>>();
+            var dbContext = builder.ApplicationServices.GetRequiredService<IdentityDbContext>();
+
+            const string roleName = "Admin";
+            for (var i = 0; i < 10; i++)
+            {
+                IdentityResultAssert.IsSuccess(await roleManager.CreateAsync(new IdentityRole(roleName + i)));
+            }
+            var username = "user" + new Random().Next();
+            var user = new IdentityUser() { UserName = username };
+            IdentityResultAssert.IsSuccess(await userManager.CreateAsync(user));
+
+            for (var i = 0; i < 10; i++)
+            {
+                IdentityResultAssert.IsSuccess(await userManager.AddToRoleAsync(user, roleName + i));
+            }
+
+            user = dbContext.Users.Include(x => x.Roles).FirstOrDefault(x => x.UserName == username);
+
+            // Assert
+            Assert.NotNull(user);
+            Assert.NotNull(user.Roles);
+            Assert.Equal(10, user.Roles.Count());
+
+            for (var i = 0; i < 10; i++)
+            {
+                var role = dbContext.Roles.Include(r => r.Users).FirstOrDefault(r => r.Name == (roleName + i));
+                Assert.NotNull(role);
+                Assert.NotNull(role.Users);
+                Assert.Equal(1, role.Users.Count());
+            }
+        }
+
+        [Fact]
+        public async Task CanIncludeRoleClaimsTest()
+        {
+            // Arrange
+            CreateContext(true);
+            var builder = new ApplicationBuilder(CallContextServiceLocator.Locator.ServiceProvider);
+
+            var services = new ServiceCollection();
+            DbUtil.ConfigureDbServices<IdentityDbContext>(ConnectionString, services);
+            services.AddIdentity<IdentityUser, IdentityRole>().AddEntityFrameworkStores<IdentityDbContext>();
+            builder.ApplicationServices = services.BuildServiceProvider();
+
+            var roleManager = builder.ApplicationServices.GetRequiredService<RoleManager<IdentityRole>>();
+            var dbContext = builder.ApplicationServices.GetRequiredService<IdentityDbContext>();
+
+            var role = new IdentityRole("Admin");
+
+            IdentityResultAssert.IsSuccess(await roleManager.CreateAsync(role));
+
+            for (var i = 0; i < 10; i++)
+            {
+                IdentityResultAssert.IsSuccess(await roleManager.AddClaimAsync(role, new Claim("foo" + i, "bar" + i)));
+            }
+
+            role = dbContext.Roles.Include(x => x.Claims).FirstOrDefault(x => x.Name == "Admin");
+
+            // Assert
+            Assert.NotNull(role);
+            Assert.NotNull(role.Claims);
+            Assert.Equal(10, role.Claims.Count());
         }
 
         [TestPriority(10000)]
--- a/test/Microsoft.AspNet.Identity.InMemory.Test/InMemoryStoreTest.cs
+++ b/test/Microsoft.AspNet.Identity.InMemory.Test/InMemoryStoreTest.cs
@@ -2,12 +2,13 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Linq.Expressions;
 using Microsoft.AspNet.Identity.Test;
 using Microsoft.Framework.DependencyInjection;
 
 namespace Microsoft.AspNet.Identity.InMemory.Test
 {
-    public class InMemoryStoreTest : UserManagerTestBase<InMemoryUser, IdentityRole>
+    public class InMemoryStoreTest : UserManagerTestBase<TestUser, TestRole>
     {
         protected override object CreateTestContext()
         {
@@ -16,12 +17,44 @@ namespace Microsoft.AspNet.Identity.InMemory.Test
 
         protected override void AddUserStore(IServiceCollection services, object context = null)
         {
-            services.AddSingleton<IUserStore<InMemoryUser>, InMemoryUserStore<InMemoryUser>>();
+            services.AddSingleton<IUserStore<TestUser>, InMemoryUserStore<TestUser>>();
         }
 
         protected override void AddRoleStore(IServiceCollection services, object context = null)
         {
-            services.AddSingleton<IRoleStore<IdentityRole>, InMemoryRoleStore<IdentityRole>>();
+            services.AddSingleton<IRoleStore<TestRole>, InMemoryRoleStore<TestRole>>();
         }
+
+        protected override void SetUserPasswordHash(TestUser user, string hashedPassword)
+        {
+            user.PasswordHash = hashedPassword;
+        }
+
+        protected override TestUser CreateTestUser(string namePrefix = "", string email = "", string phoneNumber = "",
+            bool lockoutEnabled = false, DateTimeOffset? lockoutEnd = default(DateTimeOffset?), bool useNamePrefixAsUserName = false)
+        {
+            return new TestUser
+            {
+                UserName = useNamePrefixAsUserName ? namePrefix : string.Format("{0}{1}", namePrefix, Guid.NewGuid()),
+                Email = email,
+                PhoneNumber = phoneNumber,
+                LockoutEnabled = lockoutEnabled,
+                LockoutEnd = lockoutEnd
+            };
+        }
+
+        protected override TestRole CreateTestRole(string roleNamePrefix = "", bool useRoleNamePrefixAsRoleName = false)
+        {
+            var roleName = useRoleNamePrefixAsRoleName ? roleNamePrefix : string.Format("{0}{1}", roleNamePrefix, Guid.NewGuid());
+            return new TestRole(roleName);
+        }
+
+        protected override Expression<Func<TestUser, bool>> UserNameEqualsPredicate(string userName) => u => u.UserName == userName;
+
+        protected override Expression<Func<TestRole, bool>> RoleNameEqualsPredicate(string roleName) => r => r.Name == roleName;
+
+        protected override Expression<Func<TestUser, bool>> UserNameStartsWithPredicate(string userName) => u => u.UserName.StartsWith(userName);
+
+        protected override Expression<Func<TestRole, bool>> RoleNameStartsWithPredicate(string roleName) => r => r.Name.StartsWith(roleName);
     }
 }
--- a/test/Shared/UserManagerTestBase.cs
+++ b/test/Shared/UserManagerTestBase.cs
@@ -596,7 +596,7 @@ namespace Microsoft.AspNet.Identity.Test
             var mgr = CreateManager();
             if (mgr.SupportsQueryableUsers)
             {
-                var users = GenerateUsers("CanFindUsersViaUserQuerable", 3);
+                var users = GenerateUsers("CanFindUsersViaUserQuerable", 4);
                 foreach (var u in users)
                 {
                     IdentityResultAssert.IsSuccess(await mgr.CreateAsync(u));
@@ -1105,8 +1105,7 @@ namespace Microsoft.AspNet.Identity.Test
                     IdentityResultAssert.IsSuccess(await manager.CreateAsync(r));
                 }
                 Assert.Equal(roles.Count, manager.Roles.Count(RoleNameStartsWithPredicate("CanQuerableRolesTest")));
-                var r1 = manager.Roles.FirstOrDefault(RoleNameStartsWithPredicate("CanQuerableRolesTest1"));
-                Assert.Equal(roles[1], r1);
+                Assert.Null(manager.Roles.FirstOrDefault(RoleNameEqualsPredicate("bogus")));
             }
         }
 
--- a/test/Shared/UserManagerTestBase.cs
+++ b/test/Shared/UserManagerTestBase.cs
@@ -11,18 +11,19 @@ using Microsoft.AspNet.Testing;
 using Microsoft.Framework.DependencyInjection;
 using Microsoft.Framework.Logging;
 using Xunit;
+using System.Linq.Expressions;
 
 namespace Microsoft.AspNet.Identity.Test
 {
     // Common functionality tests that all verifies user manager functionality regardless of store implementation
     public abstract class UserManagerTestBase<TUser, TRole> : UserManagerTestBase<TUser, TRole, string>
-        where TUser : IdentityUser, new()
-        where TRole : IdentityRole, new()
+        where TUser : class
+        where TRole : class
     { }
 
     public abstract class UserManagerTestBase<TUser, TRole, TKey>
-        where TUser : IdentityUser<TKey>, new()
-        where TRole : IdentityRole<TKey>, new()
+        where TUser : class
+        where TRole : class
         where TKey : IEquatable<TKey>
     {
         protected TestLoggerFactory loggerFactory;
@@ -82,15 +83,18 @@ namespace Microsoft.AspNet.Identity.Test
         protected abstract void AddUserStore(IServiceCollection services, object context = null);
         protected abstract void AddRoleStore(IServiceCollection services, object context = null);
 
-        protected TUser CreateTestUser(string namePrefix = "")
-        {
-            return new TUser() { UserName = namePrefix + Guid.NewGuid().ToString() };
-        }
+        protected abstract void SetUserPasswordHash(TUser user, string hashedPassword);
 
-        protected TRole CreateRole(string namePrefix = "")
-        {
-            return new TRole() { Name = namePrefix + Guid.NewGuid().ToString() };
-        }
+        protected abstract TUser CreateTestUser(string namePrefix = "", string email = "", string phoneNumber = "",
+            bool lockoutEnabled = false, DateTimeOffset? lockoutEnd = null, bool useNamePrefixAsUserName = false);
+
+        protected abstract TRole CreateTestRole(string roleNamePrefix = "", bool useRoleNamePrefixAsRoleName = false);
+
+        protected abstract Expression<Func<TUser, bool>> UserNameEqualsPredicate(string userName);
+        protected abstract Expression<Func<TUser, bool>> UserNameStartsWithPredicate(string userName);
+
+        protected abstract Expression<Func<TRole, bool>> RoleNameEqualsPredicate(string roleName);
+        protected abstract Expression<Func<TRole, bool>> RoleNameStartsWithPredicate(string roleName);
 
         [Fact]
         public async Task CanDeleteUser()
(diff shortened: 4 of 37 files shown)


HaoK:
Nah, its a bad idea to add EF dependencies, just copy the pocos into the shared test code and rename them for the tests

suhasj:
In doing so, I have replicate the code in UserManagerTestBase for InMemory.Test project since it uses the IdentityUser object and that is a lot of duplication. Is that the right way ?

HaoK:
There should be no replication other than copying the pocos from Core to the Shared Test code along side user manager test base with a new name for the pocos

suhasj:
Ok let me see if the amount of duplicated code will be difficult to maintain in the future. I can reason out accordingly after that

suhasj:
Now if I duplicate the POCOs in the Shared folder and make UserManagerTestBase use them, the SqlStoreBase complains since it uses UserManagerTestBase with IdentityUser from Identity.EF. If I make all the test projects to use the POCOs in the shared folder, then we need to maintain them along with the POCOs in the source. My concern is that if we forget to update them then the tests will pass incorrectly.

HaoK:
Or maybe just move the IdentityUser classes to the new shared code stuff, so both EF and tests can depend on it safely.

HaoK:
Another route could be to look at what the tests are doing to access the pocos directly and instead replace them with calls to identity, which might let you remove the IdentityUser constraint entirely

suhasj:
Sure the second option might be an easy option. Let me try that out

suhasj:
@HaoK @divega 

Updated with changes
- UserManagerTestBase now is agnostic to user model and gets all attributes using manager apis. 
- Identity POCOs in Shared folder to be used by InMemory and Core tests
- updated samples

Question
- New extension method on ServiceCollection `AddIdentityWithTestPocos` which is used only in test because it uses the Shared Pocos while the one is source uses EF ones. Instead of this should i move the IdentityBuilderTests and IdentityOptionsTests to EF ? I would want to go with move option

src/Microsoft.AspNet.Identity.EntityFramework/IdentityServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+using System;
+using Microsoft.AspNet.Identity;
+using Microsoft.AspNet.Identity.EntityFramework;
+using Microsoft.Framework.ConfigurationModel;
+
+namespace Microsoft.Framework.DependencyInjection
+{
+    public static class IdentityServiceCollectionExtensions
+    {
+        public static IdentityBuilder AddIdentity(this IServiceCollection services)
```
divega: Interesting that these overload now moves here because the typed moved and hence they are tied to EF. I think it would probably better to get rid of them.

src/Microsoft.AspNet.Identity.EntityFramework/IdentityServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            return services.AddIdentity<IdentityUser, IdentityRole>();
+        }
+
+        public static IdentityBuilder AddIdentity(
```
divega: Same as comment above.

src/Microsoft.AspNet.Identity.EntityFramework/IdentityServiceCollectionExtensions.cs
suhasj: Fixed and updated

suhasj:
Also one important feedback i need to getting username synchronously inside a Linq queries. For example 
`mgr.GetUserNameAsync(u).Result` in https://github.com/aspnet/Identity/pull/378/files#diff-bc862db146cf4a686c23d4fb0676c99eR584. The reason for doing this is that the Linq query cannot be async. Let me know if this is something that is not recommended. 

The workaround is to have a sync method on the `UserManagerTestBase` class which returns different user attributes. For example `GetUserId(user)`. This is abstract on the base and is implemented by the concrete test classes

test/Shared/IdentityUser.cs
```diff
@@ -87,5 +88,20 @@ public IdentityUser(string userName) : this()
         ///     Used to record failures for the purposes of lockout
         /// </summary>
         public virtual int AccessFailedCount { get; set; }
+
+        /// <summary>
+        ///     Navigation property for users in the role
+        /// </summary>
+        public virtual ICollection<IdentityUserRole<TKey>> Roles { get; private set; } = new List<IdentityUserRole<TKey>>();
```
divega: Will the `private` create issues with reflection in Mono?

divega:
> mgr.GetUserNameAsync(u).Result in https://github.com/aspnet/Identity/pull/378/files#diff-bc862db146cf4a686c23d4fb0676c99eR584. The reason for doing this is that the Linq query cannot be async. Let me know if this is something that is not recommended.

Can't you do this?
``` C#
var userName = await mgr.GetUserNameAsync(u);
Assert.Null(mgr.Users.FirstOrDefault(u => userName == "bogus"));
```

divega:
> New extension method on ServiceCollection AddIdentityWithTestPocos which is used only in test because it uses the Shared Pocos while the one is source uses EF ones. Instead of this should i move the IdentityBuilderTests and IdentityOptionsTests to EF ? I would want to go with move option

I don't understand this question, and I couldn't find the method right now. Did you already remove it? Were these test be independent of the provider?

test/Microsoft.AspNet.Identity.EntityFramework.Test/DefaultPocoTest.cs
```diff
@@ -65,6 +67,152 @@ public void DropDatabaseStart()
             IdentityResultAssert.IsSuccess(await userManager.DeleteAsync(user));
         }
 
+        [Fact]
```
HaoK: Are these new tests or copied tests?

test/Microsoft.AspNet.Identity.InMemory.Test/InMemoryStoreTest.cs
```diff
@@ -23,5 +23,31 @@ protected override void AddRoleStore(IServiceCollection services, object context
         {
             services.AddSingleton<IRoleStore<IdentityRole>, InMemoryRoleStore<IdentityRole>>();
         }
+
+        protected override InMemoryUser CreateTestUser(string username = "", string email = "",
```
HaoK: Isn't InMemoryUser the same as the new shared test user now?  Can we switch this over to use the common POCO?

test/Shared/IdentityUser.cs
HaoK: Why do we need navigation properties on the shared test pocos?

test/Shared/UserManagerTestBase.cs
```diff
@@ -83,15 +83,10 @@ protected RoleManager<TRole> CreateRoleManager(object context = null, IServiceCo
         protected abstract void AddUserStore(IServiceCollection services, object context = null);
         protected abstract void AddRoleStore(IServiceCollection services, object context = null);
 
-        protected TUser CreateTestUser(string namePrefix = "")
-        {
-            return new TUser() { UserName = namePrefix + Guid.NewGuid().ToString() };
-        }
+        protected abstract TUser CreateTestUser(string username = "", string email = "", string phoneNumber = "",
+            bool lockoutEnabled = false, DateTimeOffset? lockoutEnd = null);
```
HaoK: I would go back to the namePrefix instead of allowing hardcoded user names, those almost certainly will cause issues accidentally if multiple tests pick the same user name

test/Shared/UserManagerTestBase.cs
```diff
@@ -99,12 +94,12 @@ protected TRole CreateRole(string namePrefix = "")
...
-            IdentityResultAssert.VerifyUserManagerSuccessLog(manager.Logger, "CreateAsync", user.Id.ToString());
+            IdentityResultAssert.VerifyUserManagerSuccessLog(manager.Logger, "CreateAsync", await manager.GetUserIdAsync(user));
 
             IdentityResultAssert.IsSuccess(await manager.DeleteAsync(user));
-            IdentityResultAssert.VerifyUserManagerSuccessLog(manager.Logger, "DeleteAsync", user.Id.ToString());
+            IdentityResultAssert.VerifyUserManagerSuccessLog(manager.Logger, "DeleteAsync", await manager.GetUserIdAsync(user));
 
-            Assert.Null(await manager.FindByIdAsync(user.Id.ToString()));
+            Assert.Null(await manager.FindByIdAsync(await manager.GetUserIdAsync(user)));
         }
```
HaoK: nit: Get the user id once and store it

test/Shared/UserManagerTestBase.cs
```diff
@@ -139,18 +134,17 @@ protected TRole CreateRole(string namePrefix = "")
...
 
-            user.PasswordHash = manager.PasswordHasher.HashPassword(user, "New");
-            IdentityResultAssert.IsSuccess(await manager.UpdateAsync(user));
+            IdentityResultAssert.IsSuccess(await manager.ChangePasswordAsync(user, "password", "Password1!"));
             Assert.False(await manager.CheckPasswordAsync(user, "password"));
-            expectedLog = string.Format("{0} for user: {1} : {2}", "CheckPasswordAsync", user.Id.ToString(), false.ToString());
+            expectedLog = string.Format("{0} for user: {1} : {2}", "CheckPasswordAsync", await manager.GetUserIdAsync(user), false.ToString());
             IdentityResultAssert.VerifyLogMessage(manager.Logger, expectedLog);
-            Assert.True(await manager.CheckPasswordAsync(user, "New"));
+            Assert.True(await manager.CheckPasswordAsync(user, "Password1!"));
```
HaoK: You accidentally broke this test, you deleted the HashPassword line above, password should get updated to New

test/Shared/UserManagerTestBase.cs
```diff
@@ -506,27 +499,25 @@ protected TRole CreateRole(string namePrefix = "")
...
         }
 
         [Fact]
         public async Task AddDupeUserNameFails()
         {
             var manager = CreateManager();
-            var user = CreateTestUser();
-            var user2 = new TUser() { UserName = user.UserName };
+            var user = CreateTestUser("AddDupeUserNameFails");
+            var user2 = CreateTestUser("AddDupeUserNameFails");
```
HaoK: So if CreateTestUser only accepts name prefixes, just use SetuserName to explicitly set it to a fixed name after creating the user.

test/Shared/IdentityUser.cs
suhasj: True but I wanted to have parity with the Pocos in the source.

@divega for the private thing, the issue is fixed in the latest version of Mono. But should we revert it to protected ? Do we foresee users overriding this property considering it is a navigation property ?

HaoK: Its not about parity, the EF pocos are EF specific, if the navigation properties are not needed for the tests, they shouldn't be there.

test/Microsoft.AspNet.Identity.InMemory.Test/InMemoryStoreTest.cs
suhasj: Yeah we could do that. I'll InMemory user altogether. There is a TestUser also I believe. Can I remove that too ?

HaoK: I would merge the new pocos with TestUser and keep the TestUser name as opposed to the confusing IdentityUser name

test/Shared/UserManagerTestBase.cs
suhasj: I kind of liked this that took the parameter as username and appended Guid when calling this. I feel strongly about keeping it this way since it more cleaner. 

Do you feel we are creating a pit of failure here for users ? They can always refer to test code for usage ?

test/Shared/UserManagerTestBase.cs
suhasj: No I replaced the `New` with `Password1!` since it doesn't meet the password requirement. Instead of directly setting the password hash I am doing it via `ChangePasswordAsync`

test/Shared/UserManagerTestBase.cs
HaoK: This is test code not user code, you are creating a pit of failure for us, as if we write two tests that happen to choose the same user name now, they will fail/collide/ do bad things since the user name will exist.

test/Microsoft.AspNet.Identity.EntityFramework.Test/DefaultPocoTest.cs
suhasj: These are new ones I added to test if navigation properties work when used with `Include` on dbContext

test/Shared/UserManagerTestBase.cs
suhasj: good point

test/Shared/UserManagerTestBase.cs
HaoK: Read the test name more closely then :): "CanUpdatePasswordUsingHasher"

test/Shared/UserManagerTestBase.cs
suhasj: ok i'll change it if I revert to nameprefix :)

suhasj:
@divega Using a variable ! so obvious. To answer the second question, yes the class no longer exists to it is fine

test/Microsoft.AspNet.Identity.InMemory.Test/InMemoryStoreTest.cs
suhasj: sure

test/Shared/UserManagerTestBase.cs
suhasj: This is interesting, since the user object is now generic, we need to either have a SetPasswordHash on UserManager or abstract method on UserManagerTestBase to set it. I think I'll add it on the test base class. SetPasswordHash on UserManager seems not good

HaoK: I think all of these methods showing up on the test base smells a bit, a better alternative might be to create a shared ITestUser internal interface and have both TestUser and IdentityUser implement this, then you can revert all of the new async manager calls and just use the old properties again since we can constrain against ITestUser now.

test/Shared/IdentityUser.cs
suhasj: so the navigation properties are needed since the InMemory tests makes use of them. I need to keep them

test/Shared/UserManagerTestBase.cs
suhasj: I can't have this constraint on the UserManagerTestBase generic since EF implementation IdentityUser would need to implement this interface as well. If we have the interface in test base, then in the EF tests, I need to have a TestUser which implements ITestUser and EF.IdentityUser but that doesn't seem that clean

HaoK: What's wrong with having the real users implement an internal interface?

HaoK: or even better, you can put the POCOs in internal shared code, and have the EF one extend and just make it public

internal class IdentityUserBase
public class EF.IdentityUser : IdentityUserBase { }

suhasj: I like this solution. I'll have IdentityUserBase internal in Identity project all except the navigation properties. EF.IdentityUser extends this to add navigation properties and TestUser extends this to add properties as needed.

There is also an ApplicationUser in the test shared folder. Can I replace that too ?

divega: If I am reading this correctly, it sounds very complicated, and it won't work :)

http://blogs.msdn.com/b/ericlippert/archive/2012/11/13/why-is-deriving-a-public-class-from-an-internal-class-illegal.aspx

I think it is actually fine to use the API on user manager as an abstraction layer and if we just need to do some trick to set the password hash alone, we can look into that more.


-------- Original message --------

divega: >abstract method on UserManagerTestBase 

I like that more, personally. It means you can potentially derive from this class to test any provider without spoiling the entity types.

suhasj:
@divega Now I recall that I did not use the variable since i need to use the object inside the `Where` clause. So `var userName = await mgr.GetUserNameAsync(u);` would not work since `u` is to be got from the lambda. The workaround I see is to use a foreach loop instead of `Where` which is not clean either.

test/Shared/UserManagerTestBase.cs
suhasj: Going by the article we need to have IdentityUserBase public as well. Whether we do it just to unblock our tests, I am not sure. I feel we need a wider discussion to have a public user base class on Identity core. I'll go with the abstract method on UserManagerTestBase approach for now

divega:
Ah, I missed that. Sorry.

test/Shared/UserManagerTestBase.cs
suhasj: There are some tests that need to remember the usernames for the next step, for example
```
var user = CreateUser(username)
result = await manager.FindByUserName(username)
```

For the above scenario I need to have the username at hand. Changing to namePrefix, it makes it tough to maintain the username. Earlier this was done using user.Username so it back to the same problem again. Can the overload take both username and flag that if true, uses what is passed ?

suhasj:
Updated with feedback
* Removed InMemoryUser and ApplicationUser. Identity.Test and IdentityInMemory.Test refer to TestUser only
* GetPasswordHash an abstract method on UserManagerTestBase
* CreateTestUser now takes namePrefix instead of username
* CreateTestRole does not take a namePrefix since most test need a expected rolename for further assertions.

suhasj:
@HaoK @divega 

Removing the feedback tag. I have addressed all feedback so wanted to make sure it is ready to check in

src/Microsoft.AspNet.Identity.EntityFramework/IdentityUser.cs
```diff
@@ -87,5 +88,20 @@ public IdentityUser(string userName) : this()
         ///     Used to record failures for the purposes of lockout
         /// </summary>
         public virtual int AccessFailedCount { get; set; }
+
```
HaoK: You should double check with @divega but I'm pretty sure EF needs these nav properties to be public get/set to do their proxy stuff...?

test/Shared/IdentityRoleClaim.cs
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNet.Identity.Test
+{
+    public class TestRoleClaim : TestRoleClaim<string> { }
```
HaoK: Fix the file names for these shared files to match their classes

suhasj: ah rebasing messed it up. I'll fix and verify the entire PR again

src/Microsoft.AspNet.Identity.EntityFramework/IdentityUser.cs
suhasj: All the properties now have public setters. I'll change these to public as well then

divega:
I'll take another look later today.

HaoK:
You might want to rebase since I added some functionals in InMemory yesterday, but otherwise looks :shipit: to me

suhasj:
@divega pinging a reminder

src/Microsoft.AspNet.Identity.EntityFramework/IdentityUser.cs
divega: We actually don't have proxies in EF7 so far. @ajcvickers what do you think the best possible pattern would be for these collection navprops?

test/Microsoft.AspNet.Identity.EntityFramework.Test/DefaultPocoTest.cs
```diff
@@ -63,6 +65,152 @@ public void DropDatabaseStart()
...
+            var builder = new ApplicationBuilder(CallContextServiceLocator.Locator.ServiceProvider);
+
+            builder.UseServices(services =>
+            {
+                DbUtil.ConfigureDbServices<IdentityDbContext>(ConnectionString, services);
+                services.AddIdentity<IdentityUser, IdentityRole>().AddEntityFrameworkStores<IdentityDbContext>();
+            });
+
+            var userManager = builder.ApplicationServices.GetRequiredService<UserManager<IdentityUser>>();
+            var dbContext = builder.ApplicationServices.GetRequiredService<IdentityDbContext>();
```
divega: DbContext is IDisposable. Please use `using` even if doesn't seem needed right now.

divega: Sorry, scratch that. I see now that you don't control the lifetime of this one.

divega: However this still doesn't seem right. When will the context get disposed? I assume you copied this pattern from other tests?

test/Shared/UserManagerTestBase.cs
```diff
@@ -592,8 +591,8 @@ protected TRole CreateRole(string namePrefix = "")
             {
                 IdentityResultAssert.IsSuccess(await mgr.CreateAsync(u));
             }
-            var usersQ = mgr.Users.Where(u => u.UserName.StartsWith("CanFindUsersViaUserQuerable"));
-            Assert.Null(mgr.Users.FirstOrDefault(u => u.UserName == "bogus"));
+            var usersQ = mgr.Users.Where(u => mgr.GetUserNameAsync(u).Result.StartsWith("CanFindUsersViaUserQuerable"));
+            Assert.Null(mgr.Users.FirstOrDefault(u => mgr.GetUserNameAsync(u).Result == "bogus"));
```
divega: This one still seems weird. Using `Task<T>.Result` is pretty bad, and also this expression cannot be translated to a store query in EF so it will evaluate in memory. Isn't this actually an EF specific test that could depend on the EF POCOs? If we want to have it for any provider that implements IQueryable, let's chat about how to make it more general.

test/Shared/UserManagerTestBase.cs
```diff
@@ -1096,34 +1091,36 @@ public Task<IdentityResult> ValidateAsync(UserManager<TUser> manager, TUser user
             {
                 IdentityResultAssert.IsSuccess(await manager.CreateAsync(r));
             }
-            Assert.Equal(roles.Count, manager.Roles.Count(r => r.Name.StartsWith("CanQuerableRolesTest")));
-            var r1 = manager.Roles.FirstOrDefault(r => r.Name.StartsWith("CanQuerableRolesTest1"));
+            Assert.Equal(roles.Count, manager.Roles.Count(r => manager.GetRoleNameAsync(r).Result.StartsWith("CanQuerableRolesTest")));
+            var r1 = manager.Roles.FirstOrDefault(r => manager.GetRoleNameAsync(r).Result.StartsWith("CanQuerableRolesTest1"));
```
divega: Same comment as before about .Result. We should re-think how we test `IQueryables`.

test/Shared/UserManagerTestBase.cs
```diff
@@ -1208,7 +1205,7 @@ public Task<IdentityResult> ValidateAsync(UserManager<TUser> manager, TUser user
                 Assert.Equal(roles.Count, rs.Count);
                 foreach (var r in roles)
                 {
-                    Assert.True(rs.Any(role => role == r.Name));
+                    Assert.True(rs.Any(role => role == roleManager.GetRoleNameAsync(r).Result));
```
divega: See other comments about `.Result`. Let's figure out if we can something better.

test/Shared/UserManagerTestBase.cs
```diff
@@ -1633,14 +1635,14 @@ public Task<IdentityResult> ValidateAsync(UserManager<TUser> manager, TUser user
 
                 if ((i % 2) == 0)
                 {
-                    IdentityResultAssert.IsSuccess(await manager.AddToRolesAsync(user, roles.Select(x => x.Name).AsEnumerable()));
-                    IdentityResultAssert.VerifyUserManagerSuccessLog(manager.Logger, "AddToRolesAsync", user.Id.ToString());
+                    IdentityResultAssert.IsSuccess(await manager.AddToRolesAsync(user, roles.Select(x => roleManager.GetRoleNameAsync(x).Result).AsEnumerable()));
```
divega: Same about `.Result`.

test/Microsoft.AspNet.Identity.EntityFramework.Test/DefaultPocoTest.cs
suhasj: I am assuming you are worried how the database is cleared ? There is a test `DropDatabaseStart` that is marked with lowest priority that will delete the database after all the tests in the class have run. This class existed all the way long. I have only added more tests that's all. Regarding the object, since it is created in the scope of the test, then it should be disposed at the end of it. If adding a using statement for DBContext is the recommended practice then sure I can update that but will have to do it in other tests as well

test/Shared/UserManagerTestBase.cs
suhasj: sure. The `.Result` was something i din't like too but ran out of ideas to make it better since the variable was inside the lambda. We can chat about it.

divega: If we really think these test are general purpose and not EF only, e.g. if we think they have value for store providers that are not based on EF but support IQueryable, I propose that:

1. The tests for the IQueryable interfaces shouldn't fail if the IQueryable properties are not supported by the store, e.g. the test should test for the capability first.

2. The LINQ implementations are built to recognize simple patterns such as property accesses on the entities, and not calls to Identity. Since the test base class cannot depend on a specific set of entity types, we need to abstract this. We can do it easily by adding a set of virtual methods can invoke to build the expression that needs to be passed to the LINQ operators, e.g. for this case (brain-compiled):

``` C#
protected abstract Expression<Func<TUser, bool>> GetUserNamePredicate(string userName) 
{
}
``` 

An implementation for EF would be
``` C#
protected abstract Expression<Func<IdentityUser, bool>> GetUserNamePredicate(string userName) 
{
    return u => u.NormalizedUserName == userName;
}
```

suhasj: This seems fair. I agree with the first point. For the second I want to minimize the number of abstract methods we have to fix this. I hope it would be only GetUserNamePredicate and GetRoleNamePredicate

suhasj: For the first point, we should pass if the IQueryable is no implemented but rather skip it. I think it makes sense to have a custom xunit test attribute to skip the test. I'll prototype that but let me know if you guys think other wise

HaoK: Eh?  Overkill, just check whether the store supports IQueryable and no-op pass the test if not...

HaoK: I'm honestly not even sure we should build this functionality in at this point, given that we have no such stores :)

suhasj: My thought was the user shouldn't get a false sense that the test is passing but in the skip attribute I needed access to the UserManager object which lead to generic hell. I think noop is the simplest option here

divega: Since it seems we are going to keep the `IQueryable` properties it seems better to keep these tests here. I agree we should only have a few predicate building methods like this, but the pattern isn't bad.

test/Microsoft.AspNet.Identity.EntityFramework.Test/DefaultPocoTest.cs
divega: I am more worried about the risk of tests leaking resources, e.g. connections, or even memory depending how (well) the DI container implementation works and what we do in the tests.

suhasj:
Rebased with latest and remove all instances of `.Result` in the UserManagerTestBase. Added the predicate to compare user/roles names. Needed to add a contains flag since has two asserts that needed to test part of user/roles names. 

Also noop for IQueryable tests is the managers do not support them

test/Shared/UserManagerTestBase.cs
```diff
@@ -82,15 +83,16 @@ protected RoleManager<TRole> CreateRoleManager(object context = null, IServiceCo
...
-        protected TRole CreateRole(string namePrefix = "")
-        {
-            return new TRole() { Name = namePrefix + Guid.NewGuid().ToString() };
-        }
+        protected abstract TUser CreateTestUser(string namePrefix = "", string email = "", string phoneNumber = "",
+            bool lockoutEnabled = false, DateTimeOffset? lockoutEnd = null, bool useNamePrefixAsUserName = false);
+
+        protected abstract TRole CreateTestRole(string roleName = "");
+
+        protected abstract Expression<Func<TUser, bool>> UserNamePredicate(string userName, bool contains = false);
```
divega: 1. I would rather simplify the tests to just succeed with simple equality and get rid of the contains flag on both predicates.
2. Contains is not the same as StartsWith.

divega: Other than that, I think we are very close.

test/Shared/UserManagerTestBase.cs
```diff
@@ -505,43 +513,40 @@ protected TRole CreateRole(string namePrefix = "")
...
             manager.Options.User.RequireUniqueEmail = true;
-            var user = CreateTestUser();
-            var user2 = CreateTestUser();
-            string email = user.UserName + "@yup.com";
-            user.Email = email;
-            user2.Email = email;
+            var user = CreateTestUser(email: "FooUser@yup.com");
+            var user2 = CreateTestUser(email: "FooUser@yup.com");
             IdentityResultAssert.IsSuccess(await manager.CreateAsync(user));
-            IdentityResultAssert.IsFailure(await manager.CreateAsync(user2), IdentityErrorDescriber.Default.DuplicateEmail(user.Email));
```
HaoK: Why aren't we checking the failure code anymore?

test/Shared/UserManagerTestBase.cs
```diff
@@ -568,17 +573,16 @@ protected TRole CreateRole(string namePrefix = "")
...
             IdentityResultAssert.IsFailure(result, IdentityErrorDescriber.Default.LoginAlreadyAssociated());
-            IdentityResultAssert.VerifyUserManagerFailureLog(manager.Logger, "AddLoginAsync", user.Id.ToString(), IdentityErrorDescriber.Default.LoginAlreadyAssociated());
+            IdentityResultAssert.VerifyUserManagerFailureLog(manager.Logger, "AddLoginAsync", await manager.GetUserIdAsync(user), IdentityErrorDescriber.Default.LoginAlreadyAssociated());
         }
 
         // Email tests
         [Fact]
         public async Task CanFindByEmail()
         {
+            var email = "foouser@test.com";
```
HaoK: This is broken because you hardcoded an email

test/Shared/UserManagerTestBase.cs
```diff
@@ -1050,41 +1051,43 @@ public Task<IdentityResult> ValidateAsync(UserManager<TUser> manager, TUser user
...
+            IdentityResultAssert.VerifyRoleManagerSuccessLog(manager.Logger, "RemoveClaimAsync", await manager.GetRoleIdAsync(role));
         }
 
         [Fact]
         public async Task CanRoleFindById()
         {
             var manager = CreateRoleManager();
-            var role = CreateRole("FindByIdAsync");
-            Assert.Null(await manager.FindByIdAsync(role.Id.ToString()));
+            var role = CreateTestRole("FindByIdAsync");
```
HaoK: Another hardcoded name (this is why I think we should put the name randomization inside of the helper rather than expecting people to remember to do this...

test/Shared/UserManagerTestBase.cs
```diff
@@ -1093,39 +1096,45 @@ public Task<IdentityResult> ValidateAsync(UserManager<TUser> manager, TUser user
         public async Task CanQueryableRoles()
         {
             var manager = CreateRoleManager();
-            var roles = GenerateRoles("CanQuerableRolesTest", 4);
-            foreach (var r in roles)
+            if (!manager.SupportsQueryableRoles)
```
HaoK: Isn't this backwards?  When would this test ever run today?

test/Shared/UserManagerTestBase.cs
suhasj: The requireuniqueemail flag is false so this is passing with all tests. We can have a random number with the email and set that flag to true

test/Shared/UserManagerTestBase.cs
```diff
@@ -1165,19 +1174,19 @@ public Task<IdentityResult> ValidateAsync(UserManager<TUser> manager, TUser user
             var context = CreateTestContext();
             var manager = CreateManager(context);
             var roleManager = CreateRoleManager(context);
-            var role = CreateRole("addUserTest");
+            var roleName = "AddUserTest" + Guid.NewGuid().ToString();
```
HaoK: Yeah these role + guid name lines are all over, move this into the helper just like CreateTestUser

test/Shared/UserManagerTestBase.cs
HaoK: Yeah Contains != StartsWith, should be StartsWith

HaoK: We can't simplify the tests to just do equality since we aren't cleaning up the database, so a lot of the tests use the same username prefix, like "test", and we only drop/recreate the database once per run.  Rather than trying to manually maintain user name uniqueness, I went with the approach of randomizing the names instead by always jamming on a guid, and then doing StratsWith instead of strict equality.

test/Shared/UserManagerTestBase.cs
suhasj: we do add randomization in the helper when there is no argument provide. When something is, then we consider it to be fixed since they might need to compare that value later. I agree though that I missed it in this method

test/Shared/UserManagerTestBase.cs
suhasj: I fixed this for queryable users missed this for roles. Was doing some refactoring using Resharper but thanks for catching

test/Shared/UserManagerTestBase.cs
suhasj: thanks for catching. Copy paste error

test/Shared/UserManagerTestBase.cs
HaoK: Also it doesn't make sense for CreateTestUser and CreateTestRole not to look the same, for the same reason CreateTestUser's was done the way it was, CreateTestRole should follow the same pattern

test/Shared/UserManagerTestBase.cs
suhasj: We were doing `name + guid` so I presumed Contains would still be valid check instead of StartWith considering we are looking if part of the string is present. I'll revert back to StartsWith it is just that I have to now maintain the original randomized username and pass it exactly to the predicate

divega: I am ok with having StartsWith checks if we can't have just equality. But let's then split this in 4 simple predicate methods without any flags: UserNameEqualsPredicate, UserNameStartsWithPredicate, RoleNameEqualsPredicate, RoleNameStartsWithPredicate.

test/Shared/UserManagerTestBase.cs
divega: Sounds good to move this to the helper. If we need to reason about the generated value in a entity type-agnostic way, the helper can have output parameters for that.

test/Shared/UserManagerTestBase.cs
suhasj: Added the four predicates and restored the test to original assert statements

test/Shared/UserManagerTestBase.cs
suhasj: fixed. Both have the same pattern

test/Shared/UserManagerTestBase.cs
suhasj: Hao was ok with having a flag in the method that treats supplied prefix as the actual value else we append the guid. The default is to append the guid. There a few tests that need to do it and this approach is simpler than having an `out` parameter. You can look at the api and let me know your thoughts

suhasj:
Updated with 2 additional predicates to compare StartsWith for user/role name
* Some changes you might be interested in [predicates] (https://github.com/aspnet/Identity/pull/378/files#diff-bc862db146cf4a686c23d4fb0676c99eR94)
* queryable tests [1](https://github.com/aspnet/Identity/pull/378/files#diff-bc862db146cf4a686c23d4fb0676c99eR594) and [2] (https://github.com/aspnet/Identity/pull/378/files#diff-bc862db146cf4a686c23d4fb0676c99eR1097)

@ajcvickers made changes to EF for mapping. Should I rebase and send another PR to see if that works with these changes ?

suhasj:
I rebased it to latest EF which had renaming for mapping in OnModel creating. I have updated the IdentityDbContext with the latest changes

@ajcvickers can you please specifically review changes to [IdentityDbContext](https://github.com/aspnet/Identity/pull/378/files#diff-054beeacf6ff613faafca0aa8c1ad759R34)

test/Shared/UserManagerTestBase.cs
```diff
@@ -1093,39 +1097,44 @@ public Task<IdentityResult> ValidateAsync(UserManager<TUser> manager, TUser user
...
             {
-                IdentityResultAssert.IsSuccess(await manager.CreateAsync(r));
+                var roles = GenerateRoles("CanQuerableRolesTest", 4);
+                foreach (var r in roles)
+                {
+                    IdentityResultAssert.IsSuccess(await manager.CreateAsync(r));
+                }
+                Assert.Equal(roles.Count, manager.Roles.Count(RoleNameStartsWithPredicate("CanQuerableRolesTest")));
+                var r1 = manager.Roles.FirstOrDefault(RoleNameStartsWithPredicate("CanQuerableRolesTest1"));
+                Assert.Equal(roles[1], r1);
```
divega: Can we have a negativeassert as in the Users test and standardize on the same test data generation pattern?

test/Shared/UserManagerTestBase.cs
divega: Yeah sure, I guess we don't need to reason about the generated values in the test, which is fine.

test/Shared/UserManagerTestBase.cs
suhasj: sure I updated it and role name are done the same way as usernames.

src/Microsoft.AspNet.Identity.EntityFramework/IdentityUser.cs
ajcvickers: The way they are will work fine, but they don't need to be virtual (although you may want them to be virtual for other reasons) and they don't need to have a setter.

ajcvickers:
IdentityDbContext changes look fine.

src/Microsoft.AspNet.Identity.EntityFramework/IdentityUser.cs
divega: Thanks @ajcvickers. The setter on collection nav props actually looks very weird to me. @suhasj can we remove it?

divega:
:shipit: to get it on beta4. We can address any outstanding test issues later.

suhasj:
thanks to everybody for patiently reviewing this

PR #385, opened 2015-03-05
Logging API changes

BrennanConroy:


HaoK:
Changes look fine, :shipit:  Slight tangent, does the Log prefix add much value to the method names, i.e.

logger.Information, logger.Warning, logger.Verbose seems just as clear and reads much nicer...just my 2 c...

PR #397, opened 2015-03-11
Cleanup AddIdentity and project.json

HaoK:
Remove IConfiguration from AddIdentity
Switch to referencing interface assemblies

cc @divega @suhasj


divega:
:shipit:

Issue #395, opened 2015-03-11
Consider moving PrincipalExtensions to ClaimsPrincipal


HaoK:
Might not need to extend IPrincipal anymore, try moving to ClaimsPrincipal saving us from doing the cast checks

HaoK:
Can't do this until MVC fixes https://github.com/aspnet/Mvc/issues/2167

Issue #231, opened 2014-10-17
Add a functional test verifying SecurityStampValidator and middleware


HaoK:
Gap in coverage where no functional test on CI ensures that Security + Cookie + SecurityStampValidator doesn't blow up.

PR #401, opened 2015-03-12
Add initial basic functional tests

HaoK:
Add two initial functional identity tests using InMemory test stores to verify basic cookie/login/expirations.  We should flesh these out and run these on EF as well, but these are roughly what additional tests will look like

cc @divega @suhasj


HaoK:
Fixes https://github.com/aspnet/Identity/issues/395
Fixes https://github.com/aspnet/Identity/issues/231

HaoK:
Merged 2246a41978192770a08c77bab07a80c4d54a0800

PR #404, opened 2015-03-13
Updating test projects to use xunit.runner.aspnet

suhasj:
@divega @HaoK 

Changes
- Updating to use `xunit.runner.aspnet`
- Changing the `GetCustomAttributes` method to use `string` argument instead of `type` since that was causing conflicts due to two assemblies have the same extension method. This change resolves it since that's what the original method is doing internally.


divega:
:shipit:

Issue #398, opened 2015-03-12
Port 2.2 ResetAccessFailed fix


HaoK:

suhasj:
Fixed checked in

PR #407, opened 2015-03-13
Fix reset acces token

suhasj:
@HaoK 

Porting fix from 2.2 Adding test that verifies that store method is not called when the token value is zero. 

https://github.com/aspnet/Identity/issues/398


HaoK:
:shipit:

PR #410, opened 2015-03-17
Added deferred logging with scopes

suhasj:
@HaoK @divega 

This is a feedback PR so omit nitpicks and test files

Changes to include scopes for logging. Some of open thoughts and questions
* I am creating scopes in all public methods of Manager classes that return IdentityResult, SignInResult or bool. This is the base design for logging since we log only if there is an update on the user/role object. 
* Adding scopes to all Manager class results in multiple scopes within Identity layer itself. For example `PasswordSignInAsync` on SignInManager calls `CheckPasswordAsync` on UserManager. So two scopes are being created with the same userId. I am not happy with this experience but I cannot remove the one on UserManager since it is public and can be called separately too. I needed some thoughts on this.
* The scope message being created is via an async method that appends the method name and user/role Id. I have kept this experience for RC and we can change it once we settled upon the final format of scope messages across the framework. Having the scope in a single method let us edit it in a single place. Also since we can have a Dictionary that takes the method name and returns a readable sentence. For example 'PasswordSignInAsync` maps to 'SignIn using password`. This is optional
* Since the scope is always called irrespective of whether the logging is enabled or not, I have chosen to using `string.format` instead of using `Resources` to construct the message. This is also inline with what Eilon mentioned that we can as of now refrain from using Resources. The Result class read from resources though. Let me know if I can remove that as well


src/Microsoft.AspNet.Identity/RoleManager.cs
```diff
@@ -52,6 +57,7 @@ public class RoleManager<TRole> : IDisposable where TRole : class
             }
 
             Logger = logger?.CreateLogger<RoleManager<TRole>>() ?? new Logger<RoleManager<TRole>>(new LoggerFactory());
+            IdentityLogger = new IdentityLogger() { Logger = Logger };
```
HaoK: Why isn't IdentityLogger a service?

src/Microsoft.AspNet.Identity/RoleManager.cs
```diff
@@ -345,8 +364,12 @@ private IRoleClaimStore<TRole> GetClaimStore()
             {
                 throw new ArgumentNullException("role");
             }
-            await claimStore.AddClaimAsync(role, claim, CancellationToken);
-            return await LogResultAsync(await UpdateRoleAsync(role), role);
+
+            using (Logger.BeginScope(await GetScopeMessageforRoleAsync(role)))
```
HaoK: Can we move this creation of the scope to to a private helper, should make life easier if we ever have to tweak the scope

src/Microsoft.AspNet.Identity/SignInManager.cs
```diff
@@ -53,24 +56,17 @@ public class SignInManager<TUser> where TUser : class
         internal IUserClaimsPrincipalFactory<TUser> ClaimsFactory { get; private set; }
         internal IdentityOptions Options { get; private set; }
         internal ILogger Logger { get; set; }
+        protected IdentityLogger IdentityLogger { get; set; }
```
HaoK: Why is IdentityLogger special and protected unlike the other properties?

src/Microsoft.AspNet.Identity/SignInManager.cs
```diff
@@ -53,24 +56,17 @@ public class SignInManager<TUser> where TUser : class
...
-            if (Options.SignIn.RequireConfirmedEmail && !(await UserManager.IsEmailConfirmedAsync(user)))
-            {
-                return await LogResultAsync(false, user);
-            }
-            if (Options.SignIn.RequireConfirmedPhoneNumber && !(await UserManager.IsPhoneNumberConfirmedAsync(user)))
-            {
-                return await LogResultAsync(false, user);
-            }
-            return await LogResultAsync(true, user);
+            var result = !(Options.SignIn.RequireConfirmedEmail && !(await UserManager.IsEmailConfirmedAsync(user))) ||
```
HaoK: This is a pretty gross looking line of code, personally I'm not a fan of compacting things to the point where its impossible to read due to 4 !'s an or, and 2 &'s

src/Microsoft.AspNet.Identity/IdentityLogger.cs
```diff
@@ -0,0 +1,39 @@
...
+            // Check if log level is enabled before creating the message.
+            if (Logger.IsEnabled(logLevel))
+            {
+                Logger.Log(logLevel, 0, messageAccessor(), null, (msg, exp) => (string)msg);
+            }
+
+            return result;
+        }
+
+        public virtual SignInResult LogSignInResult(SignInResult result, [CallerMemberName]string methodName = null)
```
HaoK: can't we just call all of these methods Log?

src/Microsoft.AspNet.Identity/IdentityLogger.cs
```diff
@@ -0,0 +1,39 @@
...
+
+using System;
+using System.Runtime.CompilerServices;
+using Microsoft.Framework.Logging;
+
+namespace Microsoft.AspNet.Identity
+{
+    public class IdentityLogger
+    {
+        public ILogger Logger { get; set; }
```
HaoK: Why is ILogger not a required ctor argument?  Won't everything blow up if its null, there isn't any null checks in the methods currently

HaoK:
Overall I like how it looks much better than before, it looks fairly clean with a consistent using to start, do stuff, and log result.

HaoK:
But can you include a snippet of what the output looks like in this PR?

src/Microsoft.AspNet.Identity/RoleManager.cs
suhasj: Because I need to inject the Logger for the particular class. As you mentioned in the other comment, we should it take it as a mandatory attribute in the constructor

src/Microsoft.AspNet.Identity/RoleManager.cs
suhasj: sure it can hang off IdentityLogger

src/Microsoft.AspNet.Identity/SignInManager.cs
suhasj: I thought it should be accessible to the child class and hence did not make the Logger object mandatory in the constructor. Now looking back it doesn't for the child classes to create their own and I can make it internal like others

src/Microsoft.AspNet.Identity/SignInManager.cs
suhasj: Part of Resharper refactoring. I am fine either way and felt the earlier one was readable for sure

src/Microsoft.AspNet.Identity/IdentityLogger.cs
suhasj: Wanted to differentiate what is being logged since it can be a either IdentityResult or SignInResult. I can change that

src/Microsoft.AspNet.Identity/IdentityLogger.cs
suhasj: I had this and reverted it since i felt the it should be a public set property but I agree. I'll make the change

suhasj:
@HaoK thank you for the feedback. This does look better now. If we hook in the NLog provider, the output is as below for the Identity operations
```
21 RouterMiddleware.Invoke RouteCollection.RouteAsync TemplateRoute.RouteAsync MvcRouteHandler.RouteAsync PasswordSignInAsync for user: da49a8f4-0913-457c-a02a-2334bffe5327 Microsoft.AspNet.Identity.SignInManager<IdentitySample.Models.ApplicationUser> PasswordSignInAsync : Result : Succeeded
```
In the above log, the information until the user Id is the scopes information,  next the class name and then the message. We don't repeat the user id in the log messages since that is obtained from the scope. The format is {method} :{result}

src/Microsoft.AspNet.Identity/SignInManager.cs
suhasj: I changed it to internal and I feel it makes sense to have it protected. In the constructor of the subclass, I can set the Identity logger passing the logger for the subclass which is then reflected in all the log statements, else we have log statements with base class and the sub class which looked weird to me

src/Microsoft.AspNet.Identity/RoleManager.cs
suhasj: Can this hang of IdentityLogger class ? Then that class can be one place where we can edit and would reflect in all there Manager classes instead of editing 3 private helper methods

suhasj:
Updates:
- Moved the `BeginScope` method to `IdentityLogge`r and made it virtual. Creating the scope messages is virtual as well on the manager classes and this gives additional flexibility to configure it at a single place
- The above change resulted in the Logger property no longer needed by the manager class. The `ApplicationXManager` classes can override the IdentityLogger and should be good
- Made the IdentityLogger protected virtual which lets the subclasses create the logger based on that class which makes uniform log messages

suhasj:
@divega I have the tests fixed so will send a final PR if you are ok with this PR

src/Microsoft.AspNet.Identity/RoleManager.cs
```diff
@@ -60,6 +66,11 @@ public class RoleManager<TRole> : IDisposable where TRole : class
         protected IRoleStore<TRole> Store { get; private set; }
 
         /// <summary>
+        ///     Used to create scopes and messages for logging
+        /// </summary>
+        protected internal virtual IdentityLogger IdentityLogger { get; set; }
```
HaoK: Can we just call this Logger?

src/Microsoft.AspNet.Identity/RoleManager.cs
```diff
@@ -162,20 +162,23 @@ public void Dispose()
...
-
             var result = await ValidateRoleInternal(role);
             if (!result.Succeeded)
             {
                 return result;
             }
             await UpdateNormalizedRoleNameAsync(role);
-            return await LogResultAsync(await Store.CreateAsync(role, CancellationToken), role);
+            result = await Store.CreateAsync(role, CancellationToken);
+            using (IdentityLogger.BeginScope(await GetScopeMessageforRoleAsync(role)))
```
HaoK: I was actually suggesting something like:
```
using (await BeginLogScopeAsync(role))
```

src/Microsoft.AspNet.Identity/RoleManager.cs
suhasj: Since it was a wrapper around the ILogger I chose to name it differently but now that object is not there sure, I can rename it to `Logger`

src/Microsoft.AspNet.Identity/RoleManager.cs
suhasj: Sure we already have a overload method in each manager class to create the scope message. I can refactor to have something like this. I added this in the IdentityLogger class so that it can be shared among them. Either way is fine. I'll see make the change and see how it looks

src/Microsoft.AspNet.Identity/RoleManager.cs
suhasj: I chose to keep it IdentityLogger since it contains the ILogger implementation property which is named `Logger` so a call to it would be `Logger.Logger` which now is `IdentityLogger.Logger`. I would have to rename the ILogger to something else but Logger seems to be a good choice for it. Let me know if you feel strongly about it

HaoK: Does anyone actually call Logger.Logger?  The point is that the manager usage looks worse with it called IdentityLogger, that should be the usage pattern you are optimizing for, as its IdentityLogger all over the place right now instead of Logger which is much better

samples/IdentitySample.Mvc/Startup.cs
```diff
@@ -7,6 +7,10 @@
 using Microsoft.Framework.DependencyInjection;
 using Microsoft.Framework.Logging;
 using Microsoft.Framework.Logging.Console;
+#if DNX451
+using NLog.Config;
```
divega: Is the issue you mentioned with NLog support for scopes not being compatible with async still happening? Do you know if there is a bug and anyone following up on that?

src/Microsoft.AspNet.Identity/IdentityLogger.cs
```diff
@@ -0,0 +1,46 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Runtime.CompilerServices;
+using Microsoft.Framework.Logging;
+
+namespace Microsoft.AspNet.Identity
+{
+    public class IdentityLogger
```
divega: Is this class an implementation detail or we expect customers or provider writers to use it? If it is an implementation detail we should consider moving it to a different namespace, e.g. Microsoft.AspNet.Identity.Internal (to be consistent with what we are doing with internal implementation details across the framework) or Microsoft.AspNet.Identity.Infraestructure if we think the API is solid and we don't want to discourage customers from using it directly.

src/Microsoft.AspNet.Identity/IdentityLogger.cs
```diff
@@ -0,0 +1,46 @@
...
+        }
+
+        public virtual SignInResult Log(SignInResult result, [CallerMemberName]string methodName = null)
+           => Log(result, r => r.GetLogLevel(), () => Resources.FormatLoggingSigninResult(methodName, result));
+
+        public virtual IdentityResult Log(IdentityResult result, [CallerMemberName]string methodName = null)
+            => Log(result, r => r.GetLogLevel(), () => Resources.FormatLoggingIdentityResult(methodName, result));
+
+        public virtual bool Log(bool result, [CallerMemberName]string methodName = null)
+            => Log<bool>(result, (b) => b ? LogLevel.Verbose : LogLevel.Warning,
```
divega: Nit: I don't think the `(b)` needs the parenthesis.

src/Microsoft.AspNet.Identity/IdentityLogger.cs
```diff
@@ -0,0 +1,46 @@
...
+    public class IdentityLogger
+    {
+        public IdentityLogger(ILogger logger)
+        {
+            Logger = logger;
+        }
+
+        protected internal virtual ILogger Logger { get; set; }
+
+        public virtual TResult Log<TResult>(TResult result, Func<TResult, LogLevel> getLevel,
```
divega: Nit: if we are not using this method from from outside the class, you can make it protected.

src/Microsoft.AspNet.Identity/IdentityResult.cs
```diff
@@ -51,21 +46,21 @@ public static IdentityResult Failed(params IdentityError[] errors)
...
-            // TODO: Take logging level as a parameter
-            if (Succeeded)
-            {
-                logger.LogInformation(Resources.FormatLogIdentityResultSuccess(message));
-            }
-            else
-            {
-                logger.LogWarning(Resources.FormatLogIdentityResultFailure(message, string.Join(",", Errors.Select(x => x.Code).ToList())));
-            }
+            return Succeeded ? "Succeeded" : string.Format("{0} : {1}", "Failed", string.Join(",", Errors.Select(x => x.Code).ToList()));
```
divega: Nit: Consider breaking in three lines on `?` and `:`.

src/Microsoft.AspNet.Identity/IdentityResult.cs
```diff
@@ -51,21 +46,21 @@ public static IdentityResult Failed(params IdentityError[] errors)
...
+        }
+
+        /// <summary>
+        ///     Get the level to to log this result at
+        /// </summary>
+        /// <returns></returns>
+        public virtual LogLevel GetLogLevel()
+        {
+            return Succeeded ? LogLevel.Verbose : LogLevel.Warning;
```
divega: Is there anything we should be logging as information? I think we should log as information when we write changes to the Identity store (I going down from the top of the PR, so maybe I haven't got to it yet?).

src/Microsoft.AspNet.Identity/RoleManager.cs
```diff
@@ -20,13 +21,17 @@ namespace Microsoft.AspNet.Identity
...
 
         /// <summary>
-        /// Constructor
+        ///     Constructor
         /// </summary>
         /// <param name="store">The IRoleStore commits changes via the UpdateAsync/CreateAsync methods</param>
-        /// <param name="roleValidator"></param>
+        /// <param name="roleValidators">IEnumerable of role validators</param>
+        /// <param name="keyNormalizer">user property normalizers</param>
+        /// <param name="errors">IdentityErrorDescribers</param>
```
divega: OT: Parameter name should probably be errorDescriber unless it is the list of errors. Also, I noticed in the body of the constructor that we initialize a couple of classes if null. Is this the pattern we agreed to get rid of or something else?

src/Microsoft.AspNet.Identity/RoleManager.cs
suhasj: Yeah the scope method calls it in each manager class  while it is used in a lot places in the test code. If I change the Logger property to `BaseLogger` then the call would be `Logger.BaseLogger` or if we are fine that the test code and the method have `Logger.Logger` then i can make the change. Personally I don't like the `Logger.Logger` call

divega: Agree with @HaoK the property here can be just called Logger. the Logger in IdentityLogger is protected so no one can do Logger.Logger, right?

src/Microsoft.AspNet.Identity/IdentityLogger.cs
divega: Since this class doesn't have any state besides the ILogger itself i think we should make its members extension methods and get rid of it.

samples/IdentitySample.Mvc/Startup.cs
suhasj: I looked on the NLog github site and there is no issue reported. I need to file a bug but the guidelines suggest I need to provide a repro. I'll get one up using the public CTP6 bits and open the bug.

As a side, they do use Thread Local storage to store the scopes and as you pointed out, that is blowing up for our async calls

src/Microsoft.AspNet.Identity/RoleManager.cs
divega: Besides that, I am not sure we need the IdentityLogger class (I just wrote a comment for it), so this could end up being a regular ILogger.

src/Microsoft.AspNet.Identity/IdentityLogger.cs
suhasj: They might use it in a scenario where they want to change the implementation of how the IdentityResult, SignInResult, bool etc are logged. I am assuming that would be a very small subset of users. I can move it to a different namespace that's all right ? I don't think we should make the class sealed, remove virtual off methods or make it internal.

src/Microsoft.AspNet.Identity/IdentityLogger.cs
suhasj: sure

src/Microsoft.AspNet.Identity/IdentityLogger.cs
suhasj: I  was a slightly ambiguous on this since if ever they write a method that returns anything else than IdentityResult, SignInResult or bool, they could use it directly. Sure making protected lets them subclass and use it which is fine too

src/Microsoft.AspNet.Identity/RoleManager.cs
```diff
@@ -162,20 +162,23 @@ public void Dispose()
...
             if (!result.Succeeded)
             {
                 return result;
             }
             await UpdateNormalizedRoleNameAsync(role);
-            return await LogResultAsync(await Store.CreateAsync(role, CancellationToken), role);
+            result = await Store.CreateAsync(role, CancellationToken);
+            using (IdentityLogger.BeginScope(await GetScopeMessageforRoleAsync(role)))
+            {
+                return IdentityLogger.Log(result);
```
divega: I wonder if sill makes sense to create a logging scope when we know we will only log one message inside of it. I haven't build a good mental model around it but I would like to know what @lodejard thinks.

divega: BTW, I know we are doing it this way consistently because we only identify the user in the scope message. But still wondering if single entry scopes are supposed to be.

src/Microsoft.AspNet.Identity/IdentityResult.cs
suhasj: Post the logging discussion, I reverted all success messages to Information. To start with we are logging all updates to user/role object only and each one of them is hitting the store for changes. Considering Information is more of a BI scenario, successes to user transactions in Identity I thought was a verbose scenario. So Identity.Success is verbose after updates too. In a scenario we want that to be information then we revert to our old behavior of Information/Warning

src/Microsoft.AspNet.Identity/RoleManager.cs
HaoK: If we want to get tricky, we could capture the messages and only create the scope if we have more than one, since we have a using block, we have that control (to defer the logging calls until the end of the using and then decide to really create the scope or not...  I'm not sure its worth it though)

samples/IdentitySample.Mvc/Startup.cs
divega: Let's file a bug on our side for now (i.e. in Logging), so that we don't forget this is broken.

src/Microsoft.AspNet.Identity/IdentityLogger.cs
divega: I prefer that we keep things simple by making it a static class with extension utility methods (and we move it to a nested namespace). It doesn't seem that compelling to replace it and BTW, this PR does not make it because we are creating new instances with `new` as opposed to making it a service.

src/Microsoft.AspNet.Identity/IdentityResult.cs
divega: Let's check with @lodejard if he thinks verbose for identity transactions is in alignment with the spirit of the guidelines.

src/Microsoft.AspNet.Identity/RoleManager.cs
divega: Probably not :)

src/Microsoft.AspNet.Identity/RoleManager.cs
divega: I would err on keeping things simple, especially if we can just use `ILogger.BeginScope()`.

suhasj: I had that initially but was reading the BeginScope state from a message since I needed the calling method name. having a method lets the users override the behavior as well

src/Microsoft.AspNet.Identity/IdentityLogger.cs
suhasj: But having extension methods won't let users override the behavior right ? That was the feedback when we started initially hence we move to this model

src/Microsoft.AspNet.Identity/RoleManager.cs
suhasj: @divega i can see the point. This behavior is only for CreateAsync on users and roles since we wouldn't have access to the id values unless the object is created, especially for store generated ids. We still need to know that the user/role was created though, so we need to log it. Now we can special case this and log only, but need to call the Log method directly that takes result, Func<LogLevel>, Func<string> since we need to set the userId in the message instead of scope. I am fine either way. This approach only makes it looks clean

src/Microsoft.AspNet.Identity/RoleManager.cs
suhasj: @divega I am not sure I fully agree removing the class altogether. Let discuss in person

src/Microsoft.AspNet.Identity/UserManager.cs
```diff
@@ -1838,42 +1994,8 @@ public virtual Task<IList<TUser>> GetUsersInRoleAsync(string roleName)
...
-                Logger.LogInformation(string.Format("{0} : {1}", baseMessage, result.ToString()));
-            }
-            else
-            {
-                Logger.LogWarning(string.Format("{0} : {1}", baseMessage, result.ToString()));
-            }
-
-            return result;
-        }
+        protected virtual async Task<string> GetScopeMessageforUserAsync(TUser user, [CallerMemberName] string methodName = null)
```
divega: Correct PascalCase name would be `GetScopeMessageForUserAsync` (with an uppercase 'F'). My preference would be `GetLoggingScopeMessageAsync`.

test/Microsoft.AspNet.Identity.Test/IdentityResultTest.cs
```diff
@@ -24,29 +24,5 @@ public void NullFailedUsesEmptyErrors()
             Assert.False(result.Succeeded);
             Assert.Equal(0, result.Errors.Count());
         }
-
```
divega: Are we loosing all logging test coverage?

src/Microsoft.AspNet.Identity/UserManager.cs
suhasj: Based on @HaoK  comment, we can get rid of this method and have a single virtual method in the manager classes, `BeginLogScopeAsync` which returns a `IDisposable` and creating a scopes would be like 
```
using (BeginLogScopeAsync(user)){
...
}
```
users can override that if they want to change the way scope messages are created. Does that look better ?

test/Microsoft.AspNet.Identity.Test/IdentityResultTest.cs
suhasj: ah no we aren't. I have fixed it locally and will be in the final PR. We now check the log messages as well as scopes.

src/Microsoft.AspNet.Identity/RoleManager.cs
suhasj: @HaoK  can provide more insights here. The last I remember the optional ones were initialized to null and instantiated in the constructor to have  a cleaner constructor and a default behavior. Looks like they have been made mandatory again. I thought this was need to have a default behavior if the users did not want to use DI

HaoK: So unit testing is a valid scenario, MVC and Http both invested effort to make their classes new()able for ease of use, so I actually don't think this has been completed settled yet. I wouldn't rathole on this now.  Somewhere on my todo list is to revisit unit testing/etc

suhasj:
@divega @HaoK 
I made all the changes and unit + functional tests run fine. I was running it in the Identity Sample app the output is as below
```
GenerateChangePhoneNumberTokenAsync for user: da49a8f4-0913-457c-a02a-2334bffe5327 Microsoft.AspNet.Identity.UserManager<IdentitySample.Models.ApplicationUser> GenerateChangePhoneNumberTokenAsync : Succeeded
```
As you can the method name is repeated in the scope as well as the log. Considering we are using scopes everywhere it makes sense to now omit the name of the method name from the log but if the logger does not support scopes then the standalone message would be
```
Microsoft.AspNet.Identity.UserManager<IdentitySample.Models.ApplicationUser>  Succeeded
```
which does not provide any good information. I am leaning towards repeating the method since moving forward we might change what goes in the scope information but I wanted to check with you guys if you have some thoughts on this ?

HaoK:
Couldn't the scope msg just contain the user info and for the methods names to show up inside?

suhasj:
You mean something like this
```
User: da49a8f4-0913-457c-a02a-2334bffe5327 Microsoft.AspNet.Identity.UserManager<IdentitySample.Models.ApplicationUser> GenerateChangePhoneNumberTokenAsync : Succeeded
```

suhasj:
With this approach you the scope messages wouldn't be unique and that might have some issues with the loggers parsing them

divega:
I am OK with the repetition for this milestone. Let's file a bug to remember we might need to revisit based on the discussion on scopes.

suhasj:
ok I'll send out the final PR then

suhasj:
Removing the Feedback tag. I have gone over the code for nit issues but I'll make a pass again after taking a break

suhasj:
Some obvious feedback the `BeginLogScopeAsync` method is in the Manager class can be replaces with a `=>` instead of two lines. I thought it was readable so refrained from that. Let me know if i need to change it

src/Microsoft.AspNet.Identity/IdentityLogger.cs
```diff
@@ -0,0 +1,36 @@
+﻿// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Runtime.CompilerServices;
+using Microsoft.AspNet.Identity;
+
+namespace Microsoft.Framework.Logging
```
divega: These extension methods aren't for app developers. They are primarily to be used internally by Identity. So I don't think they should be in this namespace.

HaoK: Why not just make them internal extension methods?

src/Microsoft.AspNet.Identity/RoleManager.cs
divega: Bug so that we don't forget? Or is your TODO list in github? :smile:

src/Microsoft.AspNet.Identity/IdentityLogger.cs
suhasj: When I was testing by extending the manager classes on a sample app, I used those extension methods when I overrode some methods. Since the app developer uses them in custom classes, I chose not to move it to internal namespace

suhasj: For example AppSignInManager returning custom AppSignInResult with an additional status. The user needs to log that

src/Microsoft.AspNet.Identity/RoleManager.cs
```diff
@@ -284,9 +293,13 @@ public virtual string NormalizeKey(string key)
         public virtual async Task<IdentityResult> SetRoleNameAsync(TRole role, string name)
         {
             ThrowIfDisposed();
-            await Store.SetRoleNameAsync(role, name, CancellationToken);
-            await UpdateNormalizedRoleNameAsync(role);
-            return await LogResultAsync(IdentityResult.Success, role);
+
+            using (await BeginLogScopeAsync(role))
```
divega: BeginLogScope to me reads a but weird. Logging instead of Log would be my preference, but English isn't my mother's tongue.

suhasj: Mine neither so I preferred what Hao suggested since he is better at this. Some options
`BeginLoggerScope`
`BeginLoggingScope`
'GetLoggingScope`

HaoK: StartLogScope or BeginLogScope
EndLogScope

src/Microsoft.AspNet.Identity/IdentityLogger.cs
divega: Jus put them in the Microsoft.AspNet.Identity.Internal but make them public. That is something we decided to adopt across all components of the stack.

divega:
I prefer BeginLoggingScope. Consider the feedback an unless @haok has other concerns, :shipit:

src/Microsoft.AspNet.Identity/RoleManager.cs
divega: I prefer BeginLoggingScope.

src/Microsoft.AspNet.Identity/IdentityLogger.cs
HaoK: That is kind of gross, if we want them public, I would suggest we put it under Microsoft.AspNet.Identity.Logging which makes sense

src/Microsoft.AspNet.Identity/RoleManager.cs
divega: ... Async :smile:

src/Microsoft.AspNet.Identity/RoleManager.cs
HaoK: I'm sure we had a bug somewhere, no idea anymore, probably best to file a new one

src/Microsoft.AspNet.Identity/RoleManager.cs
HaoK: I'm ambivalent, BeginLoggingScope is fine

src/Microsoft.AspNet.Identity/IdentityLogger.cs
suhasj: The guidelines are i believe that the namespace for the extension methods be the same the class they are on ? In this case we don't want to the `.Log(IdentityResult)` to show up on the ILogger in places where Identity is not being used. I have conflicting views here but I still prefer keeping it in `Microsoft.Framework.Logging`

HaoK: There is no real conflict we don't want these to show up on all loggers unless they explicitly import an identity specific namespace. I just recommend the namespace is identity.logging instead of a random looking internal/infrastructure.

divega: I am fine with that too. But the Internal namespace isn't really gross :smile: It is just a departure from how we used to do things cc @davidfowl.

divega: >The guidelines are i believe that the namespace for the extension methods be the same the class they are on ? In this case...

That guideline was really meant for extension methods we want everyone to get. We have found since a couple of reasons to deviate.

HaoK: My pushback is not against the pattern in general with Infrastructure/Internal, my pushback is that this isn't internal code we happen to make public.  Its a valid extensibility scenario for custom managers to plug in their own logging and use these extensions, so from that perspective it deserves a proper real namespace like Microsoft.AspNet.Identity.Logging, not a funky (don't use this since it has infrastructure/internal) namespace...

suhasj: Ok 2:1, I move to `Microsoft.AspNet.Identity.Logging` namespace

src/Microsoft.AspNet.Identity/RoleManager.cs
suhasj: And the winner is `BeginLoggingScopeAsync`

suhasj:
Updated and addressed all feedback. Will merge

PR #422, opened 2015-03-27
Enable tests

suhasj:
/cc @HaoK @divega 

Enabling disabled tests after fix in EF. I have verified that this passes in release and dev branches


suhasj:
not for beta4

HaoK:
Did you do this for dev?

suhasj:
yeah just the dev branch

PR #427, opened 2015-04-01
Add SQL Identity support

suhasj:
Issue:
EF chose to use 'sequence' to generate PK values by default which is not supported in SQL Azure. This was causing templates to break functionally when published to Azure. It is being tracked for beta5 by the EF team https://github.com/aspnet/EntityFramework/issues/1960 

Fix:
As a workaround for beta4, we are adding the workaround in IdentityDbContext which should fix the issue for templates and music store. Apps that are started from empty template need to adds this in their DBContext class manually.

@divega @HaoK

### DIFF
--- a/src/Microsoft.AspNet.Identity.EntityFramework/IdentityDbContext.cs
+++ b/src/Microsoft.AspNet.Identity.EntityFramework/IdentityDbContext.cs
@@ -26,6 +26,10 @@ namespace Microsoft.AspNet.Identity.EntityFramework
 
         protected override void OnModelCreating(ModelBuilder builder)
         {
+            // Temporary change. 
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNet.Identity.EntityFramework/IdentityDbContext.cs
```diff
@@ -26,6 +26,10 @@ public class IdentityDbContext<TUser, TRole, TKey> : DbContext
 
         protected override void OnModelCreating(ModelBuilder builder)
         {
+            // Temporary change.
```
divega: Make this // TODO: Temporary change.

divega:
:shipit:

Issue #384, opened 2015-03-05
UserManager.SetUserNameAsync should run the validators before setting the username


suhasj:
The method `UserManager.SetUserNameAsync` is supposed to set the username and is more of an direct POCO manipulation operation. Since it is being done through UserManager methods, I feel that this should run the validators before setting the username. This would differentiate it with doing the same via DbContext.

suhasj:
This is done when UpdateUserAsync is called.

PR #433, opened 2015-04-07
[Feedback] Include validation for set username/email methods

suhasj:
@HaoK @divega 

Issue:
The methods SetUserName/SetEmail do not call user validators when setting the username/email to new values. This seems to be like an issue and hence including that in this fix

Fix:
- In Set username/email methods call `ValidateUserInternal` once the username/email is set to the new value. If the validation succeeds, then continue else revert back to the old values
- Added two tests that test this operation. We try to set the username/email to empty and duplicate values and check if the validation succeeds

Question
- In the SetUsername we don't call UpdateSecurityStamp method. I think we should be doing it. Any reasons for not doing it right now ?

https://github.com/aspnet/Identity/issues/384

### DIFF
--- a/src/Microsoft.AspNet.Identity/UserManager.cs
+++ b/src/Microsoft.AspNet.Identity/UserManager.cs
@@ -494,8 +494,16 @@ namespace Microsoft.AspNet.Identity
 
             using (await BeginLoggingScopeAsync(user))
             {
+                var oldUserName = await GetUserNameAsync(user);
                 await UpdateUserName(user, userName);
-                return Logger.Log(await UpdateUserAsync(user));
+                var result = await ValidateUserInternal(user);
+                if (result.Succeeded)
+                {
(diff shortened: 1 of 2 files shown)


HaoK:
This looks kind of gross with the setting and restoring if validation fails.

HaoK:
Actually I don't see how this is needed at all, doesn't UpdateUser already run the validators?

HaoK:
Go ahead and file a new issue for SetUserName should be calling UpdateSecurityStamp, and fix that in a different PR

suhasj:
Ah true I had missed that out. In that the scenario, the username is set on the object but it returns back during update on validation failure. The values in the object are not reverted. I guess it makes sense from an app perspective that the object should no longer used and further operations should be skipped. I'll close out this PR and send a new one for the ones I opened

PR #434, opened 2015-04-08
Changes to authentication

suhasj:
@HaoK 
The samples were broken because of the recent updates. fixing it.


HaoK:
This fix doesn't look correct, need to keep using IsSignedIn, it was just moved

suhasj:
merged

PR #435, opened 2015-04-08
[Feedback] Sample logging changes

suhasj:
@HaoK @divega 
I took a shot at the logging feedback post our meeting with @loudej . The full list of feedback is added to team one note

Some pivots I have used for the changes
* Replaced automated logging with readable statements
* All statements in English and no use of Resource file.
* Direct inplace logging with no extension methods
* check if log is enabled in places where we get the userid via await call only

Questions
* Do we want the logging statements to be of a standard format. for example `{operation} user: {userId}`. This might make it more searchable via patterns but this open to debate
* This is not extensible right now so I am too sure about this. We can read the statements from a configurable place so that at least the statements can be altered if needed by the developer. Let me know what you think.
* Should I move the calls
```
if (Logger.IsEnabled(LogLevel.Warning))
                {
                    Logger.LogWarning("Message with {userId}", await UserManager.GetUserIdAsync(user));
                }
```
to a private method 

```
await Logger.LogWarningAsync(""Message with {userId}", user);
```
This will check if logger is enabled and then log message extracting the userId. This does not look good to me since the method should be Async and then we need to await on it. I am not for this hence not included in the PR.
* I am logging signin errors in Warning, successful signin in Verbose and successful login + two factor needed in Information. Let me know if we should be reverting to using just to levels. In that case should it be Warning/Information or Warning/Verbose ?

### DIFF
--- a/src/Microsoft.AspNet.Identity/SignInManager.cs
+++ b/src/Microsoft.AspNet.Identity/SignInManager.cs
@@ -4,14 +4,11 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Runtime.CompilerServices;
 using System.Security.Claims;
-using System.Security.Principal;
 using System.Threading.Tasks;
 using Microsoft.AspNet.Hosting;
 using Microsoft.AspNet.Http;
 using Microsoft.AspNet.Http.Authentication;
-using Microsoft.AspNet.Identity.Logging;
 using Microsoft.Framework.Logging;
 using Microsoft.Framework.OptionsModel;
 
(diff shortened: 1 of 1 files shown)


suhasj:
I have done this for SignInManager. Once we have a consensus on the flow, I'll update the rest of the code and tests

HaoK:
Personally I can't stand seeing if (logLevel enabled) checks everywhere, my vote would be to either just  do it without the guard, or put it in a private method (since I am pretty sure @divega will want to avoid the await overhead)

divega:
@HaoK hate it too but I saw the alternative in previous PRs, e.g. an async extension method and it looked terrible. I wish getting the user id wasn't async.

IECOMPAT1:
cool

PR #438, opened 2015-04-09
Correct user manager setter

suhasj:
@HaoK @divega 
The old PR was not restored, hence this new one. 

Changes
- removed the private method for UpdateUserName since the normalizedUsername is called on UpdateUserAsync method
- Added unit test to check validation happen on set username/ password

### DIFF
--- a/src/Microsoft.AspNet.Identity/UserManager.cs
+++ b/src/Microsoft.AspNet.Identity/UserManager.cs
@@ -1994,7 +1994,6 @@ namespace Microsoft.AspNet.Identity
             return Logger.BeginScope(state);
         }
 
-
         private void ThrowIfDisposed()
         {
             if (_disposed)
--- a/src/Microsoft.AspNet.Identity/UserManager.cs
+++ b/src/Microsoft.AspNet.Identity/UserManager.cs
@@ -494,8 +494,16 @@ namespace Microsoft.AspNet.Identity
 
             using (await BeginLoggingScopeAsync(user))
             {
+                var oldUserName = await GetUserNameAsync(user);
                 await UpdateUserName(user, userName);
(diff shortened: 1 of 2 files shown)


test/Shared/UserManagerTestBase.cs
```diff
@@ -128,15 +128,56 @@ protected RoleManager<TRole> CreateRoleManager(object context = null, IServiceCo
...
+
+        [Fact]
+        public async Task CheckSetEmailValidatesUser()
+        {
+            var manager = CreateManager();
+            manager.Options.User.RequireUniqueEmail = true;
+            manager.UserValidators.Add(new UserValidator<TUser>());
+            var random = new Random();
+            var email = "foo" + random.Next() + "@example.com";
+            var newEmail = "foo" + random.Next() + "@example.com";
```
HaoK: Why risk collisions?  Change the prefix

src/Microsoft.AspNet.Identity/UserManager.cs
```diff
@@ -1998,7 +1993,7 @@ public virtual Task<IList<TUser>> GetUsersInRoleAsync(string roleName)
             var state = Resources.FormatLoggingResultMessageForUser(methodName, await GetUserIdAsync(user));
             return Logger.BeginScope(state);
         }
-            
+
```
HaoK: Empty line added

test/Shared/UserManagerTestBase.cs
suhasj: sure

HaoK:
:shipit:

suhasj:
@divega Reminder

divega:
:shipit:

Issue #420, opened 2015-03-25
UserStore/RoleStore ConvertIdFromString can't use GUID


cherrydev:
``` (TKey)Convert.ChangeType(id, typeof(TKey)); ``` doesn't work for System.Guid

However, ``` (TKey) TypeDescriptor.GetConverter(typeof (TKey)).ConvertFromInvariantString(id); ``` does work!

**Comment by Bartmax** [comment]
is there any known workaround for this for the time being?

**Comment by cherrydev** [comment]
My workaround was implementing my own UserStore and RoleStore.  I was porting it to EF6 anyhow.

**Comment by suhasj** [comment]
@Bartmax you can override this method in your own subclass of UserStore/RoleStore and then hook that via DI in startup class. Let me know if you need more help with that

**Comment by Bartmax** [comment]
Thanks, this worked:
```
services.AddIdentity<ApplicationUser, ApplicationRole>()
    .AddUserStore<ApplicationUserStore>()
    .AddRoleStore<ApplicationRoleStore>();


public class ApplicationUserStore : UserStore<ApplicationUser, ApplicationRole, ApplicationDbContext, Guid>
    {
        public ApplicationUserStore(ApplicationDbContext context, IdentityErrorDescriber describer)
            : base(context, describer)
        {

        }
        public override Guid ConvertIdFromString(string id)
        {
            return new Guid(id);
        }
        public override string ConvertIdToString(Guid id)
        {
            return id.ToString();
        }
    }

    public class ApplicationRoleStore : RoleStore<ApplicationRole, ApplicationDbContext, Guid>
    {
        public ApplicationRoleStore(ApplicationDbContext context, IdentityErrorDescriber describer)
            :base(context, describer)
        {

        }
        public override string ConvertIdToString(Guid id)
        {
            return id.ToString();
        }
        public override Guid ConvertIdFromString(string id)
        {
            return new Guid(id);
        }
    }
```

HaoK:
Was fixed in afff831d624b5a5e2677b0b42088626e8069336f

PR #444, opened 2015-04-15
Updating stores to use TypeDescriptor for type conversion

suhasj:
@HaoK @divega 

Fix for https://github.com/aspnet/Identity/issues/420 

The `TypeDescriptor` class is available in both desktop and core clr so taking this change. Verified it functionally. Removed the existing code in functional tests which was causing it pass incorrectly so now we have coverage there as well

### DIFF
--- a/src/Microsoft.AspNet.Identity.EntityFramework/project.json
+++ b/src/Microsoft.AspNet.Identity.EntityFramework/project.json
@@ -12,6 +12,7 @@
       "dependencies": {
         "System.Collections": "4.0.10-beta-*",
         "System.ComponentModel": "4.0.0-beta-*",
+        "System.ComponentModel.TypeConverter": "4.0.0-beta-*",
         "System.Diagnostics.Debug": "4.0.10-beta-*",
         "System.Diagnostics.Tools": "4.0.0-beta-*",
         "System.Globalization": "4.0.10-beta-*",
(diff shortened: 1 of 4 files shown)


HaoK:
:shipit:  Maybe make this change for Options as well?  https://github.com/aspnet/Options/blob/dev/src/Microsoft.Framework.OptionsModel/OptionsServices.cs#L64

suhasj:
sure I can do that.

suhasj:
@HaoK `TypeDescriptor` is not present in the portable subset. Should I put an `ifdef` for that and fall bacl to `Convert.ChangeType` ?

HaoK:
Bleh leave options alone then

suhasj:
ok

suhasj:
@divega Reminder

src/Microsoft.AspNet.Identity.EntityFramework/project.json
```diff
@@ -12,6 +12,7 @@
       "dependencies": {
         "System.Collections": "4.0.10-beta-*",
         "System.ComponentModel": "4.0.0-beta-*",
```
divega: Just in case, do we still the one above?

divega:
:shipit:

suhasj:
Had a chat with @divega and we'll be adding this to Options as well. Will send it in a separate PR.

Issue #431, opened 2015-04-05
Remove Hosting dependency 


davidfowl:
https://github.com/aspnet/Identity/blob/dev/src/Microsoft.AspNet.Identity/project.json#L8

HaoK:
3e9433639021255c1373c9a4bd36c4a511982d18

PR #445, opened 2015-04-16
Remove hosting and logging dependencies

HaoK:
Means test code has to provide logging/hosting services, and added null guards for code to compensate for lack of defaults

cc @divega @suhasj


HaoK:
Fixes https://github.com/aspnet/Identity/issues/431

suhasj:
Can we have a test when the when no logger is passed ? The TestIdentityFactory is registering AddLogging so I feel the current test will always have some value

HaoK:
We could go the other direction on this and require it via [NotNull] instead.  No test necessary that way

HaoK:
Added a new test with null logger to ensure things don't blow up, also cleaned up SignInmanager tests reducing a lot of the duplication with a SetupHelper

suhasj:
This looks good. Better than having the `NotNull` attribute since that would enforce using a logger always which I am not sure would be the right approach

HaoK:
3e9433639021255c1373c9a4bd36c4a511982d18

PR #451, opened 2015-04-21
Use configure from cookies repo

suhasj:
@HaoK @Tratcher 

This code basically encapsulates the `AddWebEncoders` call and we don't have to do it explicitly in Identity. I verified all tests pass with this change.


Tratcher:
:+1:

suhasj:
is that the new shipit ?

suhasj:
@HaoK  pushing this to unblock CI. I can revert the changes if you think this should not be in

Issue #185, opened 2014-08-14
How to know if RememberMe was checked for the cookie


rustd:
There are cases where we loose the rememberMe information about the user when we regenerate the cookie.
eg. If you change the password, then we regenerate the cookie and we have no way of knowing whether the user had selected the rememberMe option

HaoK:
c251bf0162091730b6cc74d97731db74d98e499d

PR #452, opened 2015-04-21
Add support for ReSignIn

HaoK:
- Add ResignIn api to regenerate the application cookie preserving rememberMe and AuthenticationMethod
- Update manage controller in samples to use new API

Fixes: https://github.com/aspnet/Identity/issues/185

@divega @Tratcher @suhasj @rustd @blowdart

### DIFF
--- a/src/Microsoft.AspNet.Identity/SignInManager.cs
+++ b/src/Microsoft.AspNet.Identity/SignInManager.cs
@@ -77,7 +77,7 @@ namespace Microsoft.AspNet.Identity
         /// </summary>
         /// <param name="user"></param>
         /// <returns></returns>
-        public virtual async Task ReSignInAsync(TUser user)
+        public virtual async Task RefreshSignInAsync(TUser user)
         {
             var authResult = await Context.AuthenticateAsync(IdentityOptions.ApplicationCookieAuthenticationScheme);
             var properties = authResult?.Properties ?? new AuthenticationProperties();
--- a/src/Microsoft.AspNet.Identity/SignInManager.cs
+++ b/src/Microsoft.AspNet.Identity/SignInManager.cs
@@ -32,7 +32,6 @@ namespace Microsoft.AspNet.Identity
             {
                 throw new ArgumentNullException(nameof(userManager));
             }
-
             if (contextAccessor == null || contextAccessor.HttpContext == null)
             {
                 throw new ArgumentNullException(nameof(contextAccessor));
@@ -46,7 +45,6 @@ namespace Microsoft.AspNet.Identity
             Context = contextAccessor.HttpContext;
             ClaimsFactory = claimsFactory;
             Options = optionsAccessor?.Options ?? new IdentityOptions();
-
             Logger = logger?.CreateLogger<SignInManager<TUser>>();
         }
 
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNet.Identity/SignInManager.cs
```diff
@@ -75,6 +75,21 @@ public class SignInManager<TUser> where TUser : class
...
+        /// Called to regenerate the ApplicationCookie for the user, preserving the existing
+        /// AuthenticationProperties like rememberMe
+        /// </summary>
+        /// <param name="user"></param>
+        /// <returns></returns>
+        public virtual async Task ResignInAsync(TUser user)
+        {
+            var authResult = await Context.AuthenticateAsync(IdentityOptions.ApplicationCookieAuthenticationScheme);
+            var properties = authResult?.Properties;
+            var isPersistent = properties != null && properties.IsPersistent;
```
suhasj: So we are preserving only the `IsPersistent` property ? The other property I feel might be slightly important is the Dictionary. Do you think we need that too ?

HaoK: @Tratcher what do you think we should preserve when it comes to the AuthenticationProperties, we definitely want to regenerate the claims, but we want to keep as many of the old settings as possible that make sense...

Tratcher: It's a generic property bag, you'd want to preserve anything you can.

HaoK: Is it safe to copy everything over then?  Except for maybe nulling out ExpiresUtc and IssuedUtc unless SignIn will take care of restamping those for us...

HaoK: @divega We can consider changing SignIn to mimic the underlying SignIn API more closely, remove IsPersistent and take in a raw AuthenticationProperties which would expose everything

HaoK:
Updated, kept old SignIn overload which just takes IsPersistence, but added an overload which takes AuthenticationProperties, and switched ReSignInAsync to use that.

divega:
I am ok with this, but didn't you say that we were the main users of this API and it didn't make sense to keep the overload that takes a bool? Anything changed your point of view?

HaoK:
Well, I tried it without the overload, and it was annoying and i immediately wrote a helper overload, so might as well keep it and leave it public (no breaking change at all this way too).  So that is what convinced me to keep it :)

divega:
Fair enough :smile: :shipit:

rustd:
The name of the method is not great. How about RefreshIdentity? RegenerateIdentity?

divega:
Fair point too. Seems to me that the word "renew" may be closer in meaning to what the API will do.

divega:
E.g. 'RenewSignIn' or something like that.

HaoK:
RenewSignIn or RefreshSignIn works for me

rustd:
Either one is good. Let's pick RefreshSignIn?

HaoK:
Ok updated to RefreshSignIn

divega:
:ship it:

HaoK:
c251bf0162091730b6cc74d97731db74d98e499d

Issue #411, opened 2015-03-18
Reenable Startup identity tests


HaoK:
Disabling as part of the hosting work since there's no easy way to do this today

Probably need some helper that makes it easy to run aribtrary code as part of a request (simialr to what auth functionals do)

HaoK:
Merged 288cb6c58f26aa10a76c583e18f3063b14074b4f

PR #454, opened 2015-04-27
Reenable startup tests

HaoK:
Moving to functionals rather than trying to fake them...

Fixes https://github.com/aspnet/Identity/issues/411

cc @divega @suhasj


HaoK:
@divega Updated PR removing IdentityErrorDescriber.Default and cleaned up tests + switched back to `ILogger<Manager>` instead of just taking ILoggerFactory

test/Shared/UserManagerTestBase.cs
```diff
@@ -39,7 +35,8 @@ protected virtual void SetupIdentityServices(IServiceCollection services, object
             services.AddIdentity<TUser, TRole>().AddDefaultTokenProviders();
             AddUserStore(services, context);
             AddRoleStore(services, context);
-            services.AddInstance<ILoggerFactory>(loggerFactory);
+            services.AddInstance<ILoggerFactory>(LoggerFactory);
```
divega: Is this still necessary or next line is enough?

HaoK: This is needed because its sticking a TestLoggingFactory in there will cause the normal ILoggerFactory to not get added(TryAdd will no-op), but we need AddLogging to bring in the `ILogger<>` service

divega:
:shipit: nice cleanup.

test/Shared/UserManagerTestBase.cs
divega: Ok

HaoK:
Merged 288cb6c58f26aa10a76c583e18f3063b14074b4f

Issue #456, opened 2015-04-30
Add timeout values to all regex expressions


blowdart:
To avoid regex attacks add a time-out value to the username validation regex.

HaoK:
Fixed eaa7ca9e81de89ff27038aeae4029f565cc01b24

PR #461, opened 2015-05-08
Add regex timeout, and set default.

blowdart:


HaoK:
:shipit: does this bring back fond memories of when you were a dev? :)

HaoK:
Fixes https://github.com/aspnet/Identity/issues/456

PR #462, opened 2015-05-08
Remove internal SecurityToken class

HaoK:
Unneeded, it was just wrapping a byte[], directly use the byte[] array

cc @blowdart @divega


HaoK:
Note: we are no longer cloning the bytes[] from Encoding.Unicode.GetBytes, but does that really matter here?  cc @GrabYourPitchforks

HaoK:
6d206250b3fe70cf34820130d1145be0cba1ad3a

PR #468, opened 2015-05-17
Fix build break due to Configuration rename

Eilon:
@kirthik


Eilon:
Note: This change currently depends on the volatile feed because Coherence hasn't run. That's why AppVeyor is failing.

PR #475, opened 2015-06-04
Prevent AddToRolesAsync from throwing NullRefEx

ninlar:
Custom implemented role storage providers may return null instead of an empty IList<string> for GetRolesAsync when the user has no roles assigned. AddToRolesAsync should handle this case rather than throwing a NullReferenceException when invoking userRoles.Contains(role).

### DIFF
--- a/src/Microsoft.AspNet.Identity/UserManager.cs
+++ b/src/Microsoft.AspNet.Identity/UserManager.cs
@@ -1078,7 +1078,9 @@ namespace Microsoft.AspNet.Identity
             using (await BeginLoggingScopeAsync(user))
             {
                 var userRoles = await userRoleStore.GetRolesAsync(user, CancellationToken);
-                if (userRoles.Contains(role))
+                bool hasRolesAssigned = (!ReferenceEquals(userRoles, null) && userRoles.Count > 0);
+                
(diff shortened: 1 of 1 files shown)


HaoK:
Can you make a few small tweaks and resubmit this?
- Don't use ReferenceEquality and just check == null
- Can you also look for all other places that are doing this pattern (should be at least the claims/logins apis) and fix them as well?

ninlar:
Sure, I'll make the adjustments tonight.

divega:
@ninlar @HaoK any progress on this PR?

divega:
@HaoK I looked at the current code around this PR today (needs to be rebased!) and it doesn't seem to apply to Claims (we don't check for existent claims) and Logins (we use a `Find` find method and already check for null results). I would like to take this PR (or the equivalent fix) ASPA since it has been sitting there for too long.

HaoK:
Ok I will add a check in a separate PR, this code isn't PR in its current state

ninlar:
Sorry guys, I've been swamped over the last couple of months. Glad I could at least point out the issue.

divega:
Thanks @ninlar, for bringing the issue to our attention. And please don't worry, I am sure there are many other areas in which you can contribute!

Issue #183, opened 2014-08-12
Consider overriding ToString for POCOs


HaoK:
Helpful for databinding scenarios potentially?

http://stackoverflow.com/questions/25267073/asp-net-identity-2-0-get-rolenames-in-a-dropdown-list

rustd:
@HaoK beta5 or 6?

HaoK:
4e5bba35da259fa79a90426ed02a72f0c03ec861

PR #479, opened 2015-06-08
Friendlier ToString

HaoK:
Fixes https://github.com/aspnet/Identity/issues/183

@divega @blowdart @rustd Thoughts on the text?  I wasn't sure if the "Role/User" prefix was necessary, shrug...


blowdart:
I wouldn't add them. It's not like if you call toString on a date you get "Date: 08 June 2015"

HaoK:
Ok so just print out the name by itself?

blowdart:
Yup

HaoK:
4e5bba35da259fa79a90426ed02a72f0c03ec861

PR #481, opened 2015-06-09
Move Store interfaces to Store namespace

HaoK:
We could move this into a Store assembly...  this poses the question of whether we should name the EF package... Microsoft.AspNet.Identity.Store.EntityFramework

Microsoft.AspNet.Identity
Microsoft.AspNet.Identity.Store
Microsoft.Aspnet.Identity.Store.EntityFramework
...

@divega


divega:
We discussed and decided that we are not doing this given that it feels incomplete without moving the types to a separate package and that doesn't seem to add much value (similar to the Abstractions package).

PR #482, opened 2015-06-09
Identity Logging iteration - nuke scopes/helpers

HaoK:
Cleanup logging for SignInManager (Will do User/RoleManager) in a follow up PR matching the pattern established here...

cc @divega

### DIFF
--- a/src/Microsoft.AspNet.Identity/SignInManager.cs
+++ b/src/Microsoft.AspNet.Identity/SignInManager.cs
@@ -205,14 +205,15 @@ namespace Microsoft.AspNet.Identity
             }
             if (await IsLockedOut(user))
             {
-                Logger.LogWarning("User {userId} is currently locked out.", await UserManager.GetUserIdAsync(user));
-                return SignInResult.LockedOut;
+                return await LockedOut(user);
             }
             if (await UserManager.CheckPasswordAsync(user, password))
             {
                 await ResetLockout(user);
                 return await SignInOrTwoFactorAsync(user, isPersistent);
             }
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNet.Identity/SignInManager.cs
```diff
@@ -197,34 +198,32 @@ public virtual void SignOut()
...
-                    if (await UserManager.IsLockedOutAsync(user))
-                    {
-
-                        return Logger.Log(SignInResult.LockedOut);
-                    }
+                    return await LockedOut(user);
                 }
-                return Logger.Log(SignInResult.Failed);
             }
+            // REVIEW: do we want to log normal password failures which would show up here?
```
divega: Maybe @blowdart will have an opinion. I think it is fine with logging just lockouts.

HaoK:
@blowdart is it ok if we only log Warnings for when users are locked out and don't log anything for normal password mismatch failures?

src/Microsoft.AspNet.Identity/SignInManager.cs
blowdart: Logging normal failures would enable better auditing of potential attacks.

divega:
Nice cleanup! I like that we are not logging everything but only when things go wrong, so we don't need to care that much for the overhead of logging.

:shipit:

blowdart:
Log "normal" mismatches please.

HaoK:
e2d8bfe5cfa8ed014fbce21b924d362ffc87512a

HaoK:
Added logging for password failures, updating rest of managers in the next PR

Issue #155, opened 2014-07-23
Use [NotNull] everywhere


HaoK:
Use the EF public api tests to find missing Async suffixes, missing cancellationTokens and [NotNull]

HaoK:
@davidfowl Is NotNull actually working?  i.e. can I rely on it to actually throw Argument null exception for me?  If not, when is it coming?

HaoK:
Community PR addressing this here https://github.com/aspnet/Identity/pull/485

PR #485, opened 2015-06-11
Using [NotNull] and 'nameof' operator

hishamco:


hishamco:
\cc @HaoK

HaoK:
Nice, fixes https://github.com/aspnet/Identity/issues/155

divega:
We wanted to automatically generate the checks using Roslyn but AFAIK that is still at the prototype phase so I wouldn't get rid of the actual checks for now. 

cc @AndriySvyryd @davidfowl

hishamco:
@divega But there are some cases such as checking with ```String.IsNullOrEmpty()``` or typo inside the exceptions literal may difficult little bit to discover, I already did a check in all the repos last month to make a fix, also I'm doing another check from yesterday, I hope you take this in mind during the merge to avoid such things in the future ..

hishamco:
FYI I did the same fix in almost the repos - please i hope someone merge these changes ..

hishamco:
@HaoK can you merge this

divega:
@Eilon I didn't have a chance to sync up with you about this. My take is that we want to take the addition of the attribute and the nameof() usage for sure, but that we shouldn't remove the actual checks, at least not yet. What do you think?

Eilon:
I think most repos don't bother with still having an exception, with the expectation of NotNull being implemented. I think the few cases where they're left is where perhaps some tests or whatever already assume the check will be there and we don't want to modify the tests.

divega:
In that case, @HaoK please finish reviewing and merge it.

HaoK:
Lets wait til NotNull runtime actually works and then I will merge

dougbu:
Curious: Why is this PR blocked? It wouldn't be blocked in most aspnet repos.

divega:
@HaoK and I had a chat and his judgment is that he doesn't want to remove the tests that check that the API throws when null are passed at this time. He wants to wait for [NotNull] rewriter to be fully implemented. FWIW, we are also keeping the checks in EF, at least for now, but that didn't have an influence.

hishamco:
FYI that tests fails in this PR https://github.com/aspnet/FileSystem/pull/109

HaoK:
Turns out we need to keep the null checking, can you revert the changes that remove the tests and then I can merge the nameof + [NotNull] changes

Eilon:
@HaoK FYI we're most likely going to drop `[NotNull]` and use manual `null` checks the Old School way.

HaoK:
Yeah Identity never removed them thankfully :)  So revert everything except the nameof changes then for this PR

hishamco:
@HaoK should I rebase and remove them changes in tests only

divega:
@Eilon I am not sure we should drop the attributes. Should we discuss this offline?

Eilon:
@divega Oh sorry the attributes we might want to stay in projects where we want to continue using them, but we'll still need to add explicit null checks and throws.

divega:
Thanks. For the time being the main advantage of the attributes is for Resharper users working with the codebase. I will discuss it with @HaoK although he isn't the only stakeholder :smile:

HaoK:
I don't have a strong opinion either way on the attributes. Especially if someone else is adding them :). I just don't want to remove them or add them if they aren't there already is my general rule of thumb now if they don't do anything

divega:
Then let's leave the attributes in place for now. They don't affect anything at runtime but can be still useful when working on the codebase. 

BTW we can get rid of them when we get [non-nullable reference type support](https://github.com/dotnet/roslyn/issues/5032) in the language :smile:

divega:
Doh' @HaoK clarified that we currently don't have the attributes in our codebase, so we agreed we don't need to add them. Sorry I didn't take the time to look.

src/Microsoft.AspNet.Identity/UserManager.cs
```diff
@@ -796,19 +806,19 @@ public virtual Task<TUser> FindByLoginAsync(string loginProvider, string provide
         {
             ThrowIfDisposed();
             var loginStore = GetLoginStore();
-            if (loginProvider == null)
+            if (user == null)
```
HaoK: nit: lets not change the ordering of these checks, its possible the unit tests rely on the current priority

src/Microsoft.AspNet.Identity/UserManager.cs
```diff
@@ -985,13 +998,13 @@ public virtual Task<IdentityResult> RemoveClaimAsync(TUser user, Claim claim)
...
-                throw new ArgumentNullException("user");
+                throw new ArgumentNullException(nameof(user));
             }
             if (claims == null)
             {
-                throw new ArgumentNullException("claims");
+                throw new ArgumentNullException(nameof(claims));
             }
-
+
```
HaoK: Can you autoformat the all the files as well?  It looks like you added a bunch of misc white space in a bunch of places

hishamco:
Don't worry @HaoK I will take care about these, but i'm facing some issues to run ```dnx test``` locally so I will let you know when i finish .. thanks

src/Microsoft.AspNet.Identity/UserManager.cs
hishamco: You are right

src/Microsoft.AspNet.Identity/UserManager.cs
hishamco: @HaoK that's what I have seen after the condition in almost the code

hishamco:
It takes time to figure out which of ```[NotNull]``` that doesn't let the tests run successfully, specifically if you have an issue to run ```dnx test``` locally :smile: 
@HaoK can you review and merge if everything is fine

HaoK:
Remove all of the not null attributes and usings and if it compiles that should be sufficient

On Sep 11, 2015, at 4:11 PM, Hisham Bin Ateya <notifications@github.com<mailto:notifications@github.com>> wrote:


It takes time to figure out which of [NotNull] that doesn't let the tests run successfully, specifically if you have an issue to run dnx test locally [:smile:]
@HaoK<https://na01.safelinks.protection.outlook.com/?url=https%3a%2f%2fgithub.com%2fHaoK&data=01%7c01%7chaok%40microsoft.com%7cc18e488447844eb161ee08d2bafe5cce%7c72f988bf86f141af91ab2d7cd011db47%7c1&sdata=yLaO3YGELQB2bM1DsgNC6yIsZyzb7R4oYzNLCTcI%2bJY%3d> can you review and merge if everything is fine

-

hishamco:
Unfortunately i'm facing issues with dnx commands in my end :(

hishamco:
@HaoK  Shall I revert the ```NotNull`` changes while @pranavkm take the initiative to execute NotNull -> Throw ArgumentNullException script to almost the aspnet repos

pranavkm:
@hishamco might be useful since the checks will get moved to the top of the method with my script (versus after the `ThrowIfDisposed` that it does now). Just seems like needless diffs.

hishamco:
So shall I revert ```NotNull``` changes?

HaoK:
Yeah the only changes that we need in this PR are the nameof updates, NotNull and any other checks should be reverted

hishamco:
Ok @HaoK

hishamco:
@HaoK the PR has been updated

divega:
:shipit:

hishamco:
Just I figure out there's a typo here https://github.com/aspnet/Identity/pull/485/files#diff-5fad47ab1c97fb22b91f704b2e78a262R982 , I updated the PR while @HaoK is already merge. Seem he fix it.
Thanks @HaoK

Issue #414, opened 2015-03-20
Avoid repetition of method names in log messages and scopes


suhasj:
The method names are repeated in scopes as well as log messages. Revisit them to have a better experience

suhasj:
Piggy backing on this bug for logging changes for beta5

- Intelligent logging
- Using BeginScope with string and parameters if necessary
- Log Information for methods that can be called from Controller class

HaoK:
e80853a4cf665d73e7e5cdcb652d3d51cd456068

PR #486, opened 2015-06-11
Cleaning up logging

HaoK:
Finish cleaning up logging for user manager/role manager

Fixes https://github.com/aspnet/Identity/issues/414

cc @divega


src/Microsoft.AspNet.Identity/UserManager.cs
```diff
@@ -2183,7 +2052,12 @@ private static string GetChangeEmailPurpose(string newEmail)
                     errors.AddRange(result.Errors);
                 }
             }
-            return errors.Count > 0 ? IdentityResult.Failed(errors.ToArray()) : IdentityResult.Success;
+            if (errors.Count > 0)
+            {
+                Logger.LogWarning($"User {await GetUserIdAsync(user)} validation failed: {string.Join(";", errors.Select(e => e.Code))}.");
```
divega: Starting to love string interpolation :smile:

divega:
:shipit: 

Really nice cleanup! Hope we can keep the string interpolation usage.

src/Microsoft.AspNet.Identity/UserManager.cs
dougbu: Based on previous changes, we do _not_ want to do any formatting up front. This passes an opaque string into the logging system and some of the logging providers can do useful things with the raw data. So, use the logging-specific format strings and pass the values in.

(@Eilon you wanted to be mentioned)

Eilon: Correct. Logs need to use our own custom string formatting to be useful.

HaoK: Updated to use logging params overloads everywhere

HaoK:
e80853a4cf665d73e7e5cdcb652d3d51cd456068

Issue #478, opened 2015-06-08
options.Lockout.EnabledByDefault should be true by default


rustd:

rustd:
options.Lockout.AllowedForNewUsers
default =true

parameter name for API name on PasswordSignInAsync
lockoutOnFailure

HaoK:
1ed2ebcd962cf86c57da9e54626732c9feb0d21c

PR #487, opened 2015-06-11
Enable lock out by default

HaoK:
Fixes https://github.com/aspnet/Identity/issues/478

Note this is a breaking change for templates as the AccountController was using this parameter by name

cc @divega @rustd

### DIFF
--- a/src/Microsoft.AspNet.Identity/LockoutOptions.cs
+++ b/src/Microsoft.AspNet.Identity/LockoutOptions.cs
@@ -11,15 +11,15 @@ namespace Microsoft.AspNet.Identity
     public class LockoutOptions
     {
         /// <summary>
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNet.Identity/LockoutOptions.cs
```diff
@@ -11,15 +11,15 @@ namespace Microsoft.AspNet.Identity
     public class LockoutOptions
     {
         /// <summary>
-        /// Gets or sets a flag indicating whether users are locked out upon creation.
```
HaoK: @blowdart The doc comments for this were totally wrong :)

divega:
:shipit: for me. We should make the change in beta5 since it affects templates. Can you get approval?

HaoK:
Fixed in release

PR #492, opened 2015-06-12
Rename shouldLockout

HaoK:
New PR from release branch

cc @divega

### DIFF
--- a/src/Microsoft.AspNet.Identity/LockoutOptions.cs
+++ b/src/Microsoft.AspNet.Identity/LockoutOptions.cs
@@ -10,7 +10,7 @@ namespace Microsoft.AspNet.Identity
         /// <summary>
         ///     If true, will enable user lockout when users are created
         /// </summary>
-        public bool AllowedForNewUsers { get; set; } = false;
+        public bool AllowedForNewUsers { get; set; } = true;
 
         /// <summary>
         ///     Number of access attempts allowed for a user before lockout (if enabled)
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNet.Identity/LockoutOptions.cs
```diff
@@ -10,7 +10,7 @@ public class LockoutOptions
         /// <summary>
         ///     If true, will enable user lockout when users are created
         /// </summary>
-        public bool EnabledByDefault { get; set; } = false;
+        public bool AllowedForNewUsers { get; set; } = false;
```
divega: It seems I am a bit lost with one thing here: aren't we supposed to flip the default for this? If yes, where is that flip happening?

HaoK: In a local change that I forgot to push :) Was lost in the merge from the dev branch...

HaoK: Pushed now

divega:
:shipit:

rustd:
@HaoK can you please update the templates repo as well. This is in the release branch.
https://github.com/aspnet/Templates/blob/release/src/Rules/StarterWeb/IndividualAuth/Controllers/AccountController.cs#L66

HaoK:
1ed2ebcd962cf86c57da9e54626732c9feb0d21c

PR #497, opened 2015-06-17
Makes AddEntityFrameworkStores use TryAdd instead of Add 

### LINKED ISSUE #496: Shouldn't AddEntityFrameworkStores use services.TryAdd instead of just services.Add ?
I created my own implementation of the IUserStore and added it to the services, but calling AddEntityFrameworkStores registers a new implementation for the interface (the default one) and makes UserManager to use that one instead of mine.

I believe the problem can be easily solved by changing the implementation of the extension method from:
```csharp
public static IdentityBuilder AddEntityFrameworkStores<TContext>(this IdentityBuilder builder) where TContext : DbContext
{
    builder.Services.Add(IdentityEntityFrameworkServices.GetDefaultServices(builder.UserType, builder.RoleType, typeof(TContext)));
    return builder;
}
```

to:
``` csharp
public static IdentityBuilder AddEntityFrameworkStores<TContext>(this IdentityBuilder builder) where TContext : DbContext
{
    builder.Services.TryAdd(IdentityEntityFrameworkServices.GetDefaultServices(builder.UserType, builder.RoleType, typeof(TContext)));
    return builder;
}
```

So that it only adds the services if I haven´t added my own implementation before.

eduaglz:
Fixes #496

### DIFF
--- a/src/Microsoft.AspNet.Identity.EntityFramework/IdentityEntityFrameworkBuilderExtensions.cs
+++ b/src/Microsoft.AspNet.Identity.EntityFramework/IdentityEntityFrameworkBuilderExtensions.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Framework.DependencyInjection
         public static IdentityBuilder AddEntityFrameworkStores<TContext>(this IdentityBuilder builder)
             where TContext : DbContext
         {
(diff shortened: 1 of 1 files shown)


divega:
@HaoK does `IdentityEntityFrameworkServices.GetDefaultServices()` return any service for which it is expected to have more than one registration? If no, then this should be ok to take, don't you think?

PR #506, opened 2015-06-30
Fix external sign in authenticate call and functional tests

HaoK:
Depends on https://github.com/aspnet/Security/pull/310

cc @divega @Tratcher


HaoK:
85d2d3d46a82dbe4d9a01241fcf84f0a619a2bad

PR #510, opened 2015-07-02
Added role id equality check in RemoveClaimsAsync

mhowlett:
I think you're forgetting a role id equality check here.

### DIFF
--- a/src/Microsoft.AspNet.Identity.EntityFramework/RoleStore.cs
+++ b/src/Microsoft.AspNet.Identity.EntityFramework/RoleStore.cs
@@ -278,7 +278,7 @@ namespace Microsoft.AspNet.Identity.EntityFramework
             {
                 throw new ArgumentNullException("claim");
             }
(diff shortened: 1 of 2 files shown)


HaoK:
Ah good catch, mind adding a test so we don't regress this in the future?

mhowlett:
Should be able to find time to do this on the weekend.

mhowlett:
I added an appropriate test (the system seems to have automatically picked that up in this pull request - how clever of it), however it is failing. In the debugger, I can see RemoveClaimAsync is executing as it should, however RoleClaims.Remove(c) doesn't seem to be removing the claims as I thought it would. I don't use Entity Framework, so there is likely a fundamental misunderstanding somewhere on my part... I see dbContext has a method SaveChanges, which I tried calling after attempting to remove the claims, however that made no difference. I think it's better if someone who knows what they're doing with this stuff works out what's going. Sorry, currently rather busy ...

HaoK:
Fair enough, replacing with https://github.com/aspnet/Identity/pull/527

PR #524, opened 2015-07-22
React to Auth/Options changes

HaoK:
Not a huge change to startup

Thoughts? @divega

### DIFF
--- a/src/Microsoft.AspNet.Identity/IdentityOptions.cs
+++ b/src/Microsoft.AspNet.Identity/IdentityOptions.cs
@@ -6,7 +6,7 @@ using System;
 namespace Microsoft.AspNet.Identity
 {
     /// <summary>
-    /// Represents all the options you can used to configure the identity system.
+    /// Represents all the options you can use to configure the identity system.
     /// </summary>
     public class IdentityOptions
     {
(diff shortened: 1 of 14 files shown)


HaoK:
One bit of smell is the static defaults I moved to IdentityCookieOptions, those strings are what we default the AuthenticationSchemes to for the cookie middleware we configure.  Those are then copied to the options, so we have mutable strings.  Its just a bit wonky.

src/Microsoft.AspNet.Identity/IdentityOptions.cs
```diff
@@ -83,52 +83,27 @@ public class IdentityOptions
...
-        /// <summary>
-        /// Gets or sets the authentication type used when constructing an <see cref="ClaimsIdentity"/> from an application cookie.
-        /// </summary>
-        /// <value>The authentication type used when constructing an <see cref="ClaimsIdentity"/> from an application cookie.</value>
-        public static string ApplicationCookieAuthenticationType { get; set; } = typeof(IdentityOptions).Namespace + ".Application.AuthType";
+        public string TwoFactorUserIdCookieAuthenticationScheme { get; set; } = IdentityCookieOptions.TwoFactorUserIdCookieAuthenticationScheme;
 
         /// <summary>
-        /// Gets or sets the authentication type used when constructing an <see cref="ClaimsIdentity"/> from an external identity cookie.
+        /// Gets or sets the scheme used to identify Two Factor authentication cookies for saving the Remember Me state,, must match the scheme in IdentityCookieOptions
```
divega: double comma?

divega:
Something doesn't feel quite right but it is hard for me to explain it. It might be that I just need to understand how it works :smile: Can we go over it on a whiteboard?

HaoK:
Closing and will resubmit with new iteration

PR #527, opened 2015-07-27
Fix bug in RemoveRoleClaim

HaoK:
cc @divega


HaoK:
6931ebe41dedeaf92fa206a0a6134aeafa873480

PR #530, opened 2015-08-02
Fixed UserValidator failing when UserName contains a hyphen

TsengSR:
UserValidator would fail to validate usernames containing an email address with a hyphen in either the address or domain part


HaoK:
Thanks for the PR, but we are going to remove regex entirely soon and will add hypen to the whitelist of characters.

HaoK:
See https://github.com/aspnet/Identity/pull/533

PR #533, opened 2015-08-04
Nuke regex

HaoK:
cc @blowdart @divega

Note you turn this feature off by setting the string to null or empty

### DIFF
--- a/src/Microsoft.AspNet.Identity/UserOptions.cs
+++ b/src/Microsoft.AspNet.Identity/UserOptions.cs
@@ -11,37 +11,12 @@ namespace Microsoft.AspNet.Identity
     public class UserOptions
     {
         /// <summary>
-        /// Initializes a new instance of the <see cref="UserOptions"/> class.
-        /// </summary>
-        public UserOptions()
-        {
-            //User.RequireUniqueEmail = true; // TODO: app decision?
-        }
-
-        /// <summary>
-        /// Gets or sets the regular expression used to validate user names.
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNet.Identity/UserOptions.cs
```diff
@@ -11,37 +11,12 @@ namespace Microsoft.AspNet.Identity
...
         /// </summary>
         /// <value>
-        /// The timeout value used after which user name validation via the <see cref="UserNameValidationRegex"/> will fail if it has not completed.
+        /// The list of allowed characters in the username used to validate user names.
         /// </value>
-        /// <remarks>
-        /// The default value is 20 milliseconds.
-        /// </remarks>
-        public TimeSpan UserNameValidationRegexTimeout { get; set; } = new TimeSpan(0,0,0,0,20);
+        public string AllowedUserNameCharacters { get; set; } = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-._@";
```
divega: I feel neglected that `áéíóúñÑüÜ` didn't make kit but if this is what we always had an no one complained, I guess it is ok :smile:

HaoK: I don't see any special characters in your username :). Or did it get englisized?

divega:
:shipit:

HaoK:
b2eb5feb8c26fc07f2794e32a94ebdd7cc805df5

PR #535, opened 2015-08-06
Add constructors to IdentityDbContext to allow dependency injection. …

danielcweber:
…The other approach (overriding 'OnConfiguring') does not work well with unit testing.

### DIFF
--- a/src/Microsoft.AspNet.Identity.EntityFramework/IdentityDbContext.cs
+++ b/src/Microsoft.AspNet.Identity.EntityFramework/IdentityDbContext.cs
@@ -3,6 +3,7 @@
 
 using System;
 using Microsoft.Data.Entity;
+using Microsoft.Data.Entity.Infrastructure;
 using Microsoft.Data.Entity.Metadata;
 
 namespace Microsoft.AspNet.Identity.EntityFramework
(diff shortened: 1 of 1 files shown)


HaoK:
@divega

src/Microsoft.AspNet.Identity.EntityFramework/IdentityDbContext.cs
```diff
@@ -17,6 +18,26 @@ public class IdentityDbContext<TUser, TRole, TKey> : DbContext
...
+        {
+            
+        }
+
+        public IdentityDbContext(IServiceProvider serviceProvider, DbContextOptions options) : base(serviceProvider, options)
+        {
+            
+        }
+
+        protected IdentityDbContext()
```
divega: I think this should also do `:base()`.

divega: Actually, I am pretty sure.

HaoK: Isn't :base() implied?

HaoK: I say this because I personally have never done an explicit base() call

divega: Do'h, off course you are right.

divega:
@HaoK what kind of test coverage do you think we should add for these constructors?

Issue #519, opened 2015-07-14
Custom IdentityErrorDescriber constructor hits at-least 7 times per request


imranbaloch:
When we create a custom IdentityErrorDescriber class and register it then it constructor hit at-least 7 times per request.

HaoK:
How are you registering it in DI, as Transient?

imranbaloch:
Yes doing the same that you guys  doing https://github.com/aspnet/Identity/blob/715790c2134d5bd8904da65877c1fd4bce4b8035/src/Microsoft.AspNet.Identity/IdentityServiceCollectionExtensions.cs#L92

HaoK:
This behavior is directly related to the fact that its Transient, every time its asked for, you get a new instance.  Our default implementation is stateless.  You can just register it as Scoped if you wanted only a single instance to be created for each request.

davidfowl:
Should it be transient? If it's stateless why not make it a singleton?

HaoK:
We've decided to go ahead and switch to Scoped lifetimes for all identity services that are currently transient, we felt this was the best compromise

HaoK:
https://github.com/aspnet/Identity/pull/536

gdoron:
This doesn't effect the DbContext, right?
If I set my DbContext to be Transient, it will be transient for Identity too, right?

When does the UserManager/LoginManager etc. are getting the context? when created or inside of each method call?

HaoK:
This change doesn't affect the DbContext no, the identity managers are scoped by default so each should only be created once per request.  The manager take a UserStore which is also scoped by default, those take a DbContext.

PR #536, opened 2015-08-06
Used scoped instead of transient lifetimes

HaoK:
Fixes https://github.com/aspnet/Identity/issues/519

cc @divega


HaoK:
0e77bbdbbd6917c7f5dce013c4c267d971c8e0f2

PR #541, opened 2015-08-11
React to cookie/options changes

HaoK:
Now we keep 4 instances of CookieOptions which we pull from ApplicationServices during UseIdentity instead of using named options.

cc @divega

### DIFF
--- a/samples/IdentitySample.Mvc/Startup.cs
+++ b/samples/IdentitySample.Mvc/Startup.cs
@@ -47,21 +47,6 @@ namespace IdentitySamples
                     .AddEntityFrameworkStores<ApplicationDbContext>()
                     .AddDefaultTokenProviders();
 
-            services.ConfigureFacebookAuthentication(options =>
-            {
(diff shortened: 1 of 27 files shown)


samples/IdentitySample.Mvc/Startup.cs
```diff
@@ -82,9 +67,21 @@ public void Configure(IApplicationBuilder app, ILoggerFactory loggerFactory)
             app.UseErrorPage()
                .UseStaticFiles()
                .UseIdentity()
-               .UseFacebookAuthentication()
-               .UseGoogleAuthentication()
-               .UseTwitterAuthentication()
+               .UseFacebookAuthentication(options =>
```
HaoK: Note this change is optional now, either way works (via ConfigureServices or via Use)

HaoK:
Ugh I think our TokenProviders are broken by this change actually, we were relying on Named options to name the instances before, need to find another way to do it

HaoK:
Replacing

PR #544, opened 2015-08-14
Nuke generics from validators/token provider interfaces

HaoK:
Generics moved to the methods instead.

Prep for changing how registration works

cc @divega


divega:
__TL;DR:__ I am still worried that the change causes friction/weirdness to application-specific implementers of the interfaces. We might convince ourselves that this is ok, but I would like to be more sure that we really need to do it.

__Details:__ with the `TUser` generic argument at the interface level as we have it right now it is easy for an application developer to create a custom implementation of one of these interfaces for a specific `TUser` which contains code that accesses properties on the user instance.

With this change the specific `TUser` will be unknown at compile time and remain unknown until the method is called hence the interface implementation won't be able to bind directly to properties on the user instance. You can still do it with an `as` conversion, using dynamic, reflection, etc., but it is not as intuitive and it becomes pretty weird. 

Note that it is by design that in our own implementations of these interfaces we will never need to "crack-open" the `TUser` instance, i.e. we just need the `TUser` of the `UserManager<TUser>` and the type of the user instance to match so that we can pass the latter to the former. However we can't assume applications-specific implementations wouldn't want to do it.

HaoK:
Its not great but we can add a base class to make this feel similar to what we have today.  That hides the cast from them

```
    public abstract class UserValidatorBase<TUser> : IUserValidator where TUser : class
    {
        Task<IdentityResult> IUserValidator.ValidateAsync<TUser1>(UserManager<TUser1> manager, TUser1 user)
        {
            var typedManager = manager as UserManager<TUser>;
            var typedUser = user as TUser;
            if (typedManager == null || typedUser == null)
            {
                throw new InvalidOperationException();
            }
            return ValidateAsync(typedManager, typedUser);
        }

        public abstract Task<IdentityResult> ValidateAsync(UserManager<TUser> manager, TUser user);
    }
```

HaoK:
Replacing with new PR

Issue #511, opened 2015-07-02
AddIdentity will add a marker Identity service and UseIdentity will throw if not found


HaoK:
Similar to what MVC does

HaoK:
ab29d5f5e581dd4cf9a8996b8c940bd4e5b65490

PR #548, opened 2015-08-20
AddIdentity is now required for UseIdentity

HaoK:
Fixes https://github.com/aspnet/Identity/issues/511

@divega

### DIFF
--- a/src/Microsoft.AspNet.Identity/IdentityServiceCollectionExtensions.cs
+++ b/src/Microsoft.AspNet.Identity/IdentityServiceCollectionExtensions.cs
@@ -83,18 +83,19 @@ namespace Microsoft.Framework.DependencyInjection
             services.AddAuthentication();
 
             // Identity services
-            services.TryAdd(ServiceDescriptor.Scoped<IUserValidator<TUser>, UserValidator<TUser>>());
-            services.TryAdd(ServiceDescriptor.Scoped<IPasswordValidator<TUser>, PasswordValidator<TUser>>());
-            services.TryAdd(ServiceDescriptor.Scoped<IPasswordHasher<TUser>, PasswordHasher<TUser>>());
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNet.Identity/IdentityServiceCollectionExtensions.cs
```diff
@@ -83,18 +83,19 @@ public static IServiceCollection ConfigureIdentityApplicationCookie(this IServic
             services.AddAuthentication();
 
             // Identity services
-            services.TryAdd(ServiceDescriptor.Scoped<IUserValidator<TUser>, UserValidator<TUser>>());
-            services.TryAdd(ServiceDescriptor.Scoped<IPasswordValidator<TUser>, PasswordValidator<TUser>>());
-            services.TryAdd(ServiceDescriptor.Scoped<IPasswordHasher<TUser>, PasswordHasher<TUser>>());
-            services.TryAdd(ServiceDescriptor.Scoped<ILookupNormalizer, UpperInvariantLookupNormalizer>());
-            services.TryAdd(ServiceDescriptor.Scoped<IRoleValidator<TRole>, RoleValidator<TRole>>());
+            services.TryAddSingleton<IdentityMarkerService>();
```
divega: I think it would be slightly better to make it a transient so that it is not parked in memory forever :smile:

divega:
:shipit:

src/Microsoft.AspNet.Identity/IdentityServiceCollectionExtensions.cs
HaoK: I copied what Mvc does, they stick it as a singleton so I just mimiced it:

https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNet.Mvc.Core/DependencyInjection/MvcCoreServiceCollectionExtensions.cs#L124

HaoK:
ab29d5f5e581dd4cf9a8996b8c940bd4e5b65490

Issue #493, opened 2015-06-14
AddUserManager<TUserManager> should register TUserManager as a service


PartyArk:
If I create a custom UserManager
```
public class MyUserManager : UserManager<User>
{
    public MyUserManager(IUserStore<User> store) : base(store, ...) { }
}
```
... and then register it in my Startup class:
```
services
   .AddIdentity<ApplicationUser, IdentityRole>()
   .AddUserManager<MyUserManager >()
```

I would then should expect to have it usable in my Controllers I think?

 ```
public AccountController(MyUserManager userManager, SignInManager<User> signInManager)
{
            this.userManager = userManager;
            this.signInManager = signInManager;
}
```
But this won't work: AddUserManager registers the base Interface with the derived class as its implementation.  So you can overwrite the methods of UserManager, but not extend it. The only way to make it work is to register it directly in Startup.cs - 

```
services.AddScoped<MyUserManager>()
[or]
services.AddScoped(typeof(MyUserManager))
```

I suggest that it should be fully extendable, rather than just an alternative implementation.

**Comment by rustd** [comment]
@divega

**Comment by Bartmax** [comment]
https://github.com/aspnet/Identity/issues/215

**Comment by divega** [comment]
@PartyArk With our current implementation this is the expected behavior: `AddUserManager<TUserManager>()` will register the type you provide as the implementation of the UserManager<TUser> service. It will __not__ registerer your custom user manager type as a service type so you won't be able to resolve it from DI. 

Also, your workaround may not do exactly what you expect, e.g.  AddScoped<MyUserManager>() is registering your custom user manager as its own service implementation, so when you resolve it, you will get a completely separate instance from the one that Identity is using for the same request. This might not matter a lot if your custom user manager doesn't hold to any state besides the store, but in the general case it is not safe.

Here is an alternative version of the workaround that should return the same instance regardless of whether you try to resolve `MyUserManager` or `UserMananger<User>`:

``` C#
services.AddScoped<MyUserManager>(p => p.GetRequiredService<UserManager<User>>());
``` 

Now, we have three options here:

1. Not change the behavior but make sure we document it clearly
2. Add the workaround inside the `AddUserManager<TUserManager>()` method so that is just works the way you expected it to work
3. Remove the `AdduserManager<TUserManager>()` method altogether

Note that the reasons we believe the third option might be acceptable are that:

[...]

PR #550, opened 2015-08-21
AddManager behavior changes

HaoK:
Fixes https://github.com/aspnet/Identity/issues/493

@divega

### DIFF
--- a/src/Microsoft.AspNet.Identity/IdentityBuilder.cs
+++ b/src/Microsoft.AspNet.Identity/IdentityBuilder.cs
@@ -159,7 +159,9 @@ namespace Microsoft.AspNet.Identity
         /// <returns>The <see cref="IdentityBuilder"/>.</returns>
         public virtual IdentityBuilder AddUserManager<TUserManager>() where TUserManager : class
         {
-            return AddScoped(typeof(UserManager<>).MakeGenericType(UserType), typeof(TUserManager));
+            var userManagerType = typeof(UserManager<>).MakeGenericType(UserType);
+            Services.AddScoped(typeof(TUserManager), services => services.GetRequiredService(userManagerType));
+            return AddScoped(userManagerType, typeof(TUserManager));
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/Microsoft.AspNet.Identity/IdentityBuilder.cs
```diff
@@ -159,7 +159,9 @@ public virtual IdentityBuilder AddDefaultTokenProviders()
         /// <returns>The <see cref="IdentityBuilder"/>.</returns>
         public virtual IdentityBuilder AddUserManager<TUserManager>() where TUserManager : class
```
divega: This makes me wonder what happens if the user passes UserManager<UserType> as the argument (I imagine it could cause a cycle in DI) and also in general some TUserManager that does not derive from UserManager<UserType> (I imagine it would fail later on but the exception could be not that great :smile:). 

I remember we had to not have generic constraints in this method because we had to make IdentityBuilder non-generic but I am wondering if we should add more runtime type checks to throw nicer exceptions.

divega:
Other than the feedback on adding type checks looks good.

src/Microsoft.AspNet.Identity/IdentityBuilder.cs
HaoK: Updated to throw if same type or non derived type is passed in

src/Microsoft.AspNet.Identity/Resources.resx
```diff
@@ -145,6 +145,10 @@
     <value>Email '{0}' is invalid.</value>
     <comment>Invalid email</comment>
   </data>
+  <data name="InvalidManagerType" xml:space="preserve">
+    <value>Type {0} must be derived from {1}&lt;{2}&gt;.</value>
```
divega: "Must derive from"?

divega:
:shipit: with minor feedback on the excepción message.

HaoK:
4e0000163ba5067dcbe79a33f80192623fc4d5f3

PR #553, opened 2015-08-27
React to Options/Cookie changes

HaoK:
- New IdentityCookieOptions sub option which holds the options for the 4 cookie options instances
- New TokenOptions which contains the providerName to IUserTokenProvider type map, which UserManager will use to get the actual instances for its runtime map
- IdentityBuilder now adds TokenProvider types as transients in addition to updating TokenOptions

cc @divega

### DIFF
--- a/samples/IdentitySample.Mvc/Startup.cs
+++ b/samples/IdentitySample.Mvc/Startup.cs
@@ -47,21 +47,6 @@ namespace IdentitySamples
                     .AddEntityFrameworkStores<ApplicationDbContext>()
                     .AddDefaultTokenProviders();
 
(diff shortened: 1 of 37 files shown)


samples/IdentitySample.Mvc/Startup.cs
```diff
@@ -47,21 +47,6 @@ public void ConfigureServices(IServiceCollection services)
                     .AddEntityFrameworkStores<ApplicationDbContext>()
```
HaoK: I still need to go and fix the security middleware Configure methods to be called Add

HaoK:
2113ef5abcb64bf1c49d8485bcfae45b703f4b78

Issue #557, opened 2015-09-03
Use constants in UserManager for token purposes


hvanbakel:
Currently, methods like `GenerateEmailConfirmationTokenAsync` and `ConfirmEmailAsync` in https://github.com/aspnet/Identity/blob/dev/src/Microsoft.AspNet.Identity/UserManager.cs use strings as purpose for generating a token. Say I want to override one of these methods, it would be helpful if I could just reuse the purpose. 
If this would be exposed as a const the code would be cleaner anyway since there are no 2 magic values. Obviously this also goes for the other tokens.

HaoK:
We will expose them as protected const string ResetPasswordPurpose/ConfirmEmailPurpose etc.

PR #560, opened 2015-09-10
Switch to protected constants for token purposes

HaoK:
Fixes https://github.com/aspnet/Identity/issues/557

cc @divega


divega:
:shipit:

Issue #409, opened 2015-03-16
Review if a package description is required for these packages


Praburaj:
```
Identity\src\Microsoft.AspNet.Identity
Identity\src\Microsoft.AspNet.Identity.EntityFramework
```

HaoK:
@rustd What description do we want?

rustd:
Core implementation for ASP.NET Identity.
ASP.NET Identity providers that use Entity Framework.

HaoK:
df1fe83992088fcdc831946607808a5f18650f78

PR #571, opened 2015-09-21
Enabling NuGetPackageVerifier

ajaybhargavb:
@Eilon


divega:
Does the test pass with this package? If no, and the package description is required, please see #409 for the right strings.

ajaybhargavb:
It passes. Note that the `RequiredPackageMetadataRule` is currently not being run.It will be enabled later.

divega:
Thanks. Looks :shipit: to me.

Eilon:
:shipit:

PR #575, opened 2015-09-28
Tweak IsInRole check

HaoK:
cc @divega

### DIFF
--- a/src/Microsoft.AspNet.Identity/UserManager.cs
+++ b/src/Microsoft.AspNet.Identity/UserManager.cs
@@ -1061,7 +1061,7 @@ namespace Microsoft.AspNet.Identity
                 throw new ArgumentNullException("roles");
             }
 
-            foreach (var role in roles)
+            foreach (var role in roles.Distinct())
             {
                 if (await userRoleStore.IsInRoleAsync(user, role, CancellationToken))
                 {
--- a/src/Microsoft.AspNet.Identity/UserManager.cs
+++ b/src/Microsoft.AspNet.Identity/UserManager.cs
@@ -1031,8 +1031,7 @@ namespace Microsoft.AspNet.Identity
                 throw new ArgumentNullException("user");
             }
 
-            var userRoles = await userRoleStore.GetRolesAsync(user, CancellationToken);
-            if (userRoles.Contains(role))
+            if (await userRoleStore.IsInRoleAsync(user, role, CancellationToken))
             {
                 return await UserAlreadyInRoleError(user, role);
             }
--- a/src/Microsoft.AspNet.Identity/UserManager.cs
+++ b/src/Microsoft.AspNet.Identity/UserManager.cs
@@ -205,7 +205,6 @@ namespace Microsoft.AspNet.Identity
         /// true if the backing user store supports user emails, otherwise false.
         /// </value>
         public virtual bool SupportsUserEmail 
-
         {
             get
             {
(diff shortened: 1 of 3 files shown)


src/Microsoft.AspNet.Identity/UserManager.cs
```diff
@@ -1063,10 +1062,9 @@ public virtual Task<IdentityResult> RemoveClaimAsync(TUser user, Claim claim)
                 throw new ArgumentNullException("roles");
             }
 
-            var userRoles = await userRoleStore.GetRolesAsync(user, CancellationToken);
             foreach (var role in roles)
             {
-                if (userRoles.Contains(role))
+                if (await userRoleStore.IsInRoleAsync(user, role, CancellationToken))
```
divega: Should this change also go in the `AddToRoleAsync()` method (singular) above?

src/Microsoft.AspNet.Identity/UserManager.cs
```diff
@@ -1063,10 +1062,9 @@ public virtual Task<IdentityResult> RemoveClaimAsync(TUser user, Claim claim)
                 throw new ArgumentNullException("roles");
             }
 
-            var userRoles = await userRoleStore.GetRolesAsync(user, CancellationToken);
             foreach (var role in roles)
```
divega: Should we do `roles.Distinct()` here to prevent duplicate roles from being added? I haven't looked very deep in `IsInRoleAsync()` but I assume it is possible it won't return true until things are saved to the database.

HaoK: My guess is I was trying to avoid calling the store N times to do this check before, so this actually does fix and issue with AddRoles containing dupes itself, that was in the code before.  

Here's what IsInRole does, AnyAsync should pickup any prior updates correct?:

```
            var role = await Roles.SingleOrDefaultAsync(r => r.Name.ToUpper() == roleName.ToUpper(), cancellationToken);
            if (role != null)
            {
                var userId = user.Id;
                var roleId = role.Id;
                return await UserRoles.AnyAsync(ur => ur.RoleId.Equals(roleId) && ur.UserId.Equals(userId));
            }
```

src/Microsoft.AspNet.Identity/UserManager.cs
HaoK: Updated the other overload as well

src/Microsoft.AspNet.Identity/UserManager.cs
divega: I still don't see what would prevent `await userManager.AddToRolesAsync(user, new [] {"admin", "admin"})` from trying to add "admin" twice, e.g. the LINQ queries above won't return UserRoles that haven't been saved to the database, unless we save for each iteration it won't help. But if you say it handles it, I am good :smile:

HaoK: Well UserRoles is a db set and I'm guessing Any is checking against the modified (but not saved) set of roles which is why it catches it.  So we are just relying on the store to do the right thing in the manager.  We could explicitly filter with a Distinct if you want.  Not a big deal, just a question if we want an exception or a silent distinct.

divega: A LINQ query always executes as a database query in EF and things in memory won't have any influence on what entities are included in the results. See https://github.com/aspnet/EntityFramework/issues/2982 in the backlog. I think we should take the chance to add the .Distinct() while we are changing this.

divega: It also makes it more robust for other providers.

divega:
:shipit: besides the feedback on using `Distinct()` on `AddToRolesAsync()`.

HaoK:
d60bcac9b436b84a55b0fb58fa75519a81d7e4cb

PR #591, opened 2015-10-16
Reenable tests/cascading deletes

HaoK:
cc @divega 

Looks like we were a bit agressive in disabling tests to fix travis before :)  Turns out ConditionalTheory with no InlineData no-ops the tests...  I noticed this only because the new cascading tests weren't failing doh!

### DIFF
--- a/src/Microsoft.AspNet.Identity.EntityFramework/IdentityDbContext.cs
+++ b/src/Microsoft.AspNet.Identity.EntityFramework/IdentityDbContext.cs
@@ -20,22 +20,22 @@ namespace Microsoft.AspNet.Identity.EntityFramework
     {
         public IdentityDbContext(DbContextOptions options) : base(options)
         {
-            
+
         }
 
         public IdentityDbContext(IServiceProvider serviceProvider) : base(serviceProvider)
         {
-            
+
         }
 
         public IdentityDbContext(IServiceProvider serviceProvider, DbContextOptions options) : base(serviceProvider, options)
         {
-            
+
         }
 
         protected IdentityDbContext()
         {
-            
+
         }
 
         public DbSet<TUser> Users { get; set; }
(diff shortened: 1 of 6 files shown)


src/Microsoft.AspNet.Identity.EntityFramework/IdentityDbContext.cs
```diff
@@ -48,62 +48,59 @@ protected IdentityDbContext()
...
+                b.HasIndex(u => u.NormalizedUserName).HasName("UserNameIndex");
+                b.HasIndex(u => u.NormalizedEmail).HasName("EmailIndex");
+                b.ToTable("AspNetUsers");
+                b.Property(u => u.ConcurrencyStamp).IsConcurrencyToken();
+
+                b.Property(u => u.UserName).HasMaxLength(256);
+                b.Property(u => u.NormalizedUserName).HasMaxLength(256);
+                b.Property(u => u.Email).HasMaxLength(256);
+                b.Property(u => u.NormalizedEmail).HasMaxLength(256);
+                b.HasMany(u => u.Claims).WithOne().HasForeignKey(uc => uc.UserId).OnDelete(DeleteBehavior.Cascade);
```
divega: It doesn't hurt to have it explicitly but FWIW if these FKs are not nullable cascade will be the default behavior.

src/Microsoft.AspNet.Identity.EntityFramework/IdentityDbContext.cs
```diff
@@ -48,62 +48,59 @@ protected IdentityDbContext()
...
             builder.Entity<IdentityUserRole<TKey>>(b =>
-                {
-                    b.HasKey(r => new { r.UserId, r.RoleId });
-                    b.ToTable("AspNetUserRoles");
-                });
-            // Blocks delete currently without cascade
-            //.ForeignKeys(fk => fk.ForeignKey<TUser>(f => f.UserId))
-            //.ForeignKeys(fk => fk.ForeignKey<TRole>(f => f.RoleId));
+            {
+                b.HasKey(r => new { r.UserId, r.RoleId });
```
divega: Just to make sure, what cascade behavior do you want this to have?

HaoK: I want all subentities to be deleted if the root object goes away (user => roles/claims/logins role => roles)

src/Microsoft.AspNet.Identity.EntityFramework/IdentityDbContext.cs
HaoK: Well I didn't go out of my way to add these :)

Tests fail like so if they aren't there:
```
      System.InvalidOperationException : The association between entity types 'StringUser' and 'IdentityUserRole<string>' has been severed but the foreign key for this relationship cannot be set to null. If the dependent entity should be deleted, then setup the relationship to use cascade deletes.
```

HaoK: I guess the question is are we missing some setup on the UserRole table to require non nullable keys so this works implied?

divega: Don't know what I am missing but I would expect it to just cascade. @ajcvickers do you see any reason it wouldn't?

src/Microsoft.AspNet.Identity.EntityFramework/IdentityDbContext.cs
divega: Ah, yeah. I was missing this is configured on the other side.

src/Microsoft.AspNet.Identity.EntityFramework/IdentityDbContext.cs
divega: I believe the FKs on UserRole are part of the PK and therefore should be implicitly required, but maybe it is not working that way.

divega:
There are some question to follow on for cascade delete behavior in EF but this LGTM. :shipit:

src/Microsoft.AspNet.Identity.EntityFramework/IdentityDbContext.cs
ajcvickers: @divega @HaoK From what I can tell its going to depend on the type of TKey for some cases. If TKey is a non-nullable type, then all relationships must required because TKey is used for FK properties making them all non-nullable. So in this case all relationships should have cascade delete turned on.

Assuming TKey is nullable--say string, then it looks like both User -> Claims and User => Logins will be optional relationships because the FK is not part of the PK and nothing makes the relationship required. So these two will not get cascade delete by default. User -> Roles should still be required because the FK is part of the primary key and therefore must be required.

Therefore, I would expect two of these three relationship to need OnDelete and the other one not to need it.

However, I don't know much about how the model is supposed to work, but it would seem strange to me that the behavior of the model changes based on the type of TKey, so maybe the relationships should always be marked as required rather than letting it depend on the type of TKey.

divega: That is a very good point! We need to make these relationships explicitly required regardless of whether `TKey` is nullable. @HaoK please add a call to `.IsRequired()` for `user.Logins`, `user.Claims` and `role.Claims`. 

After that please also try again `user.Roles` and `role.Users` without the `OnDelete()`. If it is needed we might have missed a bug in EF as those relationships should be implicitly required.

HaoK: Cool IsRequired does the trick much nicer :)

ajcvickers: @HaoK Does User -> Roles need to have IsRequired? It looks like it should be required by default.

HaoK:
c847f9c50df478ef826f6d4781ddcd760325da78

PR #619, opened 2015-11-05
Added doc comments

blowdart:


HaoK:
@divega We want these for RC1?  So do you want me to merge these into release?

src/Microsoft.AspNet.Identity.EntityFramework/IdentityRole.cs
```diff
@@ -7,22 +7,28 @@
...
 {
     /// <summary>
-    ///     Represents a Role entity
+    /// The default implementation of <see cref="IdentityRole{TKey}"/> which uses a string as the primary key.
     /// </summary>
     public class IdentityRole : IdentityRole<string>
     {
         /// <summary>
-        ///     Constructor
+        /// Initializes a new instance of <see cref="IdentityRole{TKey}"/>.
```
divega: This is the non generic version.

src/Microsoft.AspNet.Identity.EntityFramework/IdentityRole.cs
```diff
@@ -7,22 +7,28 @@
...
+        /// The Id property is initialized to from a new GUID string value.
+        /// </remarks>
         public IdentityRole()
         {
             Id = Guid.NewGuid().ToString();
         }
 
         /// <summary>
-        ///     Constructor
+        /// Initializes a new instance of <see cref="IdentityRole{TKey}"/>.
```
divega: Non generic.

src/Microsoft.AspNet.Identity.EntityFramework/IdentityUser.cs
```diff
@@ -6,108 +6,155 @@
 
 namespace Microsoft.AspNet.Identity.EntityFramework
 {
+    /// <summary>
+    /// The default implementation of <see cref="IdentityUser{TKey}"/> which uses a string as a primary key.
+    /// </summary>
     public class IdentityUser : IdentityUser<string>
     {
+        /// <summary>
+        /// Initializes a new instance of <see cref="IdentityUser{TKey}"/>.
```
divega: Non generic

src/Microsoft.AspNet.Identity.EntityFramework/IdentityUser.cs
```diff
@@ -6,108 +6,155 @@
...
+        /// <remarks>
+        /// The Id property is initialized to from a new GUID string value.
+        /// </remarks>
         public IdentityUser()
         {
             Id = Guid.NewGuid().ToString();
         }
 
+        /// <summary>
+        /// Initializes a new instance of <see cref="IdentityUser{TKey}"/>.
```
divega: Non generic

src/Microsoft.AspNet.Identity.EntityFramework/UserStore.cs
```diff
@@ -50,6 +70,11 @@ public class UserStore<TUser, TRole, TContext, TKey> :
         where TKey : IEquatable<TKey>
     {
 
+        /// <summary>
+        /// Creates a new instance of <see cref="UserStore"/>.
+        /// </summary>
+        /// <param name="context">The context used to access the store.</param>
+        /// <param name="describer">The <see cref="IdentityErrorDescriber"/> to place store errors into.</param>
```
divega: Not sure this is the best description. Perhaps "used to obtain descriptions for errors returned by the store"? @haok?

src/Microsoft.AspNet.Identity.EntityFramework/UserStore.cs
```diff
@@ -263,11 +364,11 @@ public virtual Task SetPasswordHashAsync(TUser user, string passwordHash, Cancel
         }
 
         /// <summary>
-        ///     Get the password hash for a user
+        /// Gets the password hash for a user, as an asynchronous operation.
```
divega: Does the "as an async operation" part add much on all these?

src/Microsoft.AspNet.Identity.EntityFramework/IdentityEntityFrameworkServices.cs
```diff
@@ -11,7 +11,7 @@ namespace Microsoft.AspNet.Identity
     /// <summary>
     /// Default services
     /// </summary>
-    public class IdentityEntityFrameworkServices
+    internal class IdentityEntityFrameworkServices
```
divega: I don't think we should make this internal. @haok didn't we get rid of this pattern in other repos?

divega:
Very nice! Only minor comments/questions.

src/Microsoft.AspNet.Identity.EntityFramework/IdentityEntityFrameworkServices.cs
HaoK: Hence why I asked him to make it internal, we got rid of the public pattern yes.  But we call this code in two places (with a different key type), so its useful as a helper method.  I can move it into the Builder extension class as a private instead, but the idea was to make this go away from a public API perspective

src/Microsoft.AspNet.Identity.EntityFramework/UserStore.cs
HaoK: Yeah just seems like extra verbiage.

src/Microsoft.AspNet.Identity.EntityFramework/IdentityEntityFrameworkServices.cs
divega: Private method sounds good. I am fine with doing it in two steps.

HaoK:
Updated

HaoK:
379f1928a4d676d2ad14ba026f50b204ed33d518

PR #641, opened 2015-11-19
Fix functional unit test after hosting changes

pakrym:
@Tratcher

### DIFF
--- a/src/Microsoft.AspNet.Identity/IdentityServiceCollectionExtensions.cs
+++ b/src/Microsoft.AspNet.Identity/IdentityServiceCollectionExtensions.cs
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Http.Internal;
 using Microsoft.AspNet.Identity;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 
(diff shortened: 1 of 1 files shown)


Tratcher:
@HaoK

Tratcher:
This probably needs to be a production dependency, not a test dependency.

pakrym:
:up: :date:

HaoK:
:shipit:

src/Microsoft.AspNet.Identity/IdentityServiceCollectionExtensions.cs
```diff
@@ -2,6 +2,8 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using Microsoft.AspNet.Http;
+using Microsoft.AspNet.Http.Internal;
```
HaoK: Why do we need both Http and Internal?

src/Microsoft.AspNet.Identity/IdentityServiceCollectionExtensions.cs
```diff
@@ -49,6 +51,8 @@ public static class IdentityServiceCollectionExtensions
                 options.SignInScheme = new IdentityCookieOptions().ExternalCookieAuthenticationScheme;
             });
 
+            // Hosting doesn't add IHttpContextAccessor byt default
```
HaoK: nit: typo

src/Microsoft.AspNet.Identity/IdentityServiceCollectionExtensions.cs
pakrym: Yep  HttpContextAccessor  is in internal

PR #649, opened 2015-11-23
Initial Interop45 sample app

HaoK:
cc @divega 

Initial sideby side sample for Identity 2.0 App configured to interop.

Migration code added under Compat directory.  

Note: NormalizedUserName/Emails are not updated automatically currently.  Need to wireup OnSaveChanges?


samples/Interop45/Compat/CompatRole.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        /// <summary>
+        ///     Normalized role name
+        /// </summary>
+        public virtual string NormalizedName { get; set; }
+
+        /// <summary>
+        ///     Concurrency stamp 
+        /// </summary>
+        public virtual string ConcurrencyStamp { get; set; } = Guid.NewGuid().ToString();
```
divega: I am not sure this work. Do we need to also change the current value on `SaveChanges()`?

HaoK: Probably not. I copied this from codeplex as is. I'm guessing it's broken in the same way that normalized user name is.

samples/IdentitySample.Mvc/Migrations/20151110211659_Identity2-3.cs
```diff
@@ -0,0 +1,155 @@
+using System;
+using System.Collections.Generic;
+using Microsoft.Data.Entity.Migrations;
+using Microsoft.Data.Entity.Metadata;
+
+namespace IdentitySample.Mvc.Migrations
+{
+    public partial class Identity23 : Migration
```
divega: Can we improve the migration name? E.g. UpgradeFromIdentity2To3?

samples/IdentitySample.Mvc/Migrations/20151110211659_Identity2-3.cs
```diff
@@ -0,0 +1,155 @@
...
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.CreateTable(
```
divega: I may be thinking at it the wrong way, but I am surprised to see all the tables created. It looks more like an initial migration. If this is supposed to be the upgrade migration these should be alters that add a few columns.

samples/IdentitySample.Mvc/Migrations/20151110211659_Identity2-3.cs
```diff
@@ -0,0 +1,155 @@
+using System;
+using System.Collections.Generic;
+using Microsoft.Data.Entity.Migrations;
```
divega: Ah, this looks like a migration done using EF7. Weren't you going to use EF6 to upgrade the database schema?

divega:
Besides the comments I already made:
1. I like the idea of fixing things up on `SaveChanges()`. It should work for normalization of user name and email and also for the concurrency token. 
2. This is a lot of sample code to look at so while reviewing I had to assume that it would be possible that it should be easy to get this code by using one of the default templates for MVC5, then making a few changes and then following whatever the migration steps are. That makes me think that it would be very useful to include a readme file with those steps alongside the sample code.

divega:
@HaoK are you planning to make an initial commit and the iterate over it or would you like to get it all fixed in this review?

Personally I think that pushig this even if it is not completely correct would be good as there is a lot of boilerplate code that I would prefer not to look at in subsequent reviews :smile:

HaoK:
Yeah just looking for major stuff. I'm happy to push an initial mostly functional iteration.

divega:
Ok, can you please answer if using EF7 for the migration was intentional? Otherwise :shipit:

HaoK:
Just old artifacts there with ef7 migrations

samples/IdentitySample.Mvc/Migrations/20151110211659_Identity2-3.cs
HaoK: Yeah this was just old files leftover, removing them from the PR now

HaoK:
267d0ac20671ec80ae0d45c1988f1347a2dff0b9 will send new PR with next iteration

PR #661, opened 2015-12-04
Make HttpContextAccessor Optional

HaoK:
cc @divega I think we should do this for most of the other services which we are adding defaults to via AddIdentity, the ctor should really only take the UserStore service

### DIFF
--- a/src/Microsoft.AspNet.Identity/SignInManager.cs
+++ b/src/Microsoft.AspNet.Identity/SignInManager.cs
@@ -22,6 +22,7 @@ namespace Microsoft.AspNet.Identity
     {
         private const string LoginProviderKey = "LoginProvider";
         private const string XsrfKey = "XsrfId";
+        private IHttpContextAccessor _contextAccessor;
 
         /// <summary>
         /// Creates a new instance of <see cref="SignInManager{TUser}"/>.
@@ -41,7 +42,7 @@ namespace Microsoft.AspNet.Identity
             {
                 throw new ArgumentNullException(nameof(userManager));
             }
-            if (contextAccessor == null || contextAccessor.HttpContext == null)
+            if (contextAccessor == null)
             {
                 throw new ArgumentNullException(nameof(contextAccessor));
             }
@@ -51,7 +52,7 @@ namespace Microsoft.AspNet.Identity
             }
 
             UserManager = userManager;
-            Context = contextAccessor.HttpContext;
+            _contextAccessor = contextAccessor;
             ClaimsFactory = claimsFactory;
             Options = optionsAccessor?.Value ?? new IdentityOptions();
             Logger = logger;
@@ -65,9 +66,10 @@ namespace Microsoft.AspNet.Identity
         /// </value>
         protected internal virtual ILogger Logger { get; set; }
         protected internal UserManager<TUser> UserManager { get; set; }
-        internal HttpContext Context { get; set; }
         internal IUserClaimsPrincipalFactory<TUser> ClaimsFactory { get; set; }
         internal IdentityOptions Options { get; set; }
+
+        internal HttpContext Context => _contextAccessor.HttpContext;
 
         /// <summary>
         /// Creates a <see cref="ClaimsPrincipal"/> for the specified <paramref name="user"/>, as an asynchronous operation.
--- a/src/Microsoft.AspNet.Identity/UserManager.cs
+++ b/src/Microsoft.AspNet.Identity/UserManager.cs
@@ -63,7 +63,6 @@ namespace Microsoft.AspNet.Identity
             }
             Store = store;
             Options = optionsAccessor?.Value ?? new IdentityOptions();
-            _context = contextAccessor?.HttpContext;
             PasswordHasher = passwordHasher;
             KeyNormalizer = keyNormalizer;
             ErrorDescriber = errors;
@@ -86,6 +85,7 @@ namespace Microsoft.AspNet.Identity
 
             if (services != null)
             {
+                _context = services.GetService<IHttpContextAccessor>()?.HttpContext;
                 foreach (var providerName in Options.Tokens.ProviderMap.Keys)
                 {
                     var provider = services.GetRequiredService(Options.Tokens.ProviderMap[providerName].ProviderType) as IUserTokenProvider<TUser>;
--- a/src/Microsoft.AspNet.Identity/UserManager.cs
+++ b/src/Microsoft.AspNet.Identity/UserManager.cs
@@ -72,10 +72,6 @@ namespace Microsoft.AspNet.Identity
                     UserValidators.Add(v);
                 }
             }
-            else
-            {
-                UserValidators.Add(new UserValidator<TUser>());
-            }
             var passwordValidators = services?.GetService<IEnumerable<IPasswordValidator<TUser>>>();
             if (passwordValidators != null)
             {
@@ -83,10 +79,6 @@ namespace Microsoft.AspNet.Identity
                 {
                     PasswordValidators.Add(v);
                 }
-            }
-            else
-            {
-                PasswordValidators.Add(new PasswordValidator<TUser>());
             }
 
             foreach (var providerName in Options.Tokens.ProviderMap.Keys)
--- a/src/Microsoft.AspNet.Identity/UserManager.cs
+++ b/src/Microsoft.AspNet.Identity/UserManager.cs
@@ -48,26 +48,23 @@ namespace Microsoft.AspNet.Identity
         /// <param name="contextAccessor">The accessor used to access the <see cref="HttpContext"/>.</param>
         public UserManager(IUserStore<TUser> store,
             IOptions<IdentityOptions> optionsAccessor,
-            IPasswordHasher<TUser> passwordHasher,
-            IEnumerable<IUserValidator<TUser>> userValidators,
-            IEnumerable<IPasswordValidator<TUser>> passwordValidators,
-            ILookupNormalizer keyNormalizer,
-            IdentityErrorDescriber errors,
             IServiceProvider services,
-            ILogger<UserManager<TUser>> logger,
-            IHttpContextAccessor contextAccessor)
+            ILogger<UserManager<TUser>> logger)
         {
             if (store == null)
             {
                 throw new ArgumentNullException(nameof(store));
             }
             Store = store;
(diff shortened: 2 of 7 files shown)


src/Microsoft.AspNet.Identity/UserManager.cs
```diff
@@ -86,6 +85,7 @@ public class UserManager<TUser> : IDisposable where TUser : class
 
             if (services != null)
```
davidfowl: This can't be null. Unit test?

HaoK: Yeah we pass in nulls in the tests

src/Microsoft.AspNet.Identity/UserManager.cs
```diff
@@ -48,27 +48,23 @@ public class UserManager<TUser> : IDisposable where TUser : class
...
-            Options = optionsAccessor?.Value ?? new IdentityOptions();
-            _context = contextAccessor?.HttpContext;
-            PasswordHasher = passwordHasher;
-            KeyNormalizer = keyNormalizer;
-            ErrorDescriber = errors;
             Logger = logger;
+            Options = optionsAccessor?.Value ?? new IdentityOptions();
 
+            _context = services?.GetService<IHttpContextAccessor>()?.HttpContext;
+            PasswordHasher = services?.GetService<IPasswordHasher<TUser>>() ?? new PasswordHasher<TUser>();
```
davidfowl: Ahhh, we don't really had a great way to detect optional services. Are these things really not required?

HaoK:
Updated, removing a bunch of default services from AddIdentity, yeah so 99% of the time noone is going to replace most of the services like LookupkeyNormalizer/passwordHasher.  They can still do it like before via adding it to the service collection, but the defaults are stateless and can be new'd up.  And the ctors are semi sane again with only 4 instead of 10 arguments!

src/Microsoft.AspNet.Identity/UserManager.cs
```diff
@@ -48,27 +48,23 @@ public class UserManager<TUser> : IDisposable where TUser : class
...
-            _context = contextAccessor?.HttpContext;
-            PasswordHasher = passwordHasher;
-            KeyNormalizer = keyNormalizer;
-            ErrorDescriber = errors;
             Logger = logger;
+            Options = optionsAccessor?.Value ?? new IdentityOptions();
+
+            _context = services?.GetService<IHttpContextAccessor>()?.HttpContext;
+            PasswordHasher = services?.GetService<IPasswordHasher<TUser>>() ?? new PasswordHasher<TUser>();
+            KeyNormalizer = services?.GetService<ILookupNormalizer>() ?? new UpperInvariantLookupNormalizer(); ;
```
divega: @HaoK Is this the third time already that you try to move us to use service locator and to take our default services out of DI? :smile:

It still looks bizarre to me and I would prefer us to have a wider discussion about either adopting this pattern (or declaring it an anti-pattern) across all of ASP.NET.

HaoK: Well, if we are doing service locator for HttpContextAccessor, the other ones are good candiates too :)

HaoK: Plus the main motivation has always been to make it easy to new up a UserManager that just works regardless of the state of the ServiceCollection.  The only thing it really needs is the Store, all of the other stuff is really optional...

halter73: I prefer to be explicit when it comes to dependencies. Now most the param doc-commments don't make sense. I wouldn't remove the doc-comments, because I think it's useful to explain which services can be replaced to change the behavior of the user manager. However, if we move to the service locator pattern, we don't have a standard way (i.e. `<param...`) to describe the dependencies and how they are used.

I also don't like hiding the default implementation of services. I guess it's not *really* hidden since the default implementation here matches those added by `AddIdentity`, but that just makes this less DRY. 

Newing up the default service implementation also isn't a good solution in the generaly case, even though it doesn't seem like it's an issue here, because things quickly get out of hand if there are transitive dependencies.

For services that really are optional, as opposed to having a known default implementation, I am more sympathetic to using the service locator pattern. Even then, there is an alternative, and that is to use `IEnumerable<TService>`. The only real downside I see to this is that the type implies that 0-many services can be provided when we really expect one service or none at all. UserManager doesn't seem to have any of the truly optional services though.

src/Microsoft.AspNet.Identity/UserManager.cs
```diff
@@ -48,27 +48,23 @@ public class UserManager<TUser> : IDisposable where TUser : class
...
+            ErrorDescriber = services?.GetService<IdentityErrorDescriber>() ?? new IdentityErrorDescriber(); ;
 
+            var userValidators = services?.GetService<IEnumerable<IUserValidator<TUser>>>();
             if (userValidators != null)
```
halter73: In the old version of the code I don't think this null check was necessary. Our IoC containers are required to pass an empty IEnumerable to the constructor when no matching services are registered.

src/Microsoft.AspNet.Identity/UserManager.cs
```diff
@@ -48,27 +48,23 @@ public class UserManager<TUser> : IDisposable where TUser : class
...
             IOptions<IdentityOptions> optionsAccessor,
-            IPasswordHasher<TUser> passwordHasher,
-            IEnumerable<IUserValidator<TUser>> userValidators,
-            IEnumerable<IPasswordValidator<TUser>> passwordValidators,
-            ILookupNormalizer keyNormalizer,
-            IdentityErrorDescriber errors,
             IServiceProvider services,
-            ILogger<UserManager<TUser>> logger,
-            IHttpContextAccessor contextAccessor)
+            ILogger<UserManager<TUser>> logger)
```
halter73: So now we're a mishmash of constructor injection and the service locator pattern? I understand it was necessary before for ProviderTypes, but this seems almost arbitrary.

src/Microsoft.AspNet.Identity/UserManager.cs
HaoK: I think it really boils down to how much do we care about being able to 'sanely' new up a UserManager.  This used to be something that was quite easy to do, not so much these days.  If we only really support getting a UserManager via services.GetService<UserManager<MyUser>>() then the ctor taking 10 arguments is less of an issue...

src/Microsoft.AspNet.Identity/SignInManager.cs
```diff
@@ -41,7 +42,7 @@ public class SignInManager<TUser> where TUser : class
             {
                 throw new ArgumentNullException(nameof(userManager));
             }
-            if (contextAccessor == null || contextAccessor.HttpContext == null)
```
halter73: Won't most of the methods in this class now NullRef in `contextAccessor.HttpContext` is null? I think it is better to throw in the ctor? Maybe it shouldn't `ArgumentNullException` but instead an `InvalidArgumentExeption` with a clear explanation that the `HttpContext` property can't be null.

HaoK: Yeah I think we just need to defer the Context guard into a helper method which will throw then if HttpContext is null, https://github.com/aspnet/Identity/issues/674 basically asked us not to get it in the ctor.

src/Microsoft.AspNet.Identity/UserManager.cs
halter73: Considering that you would need valid `IUserStore<TUser>` and `ILogger<UserManager<TUser>>` instances to new up a `UserManager`  at the very least, I don't think requiring a user to call `services.GetService<UserManager<MyUser>>()` (or getting it via ctor injection in there controller for example) is to onerous. Users would need to do the same thing for any stateful service anyway.

src/Microsoft.AspNet.Identity/SignInManager.cs
halter73: Maybe we can null check in the getter then so the user gets a clearer error message if the context it is null when it's actually needed. I guess whether or not it's worth it depends on how possible it is for this issue to come up.

src/Microsoft.AspNet.Identity/UserManager.cs
HaoK: The options and logger actually could be grouped with the other optional services, and service located/new'd up if not found.  

The only thing that really is required is the IUserStore

src/Microsoft.AspNet.Identity/UserManager.cs
HaoK: The guards were mostly unit tests/new passing in null

src/Microsoft.AspNet.Identity/UserManager.cs
divega: :+1: to everything @halter73 said. 

Besides the lack of clarity on the actual dependencies and ambiguous optionality in the constructor, from the code maintainability perspective this adds a new alternate source of truth about our dependencies. Rather than having it all encoded in `AddIdentity()` & friends, it is fragmented between that and this constructor.

Ultimately I agree that this is about how much we care about making it easy to create a UserManager without using DI, and I personally don't think it is a goal.

HaoK: If we do care about new() support, This is what I'd propose for the ctor:

```UserManager(IUserStore<TUserStore> store, IServiceProvider services)```

Where store is required and must not be null, and we'd look up everything else from the services, and use defaults if needed.  AddIdentity would actually no longer need to register any of these worker services.  It would only register the top level mangers (User/Role/SignIn)

MaximRouiller:
Just want to add one thing about Service Locator Pattern.

http://blog.ploeh.dk/2010/02/03/ServiceLocatorisanAnti-Pattern/
http://blog.ploeh.dk/2014/05/15/service-locator-violates-solid/
http://blog.ploeh.dk/2015/10/26/service-locator-violates-encapsulation/

And I agree. We don't know what depends on what, we need to examine the source... if a dependency has to be pre-filled, I'd rather have it done through properties than through the Service Locator Pattern.

I don't know how easier it would be but... classes where the Service Locator Pattern are a runtime time bomb. Even in tests... unless you start going into the source code... it gets bad.

HaoK:
Replaced by https://github.com/aspnet/Identity/pull/687

PR #664, opened 2015-12-06
Fix check of default value in ConvertIdToString if TKey is a class

fredgate:
In the UserStore class of Microsoft.AspNet.Identity.EntityFramework project, the function ConvertIdToString checks the value of key passed as an argument to verify if it is "null". The type of the argument is the generic type TKey. If the type is a struct it is ok, but if the type is a class (like a String), the test will cause an NullException.


HaoK:
Thanks! 042588e1b231e2bba0bf1748c1e0ced3e6c32a24

Issue #663, opened 2015-12-06
GetUserId extension method doesn't honor IdentityOptions.ClaimsIdentity.UserIdClaimType


brockallen:
Same for GetUserName. They're both hard coded to use the legacy SOAP/WS-* `ClaimTypes` enum.

rustd:
Add this to the announcements repo.

brockallen:
I'm guessing you're not looking for a fix, so might I suggest adding these as extension methods to the VS template project and not into the core identity framework? In many projects the display name should not be the same as the username the user enters to login.

leastprivilege:
really? I always say "good morning MICROSOFT/PRANAV" when I meet him him the hallway.

MaximRouiller:
I'll have to greet him that way at the next Summit. ;-)

HaoK:
We are actually going to nuke this extension since it was only important in 2.0 because all the user manager APIs took user id, now that they take the user itself, its no longer needed

brockallen:
yay. the less people are forced to use certain claims types, the better. thx.

HaoK:
Gone 1c26e8f5fa1b36febaca1f9a63b2a2f65fb11727

HaoK:
Looks like we need to provide alternate sugar for the templates in some from before removing these

HaoK:
Going to remove all the identity claims principal extensions and add instance methods on SignInManager:

GetUserId/UserName/IsSignedIn...

Templates will need to be updated

HaoK:
2dade94404d3198b19e6af9083b410287e11314f

Issue #630, opened 2015-11-11
Revisit User.IsSignedIn()


rustd:

rustd:
Delete this and in the templates add User.IsAuthenticated()

This is a breaking change so please post it in the Announcements repo.

HaoK:
Tracking via https://github.com/aspnet/Identity/issues/663

xperiandri:
What about User.GetUserId() ?

HaoK:
That got moved to UserManager as well.

PR #671, opened 2015-12-09
Nuke unneeded ClaimsPrincipal extensions

HaoK:
Also fix a bug with custom user id claim type with SecurityStampValidator

cc @divega

### DIFF
--- a/src/Microsoft.AspNet.Identity/PrincipalExtensions.cs
+++ b/src/Microsoft.AspNet.Identity/PrincipalExtensions.cs
@@ -12,36 +12,6 @@ namespace System.Security.Claims
     public static class PrincipalExtensions
     {
         /// <summary>
-        /// Returns the Name claim value if present otherwise returns null.
-        /// </summary>
-        /// <param name="principal">The <see cref="ClaimsPrincipal"/> instance this method extends.</param>
-        /// <returns>The Name claim value, or null if the claim is not present.</returns>
-        /// <remarks>The Name claim is identified by <see cref="ClaimsIdentity.DefaultNameClaimType"/>.</remarks>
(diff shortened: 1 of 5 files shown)


HaoK:
Addresses: https://github.com/aspnet/Identity/issues/663

src/Microsoft.AspNet.Identity/PrincipalExtensions.cs
```diff
@@ -12,36 +12,6 @@ namespace System.Security.Claims
...
-            {
-                throw new ArgumentNullException(nameof(principal));
-            }
-            return principal.FindFirstValue(ClaimTypes.NameIdentifier);
-        }
-
-        /// <summary>
         /// Returns true if the principal has an identity with the application cookie identity
         /// </summary>
         /// <param name="principal">The <see cref="ClaimsPrincipal"/> instance this method extends.</param>
```
divega: While you are here, could you fix the null-conditional operator usage in the method bellow? E.g. principal is never null after the check. Or is this the method that is going away per #630?

divega:
:shipit:

src/Microsoft.AspNet.Identity/PrincipalExtensions.cs
HaoK: Sure, done

HaoK:
1c26e8f5fa1b36febaca1f9a63b2a2f65fb11727

HaoK:
Premature removal, looks like we need to fix templates before we nuke these, as they still rely on this broken sugar method.

HaoK:
Replaced by https://github.com/aspnet/Identity/pull/701

Issue #670, opened 2015-12-09
XPlat tests: EntityFramework Store Tests need replacement for localdb


HaoK:
Currently Identity.EntityFramework provider tests are skipped for SqlServer since there's no LocalDb on xplat.

divega:
The basic plan here could be to refactor the tests to take the connection string from configuration and then point to a network instance of SQL Server. We have taken this approach for EF tests.

divega:
Another possibility is to start testing with a different database provider, e.g. PostgreSQL but that seems a bit more far out.

HaoK:
Try removing the OS skips to run tests xplat

HaoK:
Doesn't work, as would still need to switch off of localdb and configure an external SQL server.  I think we have enough coverage xplat with the custom InMemory, and EF InMemory stores running on non windows. @divega ok closing this as won't fix?  The only small gap without these tests is there if was any delta between EF tests and the specific identity store queries

HaoK:
Moving to backlog not high pri

PR #672, opened 2015-12-09
Turn on xplat tests

HaoK:


HaoK:
@divega This turns on the xplat test runs, unfortunately the current shared code has the attributes, so even though these should pass for the in memory providers, they are picking up the Skips still.  Filed an issue for this separate from the EF specific local db one...

https://github.com/aspnet/Identity/issues/673

HaoK:
Updated removing the attributes and using a manual skip method check which does the same checks, this allows the non SQL providers to run on xplat now.

test/Microsoft.AspNet.Identity.EntityFramework.Test/CustomPocoTest.cs
```diff
@@ -46,7 +46,7 @@ public class CustomDbContext<TKey> : DbContext where TKey : IEquatable<TKey>
 
         [TestPriority(-1000)]
         [ConditionalFact]
-        [FrameworkSkipCondition(RuntimeFrameworks.Mono)]
+        [FrameworkSkipCondition(RuntimeFrameworks.Mono)][OSSkipCondition(OperatingSystems.Linux)][OSSkipCondition(OperatingSystems.MacOSX)]
```
HaoK: These are one off, standalone non shared tests, so they can continue to use the skip attributes

divega: Do you actually mean we will never want to run them on Linux and Mac? Or is it just a temporary limitation?

divega: Nit: split the lines to have just one attribute per line.

HaoK: Basically until https://github.com/aspnet/Identity/issues/670 is fixed

divega:
:shipit: but the skip conditions in those "one off, standalone, non-shared" tests smell like regular engineering debt :smile: If that is the case, please make sure we have a bug tracking it.

test/Microsoft.AspNet.Identity.EntityFramework.Test/CustomPocoTest.cs
HaoK: Or we forever if decide this one test project can be windows only :)

divega: > Or forever if we decide this one test project can be windows only :)

Not sure why we would decide such thing, but we can talk :smile:

HaoK:
a66f50c1ba2dec01b0ca001dc22767d77ece29f2

PR #687, opened 2015-12-19
IHttpContextAccessor changes

HaoK:
- Optional for UserManager
- Defer checking for null until getter for SignInManager

cc @divega


HaoK:
167bb54c8d7aaafa12983164f8033cc25d0af39e

PR #701, opened 2016-01-04
Remove ClaimsPrincipal extensions, replace with instance

HaoK:
- Removes static GetUserId/GetUserName/IsSigned in from ClaimsPrincipal
- Replaced with SignInManager.GetUserId/GetUserName/IsSignedIn/GetUserAsync which accomplish the same thing but eliminates statics, and instead respect IdentityOptions.ClaimsIdentity for the claim types
- Cleaned up SecurityStamp validator to use SignInManager/IOptions as a service

cc @divega @blowdart @rustd


HaoK:
Fixes https://github.com/aspnet/Identity/issues/663

rustd:
@HaoK Can you please make the changes in the templates repo for these new APIs as well?

rustd:
Can you post to the announcements repo as well about this change.

HaoK:
Move GetUserId/GetUserName to UserManager

HaoK:
2dade94404d3198b19e6af9083b410287e11314f

PR #703, opened 2016-01-06
Updating to new options pattern

JunTaoLuo:
https://github.com/aspnet/Options/issues/105

### DIFF
--- a/src/Microsoft.AspNet.Identity/IdentityOptions.cs
+++ b/src/Microsoft.AspNet.Identity/IdentityOptions.cs
@@ -2,8 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using Microsoft.AspNet.Identity;
 
-namespace Microsoft.AspNet.Identity
+namespace Microsoft.AspNet.Builder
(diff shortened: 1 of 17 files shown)


src/Microsoft.AspNet.Identity/IdentityOptions.cs
```diff
@@ -2,8 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using Microsoft.AspNet.Identity;
 
-namespace Microsoft.AspNet.Identity
+namespace Microsoft.AspNet.Builder
```
HaoK: We are moving all options classes into the builder namespace?

HaoK:
:shipit:

src/Microsoft.AspNet.Identity/IdentityOptions.cs
JunTaoLuo: Yes since it's inconvenient otherwise when users have `app.UseMiddleware(new MiddlewareOptions());`. Moving options to the same namespace as `UseMiddleware` removes the need to add another using statement.

Issue #704, opened 2016-01-07
User Manager Add FindByPhoneNumber


prog-moh:
some times phone number used as identity of user it will be good enhancement to add method to 
find user by Phone Number

HaoK:
We decided no to add this functionality to the framework

PR #715, opened 2016-01-16
Added FindByPhoneNumberAsync 

SharePointRadi:
This PR addresses the FindByPhoneNumberAsync functionality request in UserStore and UserManager, and the IPhoneNumberStore interface
Related issue https://github.com/aspnet/Identity/issues/704

I am still trying to get xunit to run normally to commit the unit tests as well...

### DIFF
--- a/src/Microsoft.AspNet.Identity.EntityFramework/UserStore.cs
+++ b/src/Microsoft.AspNet.Identity.EntityFramework/UserStore.cs
@@ -1063,6 +1063,23 @@ namespace Microsoft.AspNet.Identity.EntityFramework
         }
 
         /// <summary>
+        /// Gets the user, if any, associated with the specified, phone number.
+        /// </summary>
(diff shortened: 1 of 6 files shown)


SharePointRadi:
I added a unit test to validate that we can find users by phone number.

divega:
@rustd, @HaoK, @blowdart I cannot recall what we decided in triage about this PR. 

In any case, if we wanted to take it I am not sure about the name of the method. Across a few components we are pretty consistently using the `Find` prefix for methods that return zero or one instances based on a condition or key. This is more of a `GetUsersByPhoneNumber()` I think.

SharePointRadi:
I am happy to follow through any decisions you guys make and amend it

HaoK:
We decided that we don't think this is appropriate as general functionality since apps can do this easily themselves.

PR #737, opened 2016-02-04
Set the server manually

hishamco:


divega:
@hishamco can you point us to some data on why this is a good change to take? Sorry I personally don't have enough context on it.

hishamco:
This is another way to use the kestrel server in code instead of having a configuration file, i did it in couple of repos Localization, Session to let the people know there's an alternative way to achieve that

Eilon:
That's actually not the reason to change it, but we should most likely change it (the current code definitely won't work as-is).

Please https://github.com/aspnet/Entropy/pull/93#issuecomment-180047517.

hishamco:
Thanks @Elion

JunTaoLuo:
merged in https://github.com/aspnet/Identity/commit/6cc6f4d88e579e2ad5b96cfe477297a168e38ebc

PR #748, opened 2016-02-18
Initial token schema changes and API

HaoK:
cc @divega @blowdart @Tratcher @rustd 

Rough initial shape of the new Authentication Token schema and APIs, we probably need to tweak the names a bit more, renamed the old Token apis a bit to consistently have TwoFactorToken.  We need to disambiguate these a bit...

Added a test for persistence/get/set/remove/update.  Still need to write one functional test mocking the end to end flow of the controller taking tokens from the external cookie and storing them into identity.


Tratcher:
Adding samples would be nice. I'm not familiar enough with Identity to follow this directly.

PinpointTownes:
I like it :clap: 

Just one question: have you considered adding an optional expiration property to `AuthenticationToken` (in aspnet/Security) and `IdentityUserToken`? Personally, I'm not a huge fan of storing `expires_at` as an authentication token.

HaoK:
Is it true that expires_at is only really ever used for the access token/bearer token and not relevant to any of the other tokens?  I'm just trying to understand the usage pattern so I can have an informed opinion :)

Tratcher:
https://tools.ietf.org/html/rfc6749#section-4.2.2
   expires_in
         RECOMMENDED.  The lifetime in seconds of the access token.  For
         example, the value "3600" denotes that the access token will
         expire in one hour from the time the response was generated.
         If omitted, the authorization server SHOULD provide the
         expiration time via other means or document the default value.

Tratcher:
Though storing an AuthToken becomes more difficult if it's more than just a Key/Value pair.

HaoK:
Updated the sample and added a test for the flow, its pretty much one line in the normal external login and external link action to a new method I added which takes care of updating the tokens from the external cookie.  Doesn't look too bad....

HaoK:
@blowdart @Tratcher @PinpointTownes @divega 

Everyone happy with the basic design now for this and https://github.com/aspnet/Security/pull/698?  If so I'll start polishing the PRs for merging this week...

PinpointTownes:
@HaoK yeah, I really like this new approach (specially the fact security middleware don't deal with token storage).

src/Microsoft.AspNetCore.Identity.EntityFrameworkCore/IdentityUserToken.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        /// <summary>
+        /// Gets or sets the LoginProvider this token is from.
+        /// </summary>
+        public virtual string LoginProvider { get; set; }
+
+        /// <summary>
+        /// Gets or sets the name of the token.
+        /// </summary>
+        public virtual string TokenName { get; set; }
```
blowdart: Remove Token here and in TokenValue

src/Microsoft.AspNetCore.Identity.EntityFrameworkCore/UserStore.cs
```diff
@@ -1196,5 +1197,69 @@ where userrole.RoleId.Equals(role.Id)
...
         }
+
+        private Task<IdentityUserToken<TKey>> FindToken(TUser user, string loginProvider, string name, CancellationToken cancellationToken)
+        {
+            var userId = user.Id;
+            return UserTokens.SingleOrDefaultAsync(l => l.UserId.Equals(userId) && l.LoginProvider == loginProvider && l.TokenName == name, cancellationToken);
+        }
+
+        // <inheritdoc>
+        public virtual async Task SetTokenAsync(TUser user, string loginProvider, string name, string value, CancellationToken cancellationToken)
```
blowdart: SetAuthenticationToken() and elsewhere

src/Microsoft.AspNetCore.Identity/ExternalLoginInfo.cs
```diff
@@ -28,5 +30,7 @@ public class ExternalLoginInfo : UserLoginInfo
         /// </summary>
         /// <value>The <see cref="ClaimsPrincipal"/> associated with this login.</value>
         public ClaimsPrincipal ExternalPrincipal { get; set; }
```
blowdart: ExternalPrincipal => Principal

src/Microsoft.AspNetCore.Identity/ExternalLoginInfo.cs
```diff
@@ -28,5 +30,7 @@ public class ExternalLoginInfo : UserLoginInfo
...
+
+        public IEnumerable<AuthenticationToken> Tokens { get; set; }
```
blowdart: AuthenticationTokens

src/Microsoft.AspNetCore.Identity/UserManager.cs
```diff
@@ -1954,6 +1969,93 @@ public virtual Task<IList<TUser>> GetUsersInRoleAsync(string roleName)
         }
 
         /// <summary>
+        /// Returns an authentication token for a user.
+        /// </summary>
+        /// <param name="user"></param>
+        /// <param name="loginProvider">The authentication scheme for the provider the token is associated with.</param>
+        /// <param name="tokenName">The name of the token.</param>
+        /// <returns></returns>
+        public virtual Task<string> GetAuthenticationTokenAsync(TUser user, string loginProvider, string tokenName)
```
blowdart: tokenName => Name

src/Microsoft.AspNetCore.Identity.EntityFrameworkCore/UserStore.cs
PinpointTownes: I'm not sure it makes things clearer as...
  - the risk is high to mix them up with 2-FA tokens, that are also "authentication tokens".
  - they are more authorization tokens you can use to query an API or to refresh a token than real authentication tokens.

src/Microsoft.AspNetCore.Identity/UserManager.cs
HaoK: keeping tokenName for these since we do the same for the other similar APIs, userName, roleName as opposed to only saying just name

src/Microsoft.AspNetCore.Identity/UserManager.cs
```diff
@@ -1954,6 +1969,93 @@ public virtual Task<IList<TUser>> GetUsersInRoleAsync(string roleName)
...
+            if (loginProvider == null)
+            {
+                throw new ArgumentNullException(nameof(loginProvider));
+            }
+            if (tokenName == null)
+            {
+                throw new ArgumentNullException(nameof(tokenName));
+            }
+
+            // REVIEW: should updating any tokens affect the security stamp?
```
HaoK: @blowdart do we want to roll the security stamp if an auth token is updated?

blowdart: Don't think so If we get a new facebook token that's not a reason to invalidate a cookie login elsewhere is it?

src/Microsoft.AspNetCore.Identity.EntityFrameworkCore/UserStore.cs
HaoK: Have another name in mind @PinpointTownes 

We are calling it AuthenticationTokenStore currently, so Get/SetAuthToken matches

src/Microsoft.AspNetCore.Identity/UserManager.cs
HaoK: That is what I'm asking you :)  I'm not sure, but I was leaning against as evidenced by the current state of things...

HaoK:
82863e94447fbe9b1724d3271535c3d2442015d9

Issue #742, opened 2016-02-10
Change username's default allowed characters to confirm better to email spec


janpieterz:
Since the default configuration in the samples, and presumably in quite a lot of systems will be that the username is the email address, it threw me off a little bit when discovering that the + sign is not an allowed character in a username.

I was thinking that it would be helpful to offer the email's allowed list (see [this ](http://stackoverflow.com/a/2049510/1202118)answer of someone diving into the allowed characters).

If this is not feasible (I can imagine that a username should be something remotely readable, although one could argue that an email should be as well), would it be possible to offer both versions of these strings as a const string in the user options class, so that this behavior is not repeatedly implemented by every application? In either case, the current implementation suggests it allows email addresses (the @ and .) but it doesn't actually fully implement this.

The string I use in my application (as far as I know I'm only missing the 'quoted' character requirement):
```
abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789!#$%&'*+-/=?^_`{|}~.@
```

HaoK:
@blowdart the delta appears to be the following characters, how many of these do you want to enable by default, some of these look pretty wonky to be valid user name characters by default...

!#$%&'*+/=?^`{|}~

blowdart:
Oh that is bad isn't it. Lots of invalid characters for emails. This is strange. Either we update the samples so it's not emails, or we pare it down a bit.

MaximRouiller:
> I can imagine that a username should be something remotely readable, although one could argue that an email should be as well

@janpieterz Well... [the original RFC compliant regex](http://ex-parrot.com/~pdw/Mail-RFC822-Address.html) doesn't argue. That is if you want to validate as per the RFC.

There is many characters that are available in an email (`()[]+-<>` among others).

[Good read](http://girders.org/blog/2013/01/31/dont-rfc-validate-email-addresses/) if you want to go further. 

One thing for sure... refusing a valid email is plainly frustrating to most users.

MaximRouiller:
I do use the `+` sign in my Gmail address. Allows me to tag emails I use to register to some services.

blowdart:
We won't be using regular expressions for validation. We made a conscious decision to remove that approach.

GuardRex:
For reference, `AllowedUserNameCharacters` is here ...

[...]

PR #755, opened 2016-02-23
Add + to allowed username characters by default

HaoK:
Fixes https://github.com/aspnet/Identity/issues/742

cc @divega @blowdart

### DIFF
--- a/src/Microsoft.AspNetCore.Identity/UserOptions.cs
+++ b/src/Microsoft.AspNetCore.Identity/UserOptions.cs
@@ -14,7 +14,7 @@ namespace Microsoft.AspNetCore.Identity
         /// <value>
         /// The list of allowed characters in the username used to validate user names.
         /// </value>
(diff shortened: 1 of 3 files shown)


divega:
:shipit: I am fine with taking this change because it seems to be an improvement. It would be great if we had more of an objective way to decide what characters to allow by default, e.g. something based on what the most popular email servers/services out there allow.

MaximRouiller:
@divega 

**Gmail**: letter, numbers, `.`, `+`

Don't know about other email providers.

divega:
Gmail actually allows letters, numbers and periods but not plus sign. Of course other systems do allow the later. Anyway, the change is in and I don't think we need to proactively revisit the list until we get a similar request,  if we ever do.

MaximRouiller:
Gmail does not allow + when registering but any emails on Gmail containing the + sign will be valid.

They call it email "tagging" or something like that.

PR #756, opened 2016-02-25
Make ServiceCollectionExtensions consistent

javiercn:
* Add an IIdentityBuilder interface.
* Move instance methods on IdentityBuilder into IIdentityBuilderExtensions

### DIFF
--- a/src/Microsoft.AspNetCore.Identity/IIdentityBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Identity/IIdentityBuilderExtensions.cs
@@ -1,0 +1,237 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNetCore.Identity
+{
+    /// <summary>
+    /// Helper functions for configuring identity services.
+    /// </summary>
+    public static class IIdentityBuilderExtensions
+    {
+        private static IIdentityBuilder AddScoped(IIdentityBuilder builder, Type serviceType, Type concreteType)
+        {
+            builder.Services.AddScoped(serviceType, concreteType);
+            return builder;
+        }
+
+        /// <summary>
+        /// Adds an <see cref="IUserValidator{TUser}"/> for the <seealso cref="IIdentityBuilder.UserType"/>.
+        /// </summary>
+        /// <typeparam name="T">The user type to validate.</typeparam>
+        /// <param name="builder">The <see cref="IIdentityBuilder"/>.</param>
+        /// <returns>The current <see cref="IIdentityBuilder"/> instance.</returns>
+        public static IIdentityBuilder AddUserValidator<T>(this IIdentityBuilder builder) where T : class
+        {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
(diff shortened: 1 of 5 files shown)


src/Microsoft.AspNetCore.Identity/IIdentityBuilderExtensions.cs
```diff
@@ -0,0 +1,237 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNetCore.Identity
+{
+    /// <summary>
+    /// Helper functions for configuring identity services.
+    /// </summary>
+    public static class IIdentityBuilderExtensions
```
divega: Naming: Don't include the "I" for interfaces in the name of the extension method sponsor class, i.e. this should be just `IdentityBuilderExtensions`.

divega:
@javiercn I think this change is correct but its value isn't obvious. E.g. the new interface you are introducing seems a bit superfluous. I get that what we had looks more like what EF uses and with your change it looks more like what MVC has, but the user experience should be the same. 

Could you elaborate a bit more on why this is better than what we had? Maybe @rynowak has an insight?

cc @HaoK @blowdart

HaoK:
Yeah I don't see the value at all of the interface for the builder. It's not replaceable or put in DI so there is no situation where we would use it is there?

javiercn:
Things to pay attention to:
* Comments (There's been a lot of copy paste, so mistakes can happen).
* Method names
* Null checks for arguments
* Overload with configuration action calls parameterless overload.
* Parameterless overload does the work and brings in direct depenendencies
  (Like routing on MVC, for most repositories it means making sure AddOptions is called).

HaoK:
I don't think its worth drilling deeper into this PR until the fundamental question about we need an IIdentityBuilder interface is first resolved...

javiercn:
@divega @HaoK A few points.
* We only care about configuring the services here, so the class is just an implementation detail.
* This doesn't affect the way the user consumes the API. For them, this is just a handle over which to define/call extension methods most of the time. (See EntityFrameworkIdentityServiceCollectionExtensions).
* There is no value in having a specific type either, as there is nothing to encapsulate (all behavior needs to be public so that we can guarantee that any user can write any extension method they want against this in the same we can (the interface enforces this)).
* Using an interface makes it consistent with the way we do things in other parts of the stack (DataProtection, MVC, EF)

From my point of view, there's not a big difference in value from having a class vs having an interface in this case, but I think that the fact that this makes it consistent across all our stack tips the balance

HaoK:
If there's no value either way and its an implementation detail, I don't think we should be adding a new interface, it just pollutes things by adding an extra public type which is basically a duplicate of IdentityBuilder just to hang extension methods.  

I know MVC does a lot more stuff with various projects layering on extensions, so maybe its worth it there, but I don't think this extra layer is worth it for Identity or DataProtection, or anywhere else that currently doesn't have an interface for their builders...

divega:
Hey @javiercn I am not sure I 100% understand all your points, so feel free to push back if you think I am missing something important. 

I think things are strictly simpler the way they are right now, because there are less types :smile: 

From the consistency aspect I think we are probably not seeing the same stuff in EF. `AddEntityFramework()` returns a class, `EntityFrameworkServicesBuilder` which happens to implement an EF-specific interface that enables getting the underlying service collection, but the class has instance methods on it (currently only `AddDbContext<T>()`). EF Providers can define extension methods on `EntityFrameworkServicesBuilder` as well and have access to its state since everythign is public.

rynowak:
Is this whole discussion about the interface vs class? I don't think it's especially important. As @divega pointed out - the user experience is the same.

IMO the interface + extension methods the most flexibility since it could be implemented by *anything* - whereas the class + concrete methods is simpler as long as everything is in the same package.

divega:
@javiercn sorry if I didn't fully understand initially what you were proposing. We don't see enough advantages in this so closing.

javiercn:
@Eilon Are you ok with this resolution?

PR #757, opened 2016-02-27
Remove usage of IOptions

HaoK:
Looks like I can do this in reverse, removing usage from all places, before removing the feature since we aren't using the new sugar...

For identity, with this change we could consider making the main IdentityOptions actually consume the sub options as services instead of new'ing up instances of each of the sub options like Password/UserOptions.  Something to consider now that our options are services...

@divega @blowdart

### DIFF
--- a/src/Microsoft.AspNetCore.Identity/IdentityOptions.cs
+++ b/src/Microsoft.AspNetCore.Identity/IdentityOptions.cs
@@ -2,7 +2,9 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Generic;
 using Microsoft.AspNetCore.Identity;
+using Microsoft.Extensions.Options;
 
 namespace Microsoft.AspNetCore.Builder
 {
@@ -11,6 +13,18 @@ namespace Microsoft.AspNetCore.Builder
     /// </summary>
     public class IdentityOptions
     {
+        public IdentityOptions(IEnumerable<IConfigureOptions<IdentityOptions>> configureOptions = null)
+        {
+            if (configureOptions != null)
+            {
+                foreach (var configure in configureOptions)
+                {
+                    configure.Configure(this);
+                }
+            }
+        }
+
+
         /// <summary>
         /// Gets or sets the <see cref="ClaimsIdentityOptions"/> for the identity system.
         /// </summary>
--- a/src/Microsoft.AspNetCore.Identity/BuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Identity/BuilderExtensions.cs
@@ -32,7 +32,7 @@ namespace Microsoft.AspNetCore.Builder
                 throw new InvalidOperationException(Resources.MustCallAddIdentity);
             }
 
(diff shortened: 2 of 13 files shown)


src/Microsoft.AspNetCore.Identity/IdentityOptions.cs
```diff
@@ -11,6 +13,18 @@ namespace Microsoft.AspNetCore.Builder
     /// </summary>
     public class IdentityOptions
     {
+        public IdentityOptions(IEnumerable<IConfigureOptions<IdentityOptions>> configureOptions = null)
```
PinpointTownes: Holy s... :smile:

HaoK: Trust me I don't like this at all...  But its a 'taste' thing...  There will be a more sane looking options pattern where you call

`services.AddOptions<YourOption>()` which registers a singleton and does the mechanical gunk for you...  

But, hopefully everyone has the same reaction as you, and we will internally use the sugar :)

PinpointTownes: I hope so, because we had added `Options.Create` to avoid having to implement `IOptions<>` to return `this` (which was ugly), and now we have this much more horrible pattern... :smile:

HaoK: Yeah well the POR is https://github.com/aspnet/Options/issues/122.  Basically implementors can choose between this pattern and the nice pattern.  @lodejard feels strongly that this is the proper way to do things for framework code, so alas no sugar for us...

GuardRex: @HaoK I realize an Annoucement on this is forthcoming, but can you commnent on what the delta will be for the current pattern ...
```c#
services.AddOptions();
services.Configure<AppOptions>(appOptions =>
{
    appOptions.Key = "value";
    ...
});
```

HaoK: It actually will be pretty much the same, you'll need to register your option explicitly is the only difference:

```
services.AddOptions<AppOptions>();
services.Configure<AppOptions>(appOptions =>
{
    appOptions.Key = "value";
    ...
});
```

You won't need to change your AppOptions if you use this approach.

If you want to avoid using AddOptions, you can do the alternative
```
// AppOptions must take the IEnumerable<IConfigureOptions<AppOptions>> 
// and apply them in the ctor if you choose this route
services.AddSingleton<AppOptions>();
services.Configure<AppOptions>(appOptions =>
{
    appOptions.Key = "value";
    ...
});
```

GuardRex: yeah ... sorry ... I pulled that question b/c this was the wrong place to pose the question, and I read back in the main issue that you showed ...
```c#
// Just registers the MyOptions singleton
services.AddOptions<MyOptions>()

// Register the options singleton and configure it
services.AddOptions<MyOptions>(options => options.Setting = "Whatever"); //
```
... so I'm good. thx.

When it gets to the point of an Annoucement, if there are two ways (i.e., "If you want to avoid using AddOptions" approach), then I hope you'll just add in the pros and cons of each approach.

Issue #763, opened 2016-03-07
Cookie names not set properly.


JunTaoLuo:
I overlooked the fact that setting the AuthenticationScheme values does not actually achieve what we want. By default, schemes are set to `.AspNetCore.{AuthenticationScheme}` and this leads to names like `.AspNetCore..AspNetCore.Identity.Application` since cookies middleware adds the `.AspNetCore.` prefix by default. We can remove the `.AspNetCore.` from the prefix of the authentication scheme.

Still investigating the why there are also cookies like `.AspNetCore.Microsoft.AspNetCore.Identity.ExternalLogin`.

**Edit**: Ignore the ExternalLogin cookie issue. That is a really old legacy name for the cookie and the test is only checking it for non-existence.

fyi @natemcmaster

natemcmaster:
>  By default, cookie names are set to .AspNetCore.{AuthenticationScheme}

Where is this in code?

JunTaoLuo:
@natemcmaster I mean the default schemes https://github.com/aspnet/Identity/blob/dev/src/Microsoft.AspNetCore.Identity/IdentityCookieOptions.cs#L16-L20

PR #764, opened 2016-03-07
Redress cookie name

JunTaoLuo:
#763 

As part of fixing MusicStore tests in https://github.com/aspnet/MusicStore/pull/601, @natemcmaster found that we are not setting the cookie names correctly for cookies used by Identity. The reason for this is in the referenced issue. 

cc @Tratcher @HaoK

### DIFF
--- a/src/Microsoft.AspNetCore.Identity/IdentityCookieOptions.cs
+++ b/src/Microsoft.AspNetCore.Identity/IdentityCookieOptions.cs
@@ -13,7 +13,7 @@ namespace Microsoft.AspNetCore.Identity
     /// </summary>
     public class IdentityCookieOptions
     {
-        private static readonly string CookiePrefix = ".AspNetCore.Identity";
+        private static readonly string CookiePrefix = "Identity";
         private static readonly string DefaultApplicationScheme = CookiePrefix + ".Application";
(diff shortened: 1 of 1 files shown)


HaoK:
What happened to the AspNetCore?

JunTaoLuo:
@HaoK As I mentioned in the issue, it is added by default by CookieAuthentication at https://github.com/aspnet/Security/blob/a5b288897d2981c241133b8d37aa22136b653e3a/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationMiddleware.cs#L35 which uses the prefix defined at https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.Cookies/CookieAuthenticationDefaults.cs#L22

natemcmaster:
So to be clear, the expected cookie name is `Identity.Application` or `Identity.External`? there is no "Microsoft" or "AspNetCore" anymore?

JunTaoLuo:
No, it is expected to be `.AspNetCore.Identity.Application` it's just that the `.AspNetCore` part is added by Cookies Middleware already.

natemcmaster:
Ok. LGTM

JunTaoLuo:
ping @HaoK

HaoK:
:shipit:

PR #769, opened 2016-03-10
Update IdentityDbContext constructors to reflect EF changes

ajcvickers:
EF now has a parameterless constructor that can only be used when a derived context has OnConfiguring and is hence protected, and a constructor that takes a DbContextOptions.

@divega


ajcvickers:
Checking this in to fix build break. @divega Please review when you can.

divega:
:shipit:

divega:
Do we need similar changes in MusicStore and Templates?

ajcvickers:
@divega Yep. I have had the changes for all repros ready to go since yesterday, but all the C.I. stuff was bropken. I will check in fixes and monitor C.I. I expect it to take some time...

PR #770, opened 2016-03-11
Target dnx451

bricelam:
Reverts 6bea51909239f5dd6c5b9f44b0575b9f7227a770 on DNX


divega:
:shipit:

bricelam:
Recalling. After chatting with @Eilon, we decided that adding the facades to user's `project.json` is an acceptable workaround.

PR #772, opened 2016-03-14
Reacting to EF changes

pranavkm:
cc @ajcvickers 

This gets the repo to compile, but there's a bunch of failing tests that I don't know how to fix:

```C#
 Microsoft.AspNetCore.Identity.EntityFrameworkCore.Test.UserStoreGuidTest.CanFindRoleByNameWithManager [FAIL]
      System.InvalidOperationException : No database provider has been configured for this DbContext. A provider can be configured by overriding the DbContext.OnConfiguring method or by using AddDbContext on the application service provider. If AddDbContext is used, then also ensure that your DbContext type accepts a DbContextOptions object in its constructor.
```

### DIFF
--- a/samples/IdentitySample.Mvc/project.json
+++ b/samples/IdentitySample.Mvc/project.json
@@ -35,10 +35,7 @@
     "web": "IdentitySample.Mvc"
   },
   "frameworks": {
-    "dnx451": {
-      "frameworkAssemblies": {
-        "System.Reflection": ""
-      }
+    "net451": {
     },
     "netstandardapp1.5": {
       "imports": [
--- a/test/Microsoft.AspNetCore.Identity.EntityFrameworkCore.InMemory.Test/RoleStoreTest.cs
+++ b/test/Microsoft.AspNetCore.Identity.EntityFrameworkCore.InMemory.Test/RoleStoreTest.cs
@@ -25,8 +25,8 @@ namespace Microsoft.AspNetCore.Identity.EntityFrameworkCore.InMemory.Test
         public async Task CanCreateRoleWithSingletonManager()
         {
             var services = TestIdentityFactory.CreateTestServices();
-            services.AddAddEntityFrameworkInMemoryDatabase();
-            services.AddTransient<InMemoryContext>();
+            services.AddEntityFrameworkInMemoryDatabase();
+            services.AddSingleton(new InMemoryContext(new DbContextOptionsBuilder().Options));
             services.AddTransient<IRoleStore<IdentityRole>, RoleStore<IdentityRole, InMemoryContext>>();
             services.AddSingleton<RoleManager<IdentityRole>>();
             var provider = services.BuildServiceProvider();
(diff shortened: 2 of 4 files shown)


ajcvickers:
@pranavkm Please talk to @divega or @rowanmiller for anything not related to app week. Thanks.

divega:
It seems this change is missing adding a call to `AddEntityFrameworkSqlServer()` somewhere, e.g. the `SqlStoreTestBase.TestDbContext` at https://github.com/aspnet/Identity/blob/dev/test/Microsoft.AspNetCore.Identity.EntityFrameworkCore.Test/SqlStoreTestBase.cs#L36. is supposed to be configured to work with SQL Server for everything to work.

I am not able to follow how this was happening before.

@haok can you shed some light on that?

HaoK:
I don't think we ever did that before.  We only put the identity store/managers into DI, and we were calling a factory method on each test class to `CreateContext()` which typically just did new DbContext and passed it in via the ctors to the store.

divega:
Yeah, @ajcvickers and I talked about this and we believe the test was passing before because the DbContext was being created with new instead of resolved from the container, so EF was creating its own separate container and adding the SQL Server provider to it automatically, e.g. the test was an example of the pit of failure described at https://github.com/aspnet/EntityFramework/issues/4668 :smile:

Now that the issue has been fixed both options (using the application's service container or letting EF create its own) are valid and it should be much more obvious in the code what is happening, so we need to choose what pattern we want to follow in Identity tests.

@ajcvickers might have a recommendation but ultimately @HaoK and I need to make the call. I supposed the simplest change is to continue having EF created with its own container, but I am not sure what is missing then. Does it have a call to `UseSqlServer()` on `OnConfiguring()`?

HaoK:
I'm not sure I follow your question, but all of the tests basically do this today for setup:  So each manager/context basically is creating its own service container and resolving a manager out, with the context as the shared state only.

```
            var context = CreateTestContext();
            var userMgr = CreateManager(context);
            var roleMgr = CreateRoleManager(context);
```

Create Context for the EF.SqlServer is calling into the DbUtil test code Nate copied over from EF I think:

```
   public TestDbContext CreateContext()
        {
            var db = DbUtil.Create<TestDbContext>(_fixture.ConnectionString);
            db.Database.EnsureCreated();
            return db;
        }
```

pranavkm:
Checking this in. We can follow up on the test failure.

samples/IdentitySample.Mvc/project.json
```diff
@@ -35,10 +35,7 @@
     "web": "IdentitySample.Mvc"
   },
   "frameworks": {
-    "dnx451": {
```
divega: @pranavkm beware this is an unintentional change.

divega: please ignore :smile:

pranavkm: Targeting `dnx451` produces issues with System.Runtime.

PR #775, opened 2016-03-15
Minor fix in description for LockoutEnabled property.

aliakseimaniuk:
According to the logic of method **IsLockedOutAsync** from **UserManager**, **LockoutEnabled** property shows that the user is not locked out or not, it indicates that the user could be locked out, so the description was changed to reflect the actual behavior.


HaoK:
Merged thanks! 6b6b21a5bf1ff831301bdd9559ec36d4eaeac6e2

PR #796, opened 2016-04-11
Add Unique constraint to username index

HaoK:
cc @divega


HaoK:
0bf7ff24185891ad41ea11b0572e79b35cb2b370

Issue #869, opened 2016-06-13
XML comment for IdentityUser.Id property is messed up


xperiandri:
[See](https://github.com/aspnet/Identity/blob/dev/src/Microsoft.AspNetCore.Identity.EntityFrameworkCore/IdentityUser.cs#L71)

MaximRouiller:
*Misaligned*.

blowdart:
@HaoK - missed :) Can you adjust?

xperiandri:
Also incorrect comment at https://github.com/aspnet/Identity/blob/dev/src/Microsoft.AspNetCore.Identity.EntityFrameworkCore/IdentityUserClaim.cs#L21

xperiandri:
https://github.com/aspnet/Identity/blob/dev/src/Microsoft.AspNetCore.Identity.EntityFrameworkCore/UserStore.cs#L54
`a user and a role`

HaoK:
7b3b3627957404fe9083268929dd5589f7a26d1b

xperiandri:
I'm just looking at the comments to make [OpenIddict](https://github.com/openiddict/openiddict-core/) comments having the same style.

xperiandri:
The style of description is inconsistent for
https://github.com/aspnet/Identity/blob/dev/src/Microsoft.AspNetCore.Identity.EntityFrameworkCore/IdentityDbContext.cs#L51-L53
and
https://github.com/aspnet/Identity/blob/dev/src/Microsoft.AspNetCore.Identity.EntityFrameworkCore/UserStore.cs#L105-L112

Which one will be agreed for entity type description?
1. The type of `entity` object**s**.
2. The type representing **a** `entity`.

And for primary key
1. The type of the primary key for `entity`**es**`1` and `entity`**es**`2`>.
2. The type of the primary key for **a** `entity`.

PR #870, opened 2016-06-13
Fix doc comments

HaoK:
@divega @blowdart @Eilon

### DIFF
--- a/src/Microsoft.AspNetCore.Identity.EntityFrameworkCore/IdentityUser.cs
+++ b/src/Microsoft.AspNetCore.Identity.EntityFrameworkCore/IdentityUser.cs
@@ -67,8 +67,8 @@ namespace Microsoft.AspNetCore.Identity.EntityFrameworkCore
         }
 
         /// <summary>
-        /// </summary>
         /// Gets or sets the primary key for this user.
+        /// </summary>
         public virtual TKey Id { get; set; }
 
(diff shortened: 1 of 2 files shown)


divega:
@HaoK thanks! Could you remove CS1591 warning from Microsoft.AspNetCore.Identity/project.json and Microsoft.AspNetCore.Identity.EntityFrameworkCore/project.json and see what happens?

HaoK:
Still missing some doc comments

divega:
Can't build it right now :sad: How much is missing?

HaoK:
Not a ton, maybe on the order of 6-10 methods, none of them are very important looking

divega:
I think there a couple of new comments in https://github.com/aspnet/Identity/issues/869. Maybe we should wait until @xperiandri finishes reviewing? :smile:

divega:
Other than that LGTM.

divega:
@HaoK is it just `IdentityCookieOptions` that is missing some comments?

Eilon:
Time has pretty much run out, so please get in whatever you can; the rest will have to wait.

HaoK:
Ok I will just check this in as is, we can fix the rest in 1.0.1

HaoK:
7b3b3627957404fe9083268929dd5589f7a26d1b

PR #893, opened 2016-07-04
fix inconsistent generic constraints

arespr:
Simple fix for inconsistent generic constraints that prevents from creating for example abstraction layer over UserStore and RoleStore


HaoK:
We unfortunately need to do this in a non breaking way, replaced with https://github.com/aspnet/Identity/pull/930.  Thanks for submitting this PR though!

Issue #646, opened 2015-11-22
Add UserLoggedIn hook to SignInManager


rhegner:
I'm writing my custom SignInManager which is derived from `Microsoft.AspNet.Identity.SignInManager`. Right now I'm looking for a way to perform some action whenever the user has successfully logged in, regardless of the method he used (password, external, with/without two factor). That action could for example be setting a `LastLogin` field in my `ApplicationUser`.

I could override `SignInAsync` but it looks like this method not only gets called when the user has explicitly logged in but also on other occasions.

Or I could override `PasswordSignInAsync`, `TwoFactorSignInAsync` and `ExternalLoginSignInAsync`. Here I have to look at the sources to find out which overload of `PasswordSignInAsync` is the right one to override (or I just override all of its overloads), but this is not so future proof, since these implementation details could change in future versions.

Long story short: I think it would be nice to have some hook in `SignInManager` which gets called whenever a user has logged in.

rustd:
Would #650 meet your requirements?

rhegner:
No because I also need to perform some project specific actions after login.

HaoK:
As discussed with @rustd last week

HaoK:
Unassigned since this is in backlog, reassign when we want to do this.

musabsudan:
adding that would be great, I noticed that it's a common pattern that leads people o custom implementations, some "hooks" to perform extra work would be fantastic.

Thanks

HaoK:
Given that the current virtual SignInAsync catches too many flows for you, I think the best approach would be for you to extend exactly which methods you want to do your additional post SignIn logic.

HaoK:
All the hooks should already be there as every method is already virtual

PR #913, opened 2016-07-18
Add new SignedIn hook for SignInManager

HaoK:
Fixes: https://github.com/aspnet/Identity/issues/646

Adds a new SignedInAsync method which is called at the end of SignInAsync. does nothing by default.

cc @blowdart @divega

### DIFF
--- a/src/Microsoft.AspNetCore.Identity/SignInManager.cs
+++ b/src/Microsoft.AspNetCore.Identity/SignInManager.cs
@@ -118,6 +118,16 @@ namespace Microsoft.AspNetCore.Identity
         }
 
         /// <summary>
+        /// Called after SignInAsync has completed.
+        /// </summary>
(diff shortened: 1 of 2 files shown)


HaoK:
@divega look reasonable?

PinpointTownes:
Out of curiosity, what's the difference between that and simply overriding `SignInAsync`?
All the managers already have a bunch of methods, so if you start adding new events for every important method, it will quickly become a nightmare :smile: 

```csharp
public override async Task SignInAsync(TUser user, AuthenticationProperties authenticationProperties, string authenticationMethod = null) {
    await base.SignInAsync(user, authenticationMethod);

    // Do whatever you need here.
}
```

HaoK:
Hrm yeah good point I think I misunderstood the original issue he wants this event to be only called on explicit signins rather than on all sign ins. I agree with you I think he can just override the methods he wants in this case.

PR #923, opened 2016-07-28
Fixed UserStore constructor to match with IdentityDbContext

furlanrapha:
Having an issue trying to create a DbContext that extends from `IdentityDbContext<UserProfile, CustomRole, Guid, CustomUserClaim, CustomUserRole, CustomUserLogin, CustomRoleClaim, CustomUserToken>` and using at `Startup.cs` the `services.AddSingleton<UserStore<UserProfile, CustomRole, CustomIdentityDbContext, Guid, CustomUserClaim, CustomUserRole, CustomUserLogin, CustomUserToken>>();`


HaoK:
There's a similar PR here https://github.com/aspnet/Identity/pull/893

Issue #852, opened 2016-06-01
Expose a SignInManager.CheckPassword API that does lockout (optionally)


brockallen:
I know there's already a confirm email on the UserManager, but imagine this scenario: 

I want my email verification page to require the user's password (in addition to the code) -- people enter wrong email addresses all the time and we need a way to protect against the wrong email recipient from using the code to pwn the account.

So before I can verify the code, I also need to verify the password. Now I can call CheckPassword on the UserManager, but that level in the API does not provide me brute force protection. So I need an API on the SignInManager that will do the brute force checks on the password without trying to force 2fa or check that the email is already confirmed (because we're in the middle of confirming the email). 

So the SignInManager could have an API that accepts the user, code, and password. 

This email verification is hard stuff.

HaoK:
Sounds like what you really want is just a way to just CheckPassword that doesn't do SignIn or TwoFactor on SignInManager, but optionally will still count against lockouts/login failures 

```
  Task<SignInResult> SignInManager.CheckPasswordAsync(user, password, lockoutOnFailure)
```

HaoK:
fe60304df54c43030d3eacfd9073e6a36990db05

PR #931, opened 2016-08-08
Add CheckPasswordSignIn method

HaoK:
Fixes https://github.com/aspnet/Identity/issues/852

cc @divega @blowdart @brockallen

### DIFF
--- a/src/Microsoft.AspNetCore.Identity/SignInManager.cs
+++ b/src/Microsoft.AspNetCore.Identity/SignInManager.cs
@@ -241,32 +241,10 @@ namespace Microsoft.AspNetCore.Identity
                 throw new ArgumentNullException(nameof(user));
             }
 
-            var error = await PreSignInCheck(user);
-            if (error != null)
-            {
-                return error;
-            }
-            if (await IsLockedOut(user))
-            {
-                return await LockedOut(user);
-            }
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Identity/SignInManager.cs
```diff
@@ -241,19 +241,64 @@ public virtual Task SignInAsync(TUser user, bool isPersistent, string authentica
...
+        /// Attempts a password sign in for a user.
+        /// </summary>
+        /// <param name="user">The user to sign in.</param>
+        /// <param name="password">The password to attempt to sign in with.</param>
+        /// <param name="lockoutOnFailure">Flag indicating if the user account should be locked if the sign in fails.</param>
+        /// <returns>The task object representing the asynchronous operation containing the <see name="SignInResult"/>
+        /// for the sign-in attempt.</returns>
+        /// <returns></returns>
+        public virtual async Task<SignInResult> CheckPasswordSignInAsync(TUser user, string password, bool lockoutOnFailure)
+        {
```
brockallen: maybe i'm missing something, but i don't see where lockoutOnFailure is used... perhaps it's the diff view here in github :/

brockallen:
yep, looks good. thanks.

HaoK:
fe60304df54c43030d3eacfd9073e6a36990db05

Issue #850, opened 2016-06-01
SignInManager:ExternalLoginSignInAsync needs flag to bypass 2fa


brockallen:
Currently if a user's account in ASP.NET Identity is configured to require 2fa it's required for both password and external authentication. Often the external provider already does the 2fa. There should be a param on the ExternalLoginSignInAsync to bypass the 2fa step, but still keep the other checks (primarily the call to PreSignInCheck).

HaoK:
fa976e7765db498b7587d4028da76f533d1590ff

PR #932, opened 2016-08-08
Add ExternalSignIn overload that can bypass TFA

HaoK:
Fixes https://github.com/aspnet/Identity/issues/850

cc @divega @blowdart @brockallen

### DIFF
--- a/src/Microsoft.AspNetCore.Identity/SignInManager.cs
+++ b/src/Microsoft.AspNetCore.Identity/SignInManager.cs
@@ -433,7 +433,19 @@ namespace Microsoft.AspNetCore.Identity
         /// <param name="isPersistent">Flag indicating whether the sign-in cookie should persist after the browser is closed.</param>
         /// <returns>The task object representing the asynchronous operation containing the <see name="SignInResult"/>
         /// for the sign-in attempt.</returns>
-        public virtual async Task<SignInResult> ExternalLoginSignInAsync(string loginProvider, string providerKey, bool isPersistent)
+        public virtual Task<SignInResult> ExternalLoginSignInAsync(string loginProvider, string providerKey, bool isPersistent) 
+            => ExternalLoginSignInAsync(loginProvider, providerKey, isPersistent, bypassTwoFactor: false);
+
+        /// <summary>
+        /// Signs in a user via a previously registered third party login, as an asynchronous operation.
+        /// </summary>
+        /// <param name="loginProvider">The login provider to use.</param>
+        /// <param name="providerKey">The unique provider identifier for the user.</param>
+        /// <param name="isPersistent">Flag indicating whether the sign-in cookie should persist after the browser is closed.</param>
+        /// <param name="bypassTwoFactor">Flag indicating whether to bypass two factor authentication.</param>
+        /// <returns>The task object representing the asynchronous operation containing the <see name="SignInResult"/>
+        /// for the sign-in attempt.</returns>
+        public virtual async Task<SignInResult> ExternalLoginSignInAsync(string loginProvider, string providerKey, bool isPersistent, bool bypassTwoFactor)
         {
             var user = await UserManager.FindByLoginAsync(loginProvider, providerKey);
             if (user == null)
@@ -446,7 +458,7 @@ namespace Microsoft.AspNetCore.Identity
             {
                 return error;
             }
-            return await SignInOrTwoFactorAsync(user, isPersistent, loginProvider);
+            return await SignInOrTwoFactorAsync(user, isPersistent, loginProvider, bypassTwoFactor);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.AspNetCore.Identity/SignInManager.cs
```diff
@@ -582,7 +594,7 @@ private ClaimsIdentity CreateIdentity(TwoFactorAuthenticationInfo info)
         }
 
 
-        private async Task<SignInResult> SignInOrTwoFactorAsync(TUser user, bool isPersistent, string loginProvider = null)
+        private async Task<SignInResult> SignInOrTwoFactorAsync(TUser user, bool isPersistent, string loginProvider = null, bool bypassTwoFactor = false)
         {
```
brockallen: is bypassTwoFactor used in here?

brockallen:
yep, looks good. thanks.

src/Microsoft.AspNetCore.Identity/SignInManager.cs
```diff
@@ -433,7 +433,19 @@ public virtual Task ForgetTwoFactorClientAsync()
         /// <param name="isPersistent">Flag indicating whether the sign-in cookie should persist after the browser is closed.</param>
         /// <returns>The task object representing the asynchronous operation containing the <see name="SignInResult"/>
         /// for the sign-in attempt.</returns>
-        public virtual async Task<SignInResult> ExternalLoginSignInAsync(string loginProvider, string providerKey, bool isPersistent)
+        public virtual Task<SignInResult> ExternalLoginSignInAsync(string loginProvider, string providerKey, bool isPersistent)
```
troydai: This is a overload method for placing a default value for one parameter. Should we just have the one of the overload to be virtual? (I assume we can't change this in 1.1.0 since take away the virtual modifier will be a breaking change. But I'm just wondering.)

src/Microsoft.AspNetCore.Identity/SignInManager.cs
```diff
@@ -582,7 +594,7 @@ private ClaimsIdentity CreateIdentity(TwoFactorAuthenticationInfo info)
         }
 
 
-        private async Task<SignInResult> SignInOrTwoFactorAsync(TUser user, bool isPersistent, string loginProvider = null)
+        private async Task<SignInResult> SignInOrTwoFactorAsync(TUser user, bool isPersistent, string loginProvider = null, bool bypassTwoFactor = false)
```
troydai: is the `bypassTwoFactor` used in this method?

troydai: But the newly added parameter is not used in this private method, is it?

HaoK: GitHub pr diff looks messed up try looking at the commit diff.

src/Microsoft.AspNetCore.Identity/SignInManager.cs
HaoK: > This is a overload method for placing a default value for one parameter. Should we just have the one of the overload to be virtual? (I assume we can't change this in 1.1.0 since take away the virtual modifier will be a breaking change. But I'm just wondering.)
> 

I am not sure what the suggestion or ask is here. Can you clarify what you mean?

src/Microsoft.AspNetCore.Identity/SignInManager.cs
troydai: The parameter `bypassTwoFactor` is added to the `SignInOrTwoFactorAsync`, however it is not used in the method body. It is a private method so it is not for overriding. Why is this parameter added?

src/Microsoft.AspNetCore.Identity/SignInManager.cs
troydai: My suggestion is in following situation, only one method among all of its overloads should be virtual. The indention of the overloads seem to be fill in the default value. So this helps to avoid accidentally override the default value attempt in derived class.

None blocking. 

```
public virtual void Core(int parameter1, int parameter2)
{
    // body
}

public void Core(int parameter1)
{
    Core(parameter1, 0);
}

public void Core()
{
    Core(0, 0);
}
```

HaoK: Ah, well the original method was already virtual and should remain so, if they wish to override the method and change the behavior to call the new overload with byPassTwoFactor: true instead, that's their prerogative, its really up to the app what behavior they want for these flags.

src/Microsoft.AspNetCore.Identity/SignInManager.cs
HaoK: Weird, I will double check the branch and see why the diff is messed up, perhaps I forgot to push a commit

src/Microsoft.AspNetCore.Identity/SignInManager.cs
troydai: Make sense.

src/Microsoft.AspNetCore.Identity/SignInManager.cs
troydai: It feels like you did miss a commit.

HaoK: Yeah I did, ok pushed it now

troydai:
:shipit:

HaoK:
fa976e7765db498b7587d4028da76f533d1590ff

PR #936, opened 2016-08-16
Support for Entity Framework 6

gravity00:
Created a project, named Microsoft.AspNetCore.Identity.EntityFramework6, that adds support for Entity Framework 6 and can be released as a NuGet package.

Reasons:
- Entity Framework Core will take some time to provide support for some databases;
- Developers are already using ASP.NET Core and Identity but prefer to stick with Entity Framework 6 for now due to stability, preference or legacy reasons;

Missing:
- Entity Framework 6 tests;


HaoK:
Thanks for reaching out, we think this looks like a component that can be build on top of identity without actually living inside this repo.  Among other things, it would need tests.

gravity00:
@HaoK thanks for the feedback. Some of the clients I work with have this requirements and can't commit to Entity Framework Core in the near future. We are already using this NuGet (different namespace ofc) but it would be nice to have an official NuGet from Microsoft.
What do you recommend to do since it wouldn't be in this repo?

> Note: didn't create any tests because I wanted to have feedback first. Since we are using it, we are trying to create a direct map from EF Core tests to EF6 at the moment.

PR #943, opened 2016-08-25
Delete Helios.cmd

tugberkugurlu:
I believe it's not used anymore.


troydai:
Yup. /cc @HaoK ?

HaoK:
Yup safe to delete

troydai:
Thanks @tugberkugurlu

PR #946, opened 2016-08-29
remove About and Contact links from the sample

tugberkugurlu:
Their corresponding views are missing and when clicked, it gives 404. Not sure if removing is OK (seems so).


HaoK:
Seems reasonable, I'll merge

HaoK:
Merged thanks! 4af4a434c563bc89064b18ff6867fb52d6bfd7d6

Issue #617, opened 2015-11-04
Provide UserClaimsPrincipalFactory<TUser>


tugberkugurlu:
Currently, `IUserClaimsPrincipalFactory<TUser>` has an implementation as `UserClaimsPrincipalFactory<TUser, TRole>`. However, this is not convinent for implementations which don't support `IRoleStore<TRole>`. Providing `UserClaimsPrincipalFactory<TUser>` would be great.

tugberkugurlu:
Happy to PR for this if this sounds reasonable.

HaoK:
Feel free to submit a PR for this sure :)

rustd:
https://github.com/aspnet/Identity/issues/581

jasonwilliams200OK:
Any progress on this? We should also have `IServiceCollection.AddIdentity<TUser>` without having to specify `TRole`.

HaoK:
@blowdart we should discuss this in triage again

HaoK:
This was done in 2.0 to support roles being optional

HaoK:
Done via https://github.com/aspnet/Identity/issues/1269

PR #952, opened 2016-09-01
Provide UserClaimsPrincipalFactory<TUser>

tugberkugurlu:
Aims to solve #617.

Currently, `IUserClaimsPrincipalFactory<TUser>` has an implementation as `UserClaimsPrincipalFactory<TUser, TRole>`. However, this is not convenient for implementations which don't support `IRoleStore<TRole>`. Providing `UserClaimsPrincipalFactory<TUser>` would be great.
- [ ] Add tests, see [UserClaimsPrincipalFactoryTest](https://github.com/aspnet/Identity/blob/dev/test/Microsoft.AspNetCore.Identity.Test/UserClaimsPrincipalFactoryTest.cs).


src/Microsoft.AspNetCore.Identity/UserClaimsPrincipalFactoryOfTUser.cs
```diff
@@ -0,0 +1,94 @@
...
+                Options.ClaimsIdentity.UserNameClaimType,
+                Options.ClaimsIdentity.RoleClaimType);
+            id.AddClaim(new Claim(Options.ClaimsIdentity.UserIdClaimType, userId));
+            id.AddClaim(new Claim(Options.ClaimsIdentity.UserNameClaimType, userName));
+            if (UserManager.SupportsUserSecurityStamp)
+            {
+                id.AddClaim(new Claim(Options.ClaimsIdentity.SecurityStampClaimType, 
+                    await UserManager.GetSecurityStampAsync(user)));
+            }
+            if (UserManager.SupportsUserRole)
```
tugberkugurlu: I am not sure about this. Is it still valuable to check this and add roles into the claims?

tugberkugurlu:
I created abstract class on `UserClaimsPrincipalFactoryBase<TUser>` https://github.com/aspnet/Identity/pull/952/commits/bb06284d018b1339f433b8d609f8e83b8c0f73b0. However, I am not sure if that's a good concept to share across `UserClaimsPrincipalFactory` impls.

HaoK:
This will finally happen via https://github.com/aspnet/Identity/issues/1269

PR #955, opened 2016-09-07
Use TaskCache class from Microsoft.Extensions.TaskCache.Sources

pakrym:
Instead of Task.FromResult(0)

@pranavkm


pranavkm:
:shipit:

PR #960, opened 2016-09-12
Update README with .NET SDK link

crystaju:
provide direct link to Visual Studio page for .NET and NuGet.


blowdart:
Why? We have a standard readme for all of asp.net's repos. Why change this readme to link to an SDK that doesn't have anything to do with ASP.NET identity?

blowdart:
Rejecting as this isn't the place for this link.

PR #963, opened 2016-09-13
[Design] Add TwoFactorRecoveryCode support and TwoFactorTokenprovider for Authenticator apps

HaoK:
Only went as far as verifying that the mainline scenario of generating/redeeming a token once so far.

We are reusing the existing token table via encoding the token id into the login provider.

The table is setup today with the key as `<userId> + <login provider> + <name>`

For the new IUserTokenStore, we are translating the new IdentityToken's Id, Type into the old table by: LoginProvider = "[Local]:<tokenId>" and using the Name field as the token Type.

@divega @blowdart


HaoK:
The main purpose of this initial PR is to start settling on the names/shape of the public surface area of UserManager/SignInManager:

New IdentityToken POCO:
```
        public string Id { get; }
        public string Type { get; }
        public string Value { get; }
```

UserManager gets:
`GetTokens`(all, and with type overload)
`StoreTokens(IEnumerable<IdentityToken>)`

SignInManager gets:
'GenerateNewRecoveryCodes(TUser user, int numCodes)`
`UseRecoveryCode(TUser user, string code)`

HaoK:
5aed9742a4f1a35558f5ca644e981398df7017c1

Issue #964, opened 2016-09-14
Specify custom token provider when using the methods which uses token provider.


akanmuratcimen:
Currently we can change the default token providers on TokenOptions in IdentityOptions. But in my case I need to specify custom token provider when using the method (assume that GenerateEmailConfirmationTokenAsync() and ConfirmEmailAsync()) because of the application which i made send a link if the user uses web client or send a code (Totp) if the user uses mobile client. In short, what do you think to specify token providers inline when generating and consuming the methods.

HaoK:
We are adding the ability to specify instances in the provider map, I'm not sure if that completely addresses your issue, but you will be able to configure a different instance for each method if you so like via https://github.com/aspnet/Identity/pull/983

If you want to switch between different providers in a different fashion you probably have to drop down and use the lower level methods: 

Generate/VerifyUserTokenAsync(user, "YourCustomProvider", "ResetPassword"), those public apis are just sugar on top of that method with the IdentityOptions.Tokens.XyzProviderName hardcoded.

HaoK:
If you are managing the token provider instances instead of the user manager, you can just directly call into the provider with the appropriate purpose/user, you don't need to go through the user manager in this case.

PR #965, opened 2016-09-14
Added specify token provider option to methods.

akanmuratcimen:
#964 
Added specify token provider option to 

> GeneratePasswordResetTokenAsync
> ResetPasswordAsync

> GenerateEmailConfirmationTokenAsync
> ConfirmEmailAsync

> GenerateChangeEmailTokenAsync
> ChangeEmailAsync


HaoK:
If you are managing the token provider instances instead of the user manager, you can just directly call into the provider with the appropriate purpose/user, you don't need to go through the user manager in this case.

Issue #939, opened 2016-08-19
Change SignInManager's private helper methods to protected.


RoySalisbury:
I have to implement a "PinSignInAsync" method on my SignInManager.  This will basically be (almost) exactly like the PasswordSignInAsync" method, but will be calling a new "UserManager.CheckPinAsync" method and does not have to deal with TwoFactor code.

To implement this I looked at the existing password method, and there are 4 private methods that it uses.  If these were protected, then I could use them in my own derived class so code is reused and not duplicated.  

        private async Task<bool> IsLockedOut(TUser user)
        private async Task<SignInResult> LockedOut(TUser user)
        private async Task<SignInResult> PreSignInCheck(TUser user)
        private Task ResetLockout(TUser user)

I would also suggest reviewing the other private and internal methods/classes to see if the would benefit from being protected/protected internal.

HaoK:
dbec1c6236ac5cc6bc31db52686259c64c89a7d3

Issue #969, opened 2016-09-17
UserManager CancellationToken private


Ketrex:
Why is the CancellationToken in UserManager private and not protected so it is available to derived class methods and properties?

```
public class Microsoft.AspNetCore.Identity.UserManager<TUser>
{
    // why private?
    private CancellationToken CancellationToken => _context?.RequestAborted ?? CancellationToken.None;
}
```

I have created a new class inheriting from UserManager and added a few new async methods that could make use of the cancellation token if only it was available.

example:

```
public class UserManager<TUser> : Microsoft.AspNetCore.Identity.UserManager<TUser> ...
{
    ...

    // Retrieving custom firstname property from TUser
    public async Task<string> GetFirstNameAsync(TUser user)
    {
        ThrowIfDisposed();
        return await Store.GetFirstNameAsync(user, CancellationToken); // won't work because cancellation token is not available in UserManager
    }
}
```

HaoK:
dbec1c6236ac5cc6bc31db52686259c64c89a7d3

PR #973, opened 2016-09-23
Expose more for extensibility

HaoK:
Fixes https://github.com/aspnet/Identity/issues/939
https://github.com/aspnet/Identity/issues/969

@divega


HaoK:
These used to be all public and the unit tests tweak the Managers this way still. They were made internal when we switched to DI to initialize in 3.0.

divega:
Thanks for the information. One more question: what criteria are you using to decide what to make virtual and what not?

HaoK:
I only made the protected methods virtual since there was no harm in letting them tweak those behaviors.  The properties are get/set, so those should be fine non virtual

HaoK:
dbec1c6236ac5cc6bc31db52686259c64c89a7d3

Issue #944, opened 2016-08-26
Unique role names


cjcfreitas:
Role names must be unique.
This is enforced by `RoleValidator`.
Shouldn't this be also enforced in the EntityFramework model?
File IdentityDbContext.cs line 170 would be:
`b.HasIndex(r => r.NormalizedName).HasName("RoleNameIndex").IsUnique();`

HaoK:
5d759abcde8e652fa83ad1a7099fffffb53ca648

PR #974, opened 2016-09-23
Make role name index unique by default

HaoK:
Fixes https://github.com/aspnet/Identity/issues/944

@divega


HaoK:
Yes the manager actually already guarantees that no role with the same name exists. This is a redundant check for the database layer.

HaoK:
5d759abcde8e652fa83ad1a7099fffffb53ca648

Issue #958, opened 2016-09-11
Make it possible to extend/hook Security stamp validation to enable keeping current cookie claims when renewing cookie


brockallen:
Vital claims that might only be captured at login time are lost when the security stamp is renewed. The implementation itself even has a comment to review this behavior:

```
var user = await _signInManager.ValidateSecurityStampAsync(context.Principal);
if (user != null)
{
    // REVIEW: note we lost login authenticaiton method
    context.ReplacePrincipal(await _signInManager.CreateUserPrincipalAsync(user));
    context.ShouldRenew = true;
}
```

I'd suggest simply using the current ticket's claims, but replacing the security stamp. If it's deemed that this is not to be changed/fixed, then perhaps add a flag to allow both behaviors?

This came up as part of using ASP.NET Identity within IdentityServer4: https://github.com/IdentityServer/IdentityServer4/issues/277

HaoK:
Yeah the current behavior assumes the UserPrincipalFactory is the source of truth for creating a current up to date principal (we don't want to just refresh the old cookie at all).  Can you just plug in your own derived factory to add the additional claims or are they stored anywhere other than the cookie?

brockallen:
They're dynamic.

Also this behavior of reloading only from the DB behavior will break OIDC federated sign-out behavior given the need to maintain the `sid` claim coming from an external provider:  https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.OpenIdConnect/OpenIdConnectHandler.cs#L123

HaoK:
Not at all, this feature is about clearing out security related claims or any other things that might have changed in the user.  I.e. some permission has been revoked, so the cookie must be regenerated as its no longer valid.  The stamp claim is only used to trigger the forced refresh.

HaoK:
We have no way to tell what's sensitive or not, hence the entire cookie needs to be regenerated (which is what the IUserClaimsPrincipal's job is)

brockallen:
The security stamp is only changed on these operations:
* create user
* set username
* remove login
* set/change email
* set/change phone
* set 2fa
* update password

it's not set when claims or roles change, so I'm not sure I agree with this:

> this feature is about clearing out security related claims or any other things that might have changed in the user

so it really seems the security stamp feature is about when the user's credentials change, not when their data changes.

HaoK:
[...]

PR #977, opened 2016-09-26
Add OnSecurityStampReplacePrincipal event

HaoK:
Fixes https://github.com/aspnet/Identity/issues/958

Adds a new event called from the default security stamp validator which allows tweaking the user principal regeneration using the old principal.

Not super happy with the name (or the fact that the event lives at the top level of IdentityOptions)

cc @divega @blowdart

### DIFF
--- a/src/Microsoft.AspNetCore.Identity/IdentityOptions.cs
+++ b/src/Microsoft.AspNetCore.Identity/IdentityOptions.cs
@@ -79,6 +79,6 @@ namespace Microsoft.AspNetCore.Builder
         /// <summary>
         /// Invoked when the default security stamp validator replaces the user's ClaimsPrincipal in the cookie.
         /// </summary>
-        public Func<SecurityStampReplacingPrincipalContext, Task> OnSecurityStampReplacingPrincipal { get; set; }
+        public Func<SecurityStampRefreshingPrincipalContext, Task> OnSecurityStampReplacingPrincipal { get; set; }
     }
 }
--- a/src/Microsoft.AspNetCore.Identity/IdentityOptions.cs
+++ b/src/Microsoft.AspNetCore.Identity/IdentityOptions.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Threading.Tasks;
 using Microsoft.AspNetCore.Identity;
 
 namespace Microsoft.AspNetCore.Builder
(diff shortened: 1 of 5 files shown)


brockallen:
> or the fact that the event lives at the top level of IdentityOptions

This is why there should be a security stamp options (maybe in 2.0.0 when you can add a breaking change).

HaoK:
@Tratcher mind reviewing as well since you are the only other one with some context :)

brockallen:
After a quick glance, it looks ok to me (FWIW)

Tratcher:
What's the other event where the user can copy claims from the external cookie to the identity cookie?

HaoK:
CreatingTicket I believe is the other OAuth event that can be hooked

Tratcher:
? Wasn't there an event at the identity layer?

HaoK:
Nope, shuttling claims around was always done directly via cookie events.  We added a method which took care of some of the token refreshing, but identity doesn't expose any cookie related events, it just exposes the cookie options directly (until this PR anyways...)

src/Microsoft.AspNetCore.Identity/SecurityStampReplacingPrincipalContext.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    /// <summary>
+    /// Used to pass information during the SecurityStamp validation event.
+    /// </summary>
+    public class SecurityStampReplacingPrincipalContext
+    {
+        /// <summary>
+        /// The principal contained in the current cookie.
+        /// </summary>
+        public ClaimsPrincipal Current { get; set; }
```
Tratcher: CurrentPrincipal and NewPrincipal?

src/Microsoft.AspNetCore.Identity/IdentityOptions.cs
```diff
@@ -74,5 +75,10 @@ public class IdentityOptions
         /// The <see cref="TimeSpan"/> after which security stamps are re-validated.
         /// </value>
         public TimeSpan SecurityStampValidationInterval { get; set; } = TimeSpan.FromMinutes(30);
+
+        /// <summary>
+        /// Invoked when the default security stamp validator replaces the user's ClaimsPrincipal in the cookie.
+        /// </summary>
+        public Func<SecurityStampReplacingPrincipalContext, Task> OnSecurityStampReplacingPrincipal { get; set; }
```
Tratcher: Refreshing?

HaoK:
Sure, updated to suggested names, New/Current Principal + Replacing => Refreshing

HaoK:
aba7cf96b9ac7c910ee7db5e34fa7cf78516116a

PR #1020, opened 2016-11-07
Removed duplicate call to IsLockedOut()

MarcoMiltenburg:
There's no need to call IsLockedOut() anymore as this is already done in PreSignInCheck().


HaoK:
Good catch, I'll merge this

HaoK:
Merged thanks! bb60afba28d7775e610afafd85cce810c9200374

Issue #1019, opened 2016-11-07
Simultaneous login control


pushist1y:
Our system requires possibility to restrict user from logging in simultaneously from different browsers. Currently Identity does not support such restrictions. 

Currently I've implemented this by inheriting from SignInManager and adding another entity describing logged in users. 

Should I try to implement this inside Identity codes and PR it here or is this functionality not wanted in this repository?

blowdart:
This isn't a common enough scenario for us to pull it into the main code, but thank you for the offer.

PR #1022, opened 2016-11-09
User online control

pushist1y:
The application we are developing requires disabling users to log in simultaneously from different browsers and also restricting maximum number of signed in users.

Firstly i've implemented that functionality on top of Identity and then i thought it could be used by someone else.

Changes:
Added LastActivity timestamp field to `IdentityUser`
Added `IUserActivityStore` interface to query active users
Added checks in `CanSignInAsync` method 
Added automatic forced signing off of idle users if maximum online is reached

Addresses #1019


blowdart:
This looks like the right approach, but it's not common enough for us to pull in.

PR #1025, opened 2016-11-09
AddEntityFramework stores now infers generics from TUser/TRole, new() constraint now required

HaoK:
Maybe we do want to do this for 1.1, looks like we missed adding the generic parameter for the token when we added the other entities.

I ran into this since I needed to add a navigation property to be able to setup the foreign key...

@divega @blowdart


divega:
@HaoK any idea of how this affects upgrading an existing application? 

>  looks like we missed adding the generic parameter for the token when we added the other entities.

Do you mean when we did some other change in 1.1? Or has it been missing from 1.0?

HaoK:
We fixed a different bug where we were one missing generic with the missing TRoleClaim on the UserStores.  This one is a bit different and not as bad, since we didn't have a generic for TUserToken on IdentityUser because until this change we didn't have a navigation property which required it...  So I guess as things stands its not critical to fix now.  So I guess we can wait til 1.2 for that reason, altho we will need to take a small breaking change when we add the additional generic argument

src/Microsoft.AspNetCore.Identity.EntityFrameworkCore/IdentityDbContext.cs
```diff
@@ -162,6 +162,7 @@ protected override void OnModelCreating(ModelBuilder builder)
                 b.HasMany(u => u.Claims).WithOne().HasForeignKey(uc => uc.UserId).IsRequired();
                 b.HasMany(u => u.Logins).WithOne().HasForeignKey(ul => ul.UserId).IsRequired();
                 b.HasMany(u => u.Roles).WithOne().HasForeignKey(ur => ur.UserId).IsRequired();
+                b.HasMany(u => u.Tokens).WithOne().HasForeignKey(uc => uc.UserId).IsRequired();
```
divega: nit: ut instead of uc

divega:
> we will need to take a small breaking change when we add the additional generic argument

I hate to ask :smile: but what would it take to do it in a non-breaking way? Does an `IdentityUser` that derives from the most generic one but fixes TUserToken help?

HaoK:
Lets see what adding a new() constraint enables

HaoK:
See if we can make something like `AddEntityFrameworkStores<TContext> and just reflect off of TUser/TRole to infer everything which is already on the builder.

HaoK:
Updated this PR to contain the changes that add the new() constraints, and eliminates the need for any explicit calls to AddUser/RoleStore, `AddEntityFrameworkStores<TContext>` should "do the right thing" now when deriving from our default pocos.

HaoK:
@divega appveyor CI appears to detect breaking changes now as part of the runs, this obviously makes some, what's the process(if any) do we need to follow for exemptions/intentional breaks?

divega:
@HaoK see https://github.com/aspnet/BuildTools/wiki/Api-Check.

HaoK:
Cool, we are ok adding exceptions for these breaking changes right?

divega:
My evaluation is that the the potential impact of the `new()` constraint per se is tiny. Are we doing more than that?

HaoK:
Also removed a few of the overloads of the store classes that took different amounts of generics, i.e.

```
public class UserStore<TUser, TRole, TContext, TKey> : UserStore<TUser, TRole, TContext, TKey, IdentityUserClaim<TKey>, IdentityUserRole<TKey>, IdentityUserLogin<TKey>, IdentityUserToken<TKey>>
```
got replaced with one that takes IdentityRoleClaim as well.

And the overload for AddEntityFramework that specified TKey was removed since we now can infer that anyways.

HaoK:
57a25d080d1d0337b3b2dd8a2906f74a6e94370c

Issue #1016, opened 2016-11-05
Consider having the UserValidator ensure that SecurityStamp != null


shima20:
When I called `var principal = await _signInManager.CreateUserPrincipalAsync(user);` , I was receiving an ArgumentNullException and had no idea what the issue was.  Kevin from Openiddict told me to register UserClaimsPrincipalFactory to step in and see what field was null.  This led me to find that my SecurityStamp was null and therefore caused the exception.  Is this a bug?

```
if (UserManager.SupportsUserSecurityStamp)
            {
                id.AddClaim(new Claim(Options.ClaimsIdentity.SecurityStampClaimType,
                    await UserManager.GetSecurityStampAsync(user)));
            }
```

HaoK:
SecurityStamps should never be null for a store that supports user security stamp.

PinpointTownes:
> SecurityStamps should never be null for a store that supports user security stamp.

Except it was, in this case (using the default EF Core store).

It's also worth mentioning that there's nothing that prevents setting `SecurityStamp` to `null`: https://github.com/aspnet/Identity/blob/dev/src/Microsoft.AspNetCore.Identity.EntityFrameworkCore/UserStore.cs#L1242-L1252

HaoK:
Typically this usage is updated by the manager, I mean sure you can set it to null directly but things will blow up.

We could consider adding validation in the default user validator for this since this does seem to trip more than a few people up.

PinpointTownes:
@shima20's case is obviously an atypical case, as he somehow managed to get a null security stamp.

FWIW, I don't think the principal factory should shock on null values. It should be a bit smarter and ignore null claim values.

HaoK:
Not for something like this, the security stamp is critical well for 'security' :)  Its a major error if its null...  similar to having no password hash or user name etc, its the user validators job to make sure the user is in a good state

PinpointTownes:
Critical seems a bit excessive (it's only used for cookies invalidation), but okay. As long as there's a check somewhere that prevents this illegal case, I'm fine :smile:

HaoK:
4dd38e8ca81a25d55fd94d7efd25b1ff256b65ab

PR #1026, opened 2016-11-11
UserManager now blocks null security stamps

HaoK:
Fixes https://github.com/aspnet/Identity/issues/1016

cc @divega @blowdart

### DIFF
--- a/src/Microsoft.AspNetCore.Identity.EntityFrameworkCore/UserStore.cs
+++ b/src/Microsoft.AspNetCore.Identity.EntityFrameworkCore/UserStore.cs
@@ -1247,6 +1247,10 @@ namespace Microsoft.AspNetCore.Identity.EntityFrameworkCore
             {
                 throw new ArgumentNullException(nameof(user));
             }
+            if (stamp == null)
+            {
+                throw new ArgumentNullException(nameof(stamp));
+            }
             user.SecurityStamp = stamp;
             return TaskCache.CompletedTask;
         }
--- a/src/Microsoft.AspNetCore.Identity/IdentityErrorDescriber.cs
+++ b/src/Microsoft.AspNetCore.Identity/IdentityErrorDescriber.cs
@@ -14,7 +14,7 @@ namespace Microsoft.AspNetCore.Identity
         /// <summary>
         /// Returns the default <see cref="IdentityError"/>.
         /// </summary>
-        /// <returns>The default <see cref="IdentityError"/>,</returns>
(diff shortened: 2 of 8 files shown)


HaoK:
Note we usually take care of modifying the security stamp, but this guards against some edge cases where they mucked with the entity directly, or end up with a null security stamp via other means outside of our APIs

PinpointTownes:
So, in the "rare" case where a security steam would be missing, the user wouldn't be able to update his profile, right?

HaoK:
Yes, basically its a developer error to end up with a user with a security stamp missing, you won't be able to do save a user with no security stamp anymore.

PinpointTownes:
> its a developer error to end up with a user with a security stamp missing

My point is that it shouldn't be returned to the end user, as there's nothing it can do to fix it.
Is there a particular reason you prefer returning an error instead of generating a new security stamp on the fly?

HaoK:
Ah yes, your point about this being a pretty useless message to the user is valid.  I guess this behavior also shouldn't be optional based on the user validator, this check should be explicit in the user manager and throw an exception instead.

Its not really safe to automatically generate a new security stamp, that would have the effect of result in really hard to repro cases of users that are suddenly signed out.

HaoK:
@divega @blowdart do you guys agree that a missing security stamp on a user when the feature is enabled should result in a runtime exception instead of a 'recoverable' identity error.

PinpointTownes:
> Its not really safe to automatically generate a new security stamp, that would just result in really hard to repro cases of users that are suddenly signed out.

Well, if a user has a null security stamp at some point, he will be unable to log in at all, as it will crash when creating the `ClaimsPrincipal` :trollface:

blowdart:
Sure, throw. It's unexpected enough to be an exceptional event.

HaoK:
Updated moving null stamp check to unoverridable user manager code.  @divega look reasonable?

PinpointTownes:
No love for `UserStore.SetSecurityStampAsync`? If null stamps are illegal, maybe it should throw an `ArgumentException` when it's null or empty.

https://github.com/aspnet/Identity/blob/dev/src/Microsoft.AspNetCore.Identity.EntityFrameworkCore/UserStore.cs#L1242

HaoK:
Sure that's valid defense in depth, but its very hard to call that today using the identity APIs as its only called in the UpdateSecurityStamp internal method which never passes in null, but it doesn't hurt to add

HaoK:
4dd38e8ca81a25d55fd94d7efd25b1ff256b65ab

Issue #1001, opened 2016-10-20
Make AddEntityFrameworkStores smarter with generics


lixiaoyuan:
```
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. ---> System.ArgumentException: GenericArguments[0], 'BusinessDb.Cor.EntityModel.SystemUser`1[System.Guid]', on 'Microsoft.AspNetCore.Identity.EntityFrameworkCore.UserStore`4[TUser,TRole,TContext,TKey]' violates the constraint of type 'TUser'. ---> System.TypeLoadException: GenericArguments[0], 'BusinessDb.Cor.EntityModel.SystemUser`1[System.Guid]', on 'Microsoft.AspNetCore.Identity.EntityFrameworkCore.UserStore`8[TUser,TRole,TContext,TKey,TUserClaim,TUserRole,TUserLogin,TUserToken]' violates the constraint of type parameter 'TUser'.
   at System.RuntimeTypeHandle.Instantiate(RuntimeTypeHandle handle, IntPtr* pInst, Int32 numGenericArgs, ObjectHandleOnStack type)
   at System.RuntimeTypeHandle.Instantiate(Type[] inst)
   at System.RuntimeType.MakeGenericType(Type[] instantiation)
   --- End of inner exception stack trace ---
   at System.RuntimeType.ValidateGenericArguments(MemberInfo definition, RuntimeType[] genericArguments, Exception e)
   at System.RuntimeType.MakeGenericType(Type[] instantiation)
   at Microsoft.Extensions.DependencyInjection.IdentityEntityFrameworkBuilderExtensions.GetDefaultServices(Type userType, Type roleType, Type contextType, Type keyType)
   at Microsoft.Extensions.DependencyInjection.IdentityEntityFrameworkBuilderExtensions.AddEntityFrameworkStores[TContext,TKey](IdentityBuilder builder)
   at ErpApplication.Web.Startup.ConfigureServices(IServiceCollection services) in D:\外接系统\ErpWebApplication\src\ErpApplication.Web\Startup.cs:line 40
   --- End of inner exception stack trace ---
   at System.RuntimeMethodHandle.InvokeMethod(Object target, Object[] arguments, Signature sig, Boolean constructor)
   at System.Reflection.RuntimeMethodInfo.UnsafeInvokeInternal(Object obj, Object[] parameters, Object[] arguments)
[...]

Issue #876, opened 2016-06-17
Consider improvements for custom key type in AddEfStores


HaoK:
We could special case IdentityDbContext derived DbContext's to determine the PKey type to register the appropriate type automatically.

https://github.com/aspnet/Identity/issues/780#issuecomment-226687826

HaoK:
13ae7b21f3459e97cd5853b6e6ccb229dac1bafe

PR #1031, opened 2016-11-16
Improve error for AddEntityFrameworkStores

HaoK:
Fixes https://github.com/aspnet/Identity/issues/1001

@divega

### DIFF
--- a/src/Microsoft.AspNetCore.Identity.EntityFrameworkCore/IdentityEntityFrameworkBuilderExtensions.cs
+++ b/src/Microsoft.AspNetCore.Identity.EntityFrameworkCore/IdentityEntityFrameworkBuilderExtensions.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Reflection;
 using Microsoft.AspNetCore.Identity;
 using Microsoft.AspNetCore.Identity.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore;
@@ -23,7 +24,7 @@ namespace Microsoft.Extensions.DependencyInjection
         public static IdentityBuilder AddEntityFrameworkStores<TContext>(this IdentityBuilder builder)
             where TContext : DbContext
         {
-            builder.Services.TryAdd(GetDefaultServices(builder.UserType, builder.RoleType, typeof(TContext)));
+            AddStores(builder.Services, builder.UserType, builder.RoleType, typeof(TContext), typeof(string));
             return builder;
         }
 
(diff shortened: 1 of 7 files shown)


src/Microsoft.AspNetCore.Identity.EntityFrameworkCore/IdentityEntityFrameworkBuilderExtensions.cs
```diff
@@ -38,26 +39,28 @@ public static IdentityBuilder AddEntityFrameworkStores<TContext>(this IdentityBu
...
-            keyType = keyType ?? typeof(string);
-            userStoreType = typeof(UserStore<,,,>).MakeGenericType(userType, roleType, contextType, keyType);
-            roleStoreType = typeof(RoleStore<,,>).MakeGenericType(roleType, contextType, keyType);
-
-            var services = new ServiceCollection();
-            services.AddScoped(
+            var identityUserType = typeof(IdentityUser<>).MakeGenericType(keyType);
+            if (!identityUserType.GetTypeInfo().IsAssignableFrom(userType.GetTypeInfo()))
+            {
+                throw new InvalidOperationException(Resources.NotIdentityUser);
```
divega: AFAIR there were a couple of ways to hit the current `TargetInvocationException`. Need to search around to make sure we are covered by this check.

HaoK:
We could just catch all exceptions and rethrow with our message and the original exception as an inner exception to be safe.

divega:
@HaoK I looked around, and I am almost sure the two cases I had seen in which we would throw the exception are:
- Inherit from the IdentityUser that has more generic arguments
- Change TKey to a different type from the default but forget to pass the new type as a generic argument in the call to AddEntityFrameworkStores. 

For the latter I noticed you filed https://github.com/aspnet/Identity/issues/876.

src/Microsoft.AspNetCore.Identity.EntityFrameworkCore/Resources.resx
```diff
@@ -117,6 +117,14 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="NotIdentityRole" xml:space="preserve">
+    <value>AddEntityFrameworkStores can only be called with a role that derives from IdentityRole&lt;TKey&gt;. If you are specifying more generic arguments, use IdentityBuilder.AddRoleStore&lt;TStore&gt;() instead.</value>
```
divega: Couple of observations:
- I don't think the mention of `IdentityBuilder` helps here. I would just remove it. 
- Customers also need to derive their own store classes (by overriding a method in the abstract base class), right? If that is the case, can you also mention it in the messages?

divega:
BTW, a couple of simpe searches shows that this exception is [one of the most common reasons customers file bugs on the Identity repo](https://github.com/aspnet/Identity/search?q=GenericArguments+&type=Issues&utf8=%E2%9C%93) and [ask question about Identity over in StackOverflow](http://stackoverflow.com/search?q=Identity+GenericArguments). 

If we get it right this could help a lot!

HaoK:
I'll try to fix #876 in this PR as well

HaoK:
Updated with the error message tweaks, also added support for infering the TKey from any DbContext that dervies from IdentityDbContext<TUser,TRole,TKey>.  This makes the second overload unnecessary but we can keep it for breaking changes, and also supporting custom DbContext's that don't derive from our DbContext (which isn't a requirement today)

HaoK:
13ae7b21f3459e97cd5853b6e6ccb229dac1bafe

PR #1039, opened 2016-11-29
Run tests on Linux/MacOS as well

HaoK:
Trying on one test to see what travis does


HaoK:
:boom: no worky, localdb is the issue still, would need to run an external sql server

Issue #696, opened 2015-12-31
Microsoft.AspNet.Identity.Test as a nuget package


SharePointRadi:
Do you have plans to release the Microsoft.AspNet.Identity.Test project as a nuget package? Currently, if we are building a custom user store and want to reuse the UserManagerTestBase class (from the Shared folder), we have to clone, build, reference, etc.

It would be useful if it is reusable.

HaoK:
Seems reasonable to consider in the future...backlog @divega @rustd ?

divega:
Yes. We did this with the EF tests for provider writers and has been working well.

SharePointRadi:
Further to this discussion, UserManager.cs has internal methods that are referenced from UserManagerTestBase.cs.

This means that we (provider writers?) could not really reuse UserManagerTestBase because it refers those internal methods. I had a similiar need and had to use reflection to refer to those internal methods.

https://github.com/OneBitSoftware/Microsoft.AspNet.Identity.EntityFramework6/blob/master/test/Microsoft.AspNet.Identity.EntityFramework6.Test/TestingUserManager.cs

HaoK:
5e4e9ccdf164b622872c397c27b75d2bc8a329da

PR #1043, opened 2016-12-06
Create Identity.Specification.Tests package

HaoK:
- Moved most of the shared tests code here.
- Switched to new package for tests
- Turned on public API checks for Core store (and fixed the missing virtuals resulting from this)

@divega


HaoK:
Fixes https://github.com/aspnet/Identity/issues/696

src/Microsoft.AspNetCore.Identity.EntityFrameworkCore/RoleStore.cs
```diff
@@ -220,7 +220,7 @@ public RoleStore(TContext context, IdentityErrorDescriber describer = null)
         /// <param name="role">The role whose ID should be returned.</param>
         /// <param name="cancellationToken">The <see cref="CancellationToken"/> used to propagate notifications that the operation should be canceled.</param>
         /// <returns>A <see cref="Task{TResult}"/> that contains the ID of the role.</returns>
-        public Task<string> GetRoleIdAsync(TRole role, CancellationToken cancellationToken = default(CancellationToken))
+        public virtual Task<string> GetRoleIdAsync(TRole role, CancellationToken cancellationToken = default(CancellationToken))
```
divega: Not against this change, but I am curious on how it is related.

src/Microsoft.AspNetCore.Identity.Specification.Tests/TestRole.cs
```diff
@@ -35,6 +35,9 @@ public TestRole(string roleName) : this()
     /// <typeparam name="TKey"></typeparam>
     public class TestRole<TKey> where TKey : IEquatable<TKey>
     {
+        /// <summary>
```
divega: FWIW, I did not expect the specification package for providers to contain entity types because I thought each provider would bring their own. I rather expected the specification package to contain a set of base test classes which provider writers could derive from and template methods which provider writers could override/implement. 

Are we taking a different approach?

HaoK: The entity types are optional, but they were shared between the two in memory implementations, I can move them back into our shared folder if you like.  The EF store tests don't use them today.

HaoK:
Updated moving the test entities back to shared.  I think as part of this PR its time to spend a bit of time on engineering debt, I'll revisit the test setup and make a pass to make it more DI friendly since it hasn't really been updated since v2, if we are going to make it public, it should get some more love...

HaoK:
Actually looks like I already did the test cleanup I was thinking about (but forgot)...

What should we name the base class?  I renamed it from UserManagerTestBase => IdentityApiTestBase since it was testing more than user manager functionality (also role manager), its basically testing the entire surface area.

HaoK:
I think the naming should probably line up with the package, so maybe IdentitySpecificationTestBase?

divega:
@HaoK I am fine with that since we don't seem to have finer granularity than that.

HaoK:
5e4e9ccdf164b622872c397c27b75d2bc8a329da

PR #1053, opened 2016-12-12
Fix property name

### LINKED ISSUE #1052: Wrong property name
Hi,

I'm receiving the error below when I try to change a user's password using the Microsoft.AspNet.Identity.AspNetCoreCompat project.

> An exception of type 'System.ArgumentException' occurred in EntityFramework.dll but was not handled in user code
> 
> Additional information: The property 'ConcurrencyToken' on type 'ApplicationUser' is not a primitive or complex property. The Property method can only be used with primitive or complex properties. Use the Reference or Collection method.

The cause is in the line https://github.com/aspnet/Identity/blob/dev/src/Microsoft.AspNet.Identity.AspNetCoreCompat/IdentityDbContext.cs#L129.

There is no ConcurrencyToken property in the IdentityUser class. Changing the line to `entityEntry.Property("ConcurrencyStamp").CurrentValue = Guid.NewGuid().ToString();` fixed the error.

And I had to convert the Guid to string since the type of the ConcurrencyStamp property is string.

How to replicate:

* Create an ASP.NET MVC 4.6.1 application with Individual User Accounts
* Reference the AspNet.Identity.CoreCompat assembly
* Change the ApplicationUser class to inherit from the Microsoft.AspNet.Identity.CoreCompat.IdentityUser
* Change the ApplicationDbContext class to inherit from the Microsoft.AspNet.Identity.CoreCompat.IdentityDbContext
* Run the application
* Create a user and try to change the password

fabiano:
Fix #1052 issue.


src/Microsoft.AspNet.Identity.AspNetCoreCompat/IdentityDbContext.cs
```diff
@@ -126,16 +126,15 @@ protected override DbEntityValidationResult ValidateEntity(DbEntityEntry entityE
         {
             if (entityEntry.Entity is TUser && entityEntry.State == EntityState.Modified)
             {
-                entityEntry.Property("ConcurrencyToken").CurrentValue = Guid.NewGuid();
+                entityEntry.Property("ConcurrencyStamp").CurrentValue = Guid.NewGuid().ToString();
```
divega: FYI, if you cast the `DbEntityEntry` to a generic `DbEntityEntry<TEntity>`  for the right entity type I think you should be able to use the lambda syntax. It is fine to keep it like this though.

divega: E.g. something like this:
``` C#
        protected override DbEntityValidationResult ValidateEntity(DbEntityEntry entityEntry, IDictionary<object, object> items)
        {
            if (entityEntry.State == EntityState.Modified)
            {
                if(entityEntry.Entity is TUser) 
                {
                    entityEntry.Cast<TUser>().Property(p => p.ConcurrencyStamp).CurrentValue = Guid.NewGuid().ToString();
                }
                else if (entityEntry.Entity is TRole)
                {
                    entityEntry.Cast<TRole>().Property(p => p.ConcurrencyStamp).CurrentValue = Guid.NewGuid().ToString();
                }
            }
            return base.ValidateEntity(entityEntry, items);
        }
```

HaoK:
merged thanks! d4d105d5b529c8e1701010cb49bc115f0aa23ed0

PR #1054, opened 2016-12-14
In sample use FindByEmailAsync in ResetPassword flow

jyunderwood:
Since the password reset flow is asking for an email, use `FindByEmailAsync` instead of `FindByNameAsync` in case the developer has decoupled the two fields to use usernames.


HaoK:
Yep good catch the templates already made this change I will merge

HaoK:
Merged thanks! 9e59e79a94d995dcc2edbcdfab43889c26a7156e

PR #1055, opened 2016-12-15
Add FindByPhoneNumberAsync

RainingNight:
Used to phonenumber login.


HaoK:
Phone numbers are not guaranteed to be unique today, they are not keys

maliming:
@HaoK 
In China, the mobile phone number is the only (by the system recycling to other people except the case), most of the App and Web sites are through the user name, mobile phone number login system, but use the email login situation very few hope to consider this problem!

xyfy:
I am looking for this method too,but said to see this result

PR #1064, opened 2016-12-21
IdentityUserLogin Key mapping mistake

hikalkan:
I think UserId is missed in HasKey definition.


HaoK:
No, this would let you reuse login provider + provider id's for different users, that combination needs to be unique.

hikalkan:
OK, now understood the point. Thanks a lot for explanation.
But this will be a problem when I use the same table in a multitenant application. In a multitenant application, a person (with same facebook account) can create two different users in two different tenants and may want to login with same facebook account. But it seems this is my problem since AspNet Identity is not designed for multitenancy.

PR #1076, opened 2017-01-12
return early if possible, without getting stores

rm2k:
`UserManager` methods return early if possible, without getting stores


HaoK:
The ordering is intentional, we want the NotSupportedException before argument validation.

Issue #1060, opened 2016-12-19
Remove cancellationToken parameter in UserManager::RemovePasswordAsync


dealdiane:
Inline with the other methods in UserManager, the cancellationToken parameter in the RemovePasswordAsync method should be removed and if required, the CancellationToken property should be used instead.

HaoK:
7fbc36c56684769d3535cf1dfd1cddf88fb26343

PR #1077, opened 2017-01-13
Remove cancellation token overload

HaoK:
Fixes https://github.com/aspnet/Identity/issues/1060

Requires another baseline update :(

@divega


HaoK:
7fbc36c56684769d3535cf1dfd1cddf88fb26343

Issue #1061, opened 2016-12-20
Make password validator options get public


Saticmotion:
I want to show the current password rules on a webpage, without having to hardcode it in HTML. How do I get the current rules?

An example:

```HTML
<p>Minimum password length: @PasswordRules.RequiredLength</p>
```

maliming:
```
//or as CustomPasswordValidator
var rules = (UserManager.PasswordValidator as PasswordValidator); 
rules.RequireDigit;
rules.RequiredLength;
```

Saticmotion:
@maliming I'm using .NET Core, so I tried your second option. But it doesn't work, since Options is protected internal. I'm using the "Microsoft.AspNetCore.Identity.EntityFramewordCore" package, version 1.1.0.

maliming:
@Saticmotion 
Derive a class from UserManager

![test](https://cloud.githubusercontent.com/assets/6908465/21385672/63c59f0a-c7aa-11e6-877a-0a2ef5c06764.png)

maliming:
services.AddIdentity(.....).AddUserManager<MyUserManager>()

Saticmotion:
This works, thank you! Though it's kind of annoying I have to solve it this way.

divega:
@Saticmotion your controller can obtain the same `IOptions<IdentityOptions>` singleton from DI that the `UserManager` gets in its constructor:  https://github.com/aspnet/Identity/blob/dev/src/Microsoft.AspNetCore.Identity/UserManager.cs#L75.

divega:
BTW I am not against making the property getter public in 2.0. cc @HaoK

maliming:
DI Is a best practice.😁

Saticmotion:
@divega That's probably a better option. Thanks!!

divega:
Reopening so that we can consider making the property getter public.

HaoK:
b64a32f089f06c8c403f8bb9ee085a580c9555e2

PR #1078, opened 2017-01-13
Expose options/other services on managers

HaoK:
Fixes https://github.com/aspnet/Identity/issues/1061

To be consistent (and to return to how things were in 2.0) I restored public access to all the dependant services again (which were protected internal) except the Store which we don't want people to use directly

cc @divega @blowdart


HaoK:
:shipit: @divega says

HaoK:
b64a32f089f06c8c403f8bb9ee085a580c9555e2

Issue #1040, opened 2016-12-01
EF UserStore: Expose a new protected virtual for each Find() query


msmolka:
I'm facing issues with using default configuration of identity using entity framework core. There are restrictions on databases like max key length on mysql. So handling complex keys requires tricks and a lot overwritten code. I decided to not use complex key as primary key (also simpler change tracking later if required). But because internal code is using FindAsync method now I need to overwrite more methods in userstore just to handle single find.

My idea is use SingleOrDefaultAsync or at least make find method when complex key is used overridable. E.g. FindToken is private. Making it overridable simplifies code. At this moment I need to copy FindToken, SetTokenAsync and RemoveTokenAsync to my own code.

divega:
@msmolka this all sounds reasonable, although I am not sure that making these methods overridable is all you are going to need. Changing the definitions of the keys in the model sounds like could require more changes in the store. 

Would it be possible for you to make all the necessary changes to our source code until your provider works end-to-end and then send a PR with that?

Also, it would be interesting to know more details on why you gave up on just limiting the size of keys for MySQL.

msmolka:
About FindAsync, as far I can it was changed from SingleOrDefault to find between version 1 and 1.1. And looking into latest code (not released) it is now changed back. Having FindAsync  removing flexibility of extending base classes and add custom keys. As far as I know, from database point of view there is no benefit because query is exactly the same. 

MySQL has limitation on key column which is for 767 length in InnoDB tables. So single column for UTF8_unicode is max 255 or utf8mb4 max 191. I know I can reduce size in model builder.

However, just theoretically, if I would like to use SQL change tracking, the single column primary key is easier to maintain, than complex key, and that's why I would like to have possibility to change that behaviour. And currently I have, I can override as many as I need. My point was only to not use api that requires more work when customizing, and does not give any benefits.

divega:
> as far I can it was changed from SingleOrDefault to find between version 1 and 1.1. And looking into latest code (not released) it is now changed back

@msmolka I checked in the dev branch and it seems to be using `FindAsync()` still. If you see something else, can you please point me to it?

>  Having FindAsync removing flexibility of extending base classes and add custom keys. As far as I know, from database point of view there is no benefit because query is exactly the same.

The potential benefit of using `FindAsync()` is that it avoids a database call if the entity was already loaded in a previous call in the same request. 

[...]

PR #1083, opened 2017-01-20
Add protected virtuals for querys

HaoK:
Fixes https://github.com/aspnet/Identity/issues/1040

I didn't add new protected virtuals for any existing queries which were already exposed as public virtuals (FindByXyz, GetXyz)

@divega


HaoK:
2d157783465505bd0d36480ebe2e595543faff34

PR #1093, opened 2017-02-04
Update IdentityBuilder.cs <typeparam> tag definition

jamiewest:
Updated the <typeparam> tag used for AddUserStore<T> to match the definition used for AddRoleStore<T>.


HaoK:
b10c8ec0b8140b6b40026e10c81497140de0cb79

PR #1096, opened 2017-02-08
React to clock changes

HaoK:
Fixes CI failures caused by https://github.com/aspnet/Security/commit/fea5d5cfdc8fc9413ba7816cc108f4d893ac16e1

cc @anurse @Tratcher @divega @blowdart 

Since cookies clock is now seconds granulaurity, maybe we should change the corresponding SecurityStampValidationInterval to be total seconds instead of a Timespan which causes issues now


Tratcher:
It's always been in seconds, our rounding just want very good.

anurse:
I restarted the failed Travis macOS build, so let's wait for that to complete, but otherwise looks good to me.

PR #1099, opened 2017-02-09
Allow derived types of UserClaimsPrincipalFactory to return ClaimsPrincipals.

### LINKED ISSUE #1098: Protected CreatePrincipal Method for UserClaimsPrincipalFactory
Hello wonderful people,

I'd like to hook the creation of [`new ClaimsPrincipal(id)` in `UserClaimsPrincipalFactory:CreateAsync` at this location](https://github.com/aspnet/Identity/blob/dev/src/Microsoft.AspNetCore.Identity/UserClaimsPrincipalFactory.cs#L116) and return my own `MyAppClaimsPrincipal(id)` implementation.

The following is what I'd like to change in **UserClaimsPrincipalFactory.cs**:

[**UserClaimsPrincipalFactory.cs**](https://github.com/aspnet/Identity/blob/dev/src/Microsoft.AspNetCore.Identity/UserClaimsPrincipalFactory.cs#L116)
```
public class UserClaimsPrincipalFactory<TUser, TRole> : IUserClaimsPrincipalFactory<TUser>
    where TUser : class
    where TRole : class
{
        public virtual async Task<ClaimsPrincipal> CreateAsync(TUser user)
        {
            //... keep all microsoft's stuff
            return CreatePrincipal(id);
        }
        protected virtual ClaimsPrincipal CreatePrincipal(ClaimsIdentity id)
        {
            return new ClaimsPrincipal(id);
        }
}
```
-----
Currently, in order to wrap the `ClaimsIdentity` from `CreateAsync`, I have to call the `base.CreateAsync` like this:
```
public class MyAppClaimsPrincipalFactory : UserClaimsPrincipalFactory<User, Role>
  public override async Task<ClaimsPrincipal> CreateAsync(User user)
  {
      var basePrincipal = await base.CreateAsync(user);
      var myAppPrincipal = new MyAppClaimsPrincipal(basePrincipal.Identities.First(), this.Options);
      return myAppPrincipal;
  }
}
```
It would be great if I could simply handle the `ClaimsPrincipal` creation itself by overriding `CreatePrincipal(id)` as I've proposed above. I could simply do this instead:
```
  protected override ClaimsPrincipal CreatePrincipal(ClaimsIdentity id)
  {
      return new MyAppClaimsPrincipal(id);
  }
```
-----
Hm. Yea, it's a small nit-pick on the principal creation process but it does avoid a wasteful principal object allocation. I'd be happy to send a PR if you'd be okay with this small change. It's cool if not.

Thank you for your consideration, :v: 
Brian

:thought_balloon: :mag: ***["I want to know... what you're thinking..."](https://www.youtube.com/watch?v=ijAYN9zVnwg)***

bchavez:
Fixes #1098. Allows derived types of `UserClaimsPrincipalFactory` to return their own `ClaimsPrincipal`s.

* Wasn't sure if I should make `CreatePrincipal` **`async`** or not. Thoughts? :thinking: :thought_balloon: 

Thanks again for your consideration,
Brian

:car: :blue_car: ***["Let the good times roll..."](https://www.youtube.com/watch?v=7BDBzgHXf64)***

### DIFF
--- a/src/Microsoft.AspNetCore.Identity/UserClaimsPrincipalFactory.cs
+++ b/src/Microsoft.AspNetCore.Identity/UserClaimsPrincipalFactory.cs
@@ -113,7 +113,17 @@ namespace Microsoft.AspNetCore.Identity
             {
                 id.AddClaims(await UserManager.GetClaimsAsync(user));
             }
-            return new ClaimsPrincipal(id);
+            return await CreatePrincipalAsync(id);
+        }
+
+        /// <summary>
+        /// Creates a <see cref="ClaimsPrincipal"/> from a <see cref="ClaimsIdentity"/>.
+        /// </summary>
+        /// <param name="id">The <see cref="ClaimsIdentity"/> with claims.</param>
+        /// <returns>The <see cref="Task"/> that represents the asynchronous creation operation, containing the <see cref="ClaimsPrincipal"/> with the <see cref="ClaimsIdentity"/>.</returns>
+        protected virtual Task<ClaimsPrincipal> CreatePrincipalAsync(ClaimsIdentity id)
+        {
+            return Task.FromResult(new ClaimsPrincipal(id));
         }
     }
 }


HaoK:
In general we have made things async so CreatePrincipalAsync is probably best

bchavez:
Hi @HaoK , no problem! Thanks! Updated PR with `CreatePrincipalAsync`. :+1:

PR #1100, opened 2017-02-09
Implementing PasswordOptions.RequiredUniqueChars

### LINKED ISSUE #1097: Add a Password Option to set the number of unique chars
I think it's a pretty common idea that forcing different types of chars (numeric, special, upper, ...) is not more secure than the chars just being non-repetitive and use a long password. We can already set the minimum length for a password, but I would like to also set how many different chars it should be made of.

` PasswordOptions.RequiredUniqueChars`

Obviously it would have to be less than `PasswordOptions.RequiredLength` or equal.

c.f. https://twitter.com/codinghorror/status/829484158208471041

sebastienros:
Fixes #1097 

Took the liberty to create a branch here. Feel free to delete if the issue is not accepted, no bad feelings ;)

### DIFF
--- a/src/Microsoft.AspNetCore.Identity/IdentityErrorDescriber.cs
+++ b/src/Microsoft.AspNetCore.Identity/IdentityErrorDescriber.cs
@@ -242,6 +242,20 @@ namespace Microsoft.AspNetCore.Identity
         }
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


HaoK:
looks fine to me, any opposition to adding this @blowdart @divega ?

blowdart:
Aside from reading this as uniquecharrrrrrrrrrrrrrrrrrrs, yea, fine by me

HaoK:
:shipit: go ahead and merge this in, thanks!

PR #1119, opened 2017-03-03
Auth/Options 2.0 + Identity

HaoK:


HaoK:
@Tratcher @ajcvickers These are the initial identity changes needed for Auth 2.0, mostly switching from direct configuration of cookie options instances that live inside of identity options, to configuring them via fixed names.  MVC needs to be updated as well before Identity.Sample can be updated.

src/Microsoft.AspNetCore.Identity/BuilderExtensions.cs
```diff
@@ -1,43 +0,0 @@
...
-    /// Identity extensions for <see cref="IApplicationBuilder"/>.
-    /// </summary>
-    public static class BuilderExtensions
-    {
-        /// <summary>
-        /// Enables ASP.NET identity for the current application.
-        /// </summary>
-        /// <param name="app">The <see cref="IApplicationBuilder"/> instance this method extends.</param>
-        /// <returns>The <see cref="IApplicationBuilder"/> instance this method extends.</returns>
-        public static IApplicationBuilder UseIdentity(this IApplicationBuilder app)
```
Tratcher: This is going to come back as obsolete with the fwlink right?

src/Microsoft.AspNetCore.Identity/SecurityStampValidator.cs
```diff
@@ -18,13 +19,15 @@ public class SecurityStampValidator<TUser> : ISecurityStampValidator where TUser
...
         private readonly IdentityOptions _options;
+        private ISystemClock _clock;
 
         /// <summary>
         /// Creates a new instance of <see cref="SecurityStampValidator{TUser}"/>.
         /// </summary>
         /// <param name="options">Used to access the <see cref="IdentityOptions"/>.</param>
         /// <param name="signInManager">The <see cref="SignInManager{TUser}"/>.</param>
-        public SecurityStampValidator(IOptions<IdentityOptions> options, SignInManager<TUser> signInManager)
+        /// <param name="clock">The system clock.</param>
```
Tratcher: is this the internal one from Security?

src/Microsoft.AspNetCore.Identity/BuilderExtensions.cs
HaoK: Ah yeah good catch this PR was created before we decided to do that, I'll restore with obsolete

src/Microsoft.AspNetCore.Identity/SecurityStampValidator.cs
HaoK: Its no longer internal, since you can implement your own ISystemClock and add it to the container (what we do for tests).  But yeah its definied in the Security/Authentication package

HaoK:
14b4bea964875b605cef5a0fef7ed21ceb51aa6f

Issue #650, opened 2015-11-25
Add CreatedDate, LastLoginDate, etc... to IdentityUser


madmunsterdaddy:
These were part of former Membership Provider. Would be nice to have this in the default implementation again as it seems like a common requirement? :)

rustd:
What other properties were you looking for?

madmunsterdaddy:
Just some of those that would be common. CreatedDateTime, LastLoginDateTime, LastActivityDateTime and have signinmanager/usermanager deal with them. Like I said, seems like a pretty common scenario that would be nice to have supported OOTB.

HaoK:
Unassigned until we want to do this and move it out of backlog

RoySalisbury:
I have a need to add functionality to the Identity system that includes a LastLoginDate field (along with a host of other things).  The two places that I can see will need to be updated are in the following:

PasswordSignInAsync
- This one is pretty easy.  I can override the method in the SignInManager and add my functionality to update the necessary user property (or claim, or wherever).  You need to also override the UserManager and UserStore to add the necessary logic and backing. No original source changes required.

TwoFactorSignInAsync
- This one is trickier ...  PasswordSignIn can trigger the logic for the TwoFactorSignIn.  So you would not update the LastLoginDate until the two factor was completed.  And when the TwoFactorSignIn is called, you no longer have a reference to the User object.  It is retrieved internally to the TwoFactor call by accessing the cookies.  And to do that it needs access to some internal fields in the SignInManager that can't be used in a derived class.  So it will require source changes.

There is one place that it MAY be possible to add this functionality and that is in the SignInAsync method directly.  Both of these call that method at the end, but it is also called for some other internal stuff not related so initial SignIn (e.g., RefreshSignIn to get updated cookies). If you made the argument that anytime the cookies got updated, then that would mean a Login, then your set.  But that is a bad argument. 

ExternalLoginSignInAsync basically relies on one of the other two, so it will get handled on its own.

In my derived classes I have done most of this already.  For a TwoFactor "hack" I basically copied the "internal" code into my derived class.  However, I'm not sure that is going to be correct.  I'm no longer using the same instance of some of the objects and if any internal updates get done to them things will get messed up.  

I have to do some other major updates to the SignInManager to add some other functionality (passwords that expire and prior password history), so I think I will need to branch the original SignInManager (and SignInResult) to add it.  Once I am done I would be more than happy to provide the code for review if its helpful.  I've never done this before (branching) so its something new for me.

[...]

PR #1125, opened 2017-03-08
Enable Create/PasswordChange/SignIn date tracking + Store versioning

HaoK:
Fixes https://github.com/aspnet/Identity/issues/650

@ajcvickers we can review/discuss this in person, this actually is a good cross cutting PR that highlights how identity is designed, and how typically we add features.

Things to consider:
- Is adding a single "UserActivityStore" better than adding 3 stores (one per property), the amount of stores we have is getting a bit out of hand so my default is to try to consolidate.
- RefreshSignIn does not update the last sign in time (since this is used when cookies are renewed automatically)
- This might be a good opportunity to add some basic automated schema migration test coverage, since this introduces our first schema change from 1.0 (3 new DateTimeOffset?properties).  Can someone point me to any ef tests that might be relevant?

cc @divega @blowdart


test/Microsoft.AspNetCore.Identity.Test/SignInManagerTest.cs
```diff
@@ -543,23 +562,39 @@ public override Task<bool> ValidateAsync(string purpose, string token, UserManag
...
-        [InlineData(true, true, true, false)]
-        [InlineData(true, true, false, false)]
-        [InlineData(true, false, true, false)]
-        [InlineData(true, false, false, false)]
-        [InlineData(false, true, true, false)]
-        [InlineData(false, true, false, false)]
-        [InlineData(false, false, true, false)]
-        [InlineData(false, false, false, false)]
-        public async Task CanTwoFactorSignIn(bool isPersistent, bool supportsLockout, bool externalLogin, bool rememberClient)
+        [InlineData(true, true, true, true, true)]
```
HaoK: I'm not sure how I feel about this block of 32 permutations of true/false, but I guess its a good way to test every permutation :/  But I am not looking forward to any doubling in the future...

test/Microsoft.AspNetCore.Identity.Test/UserManagerTest.cs
```diff
@@ -694,146 +705,6 @@ public void UserManagerWillUseTokenProviderInstance()
         }
 
         [Fact]
-        public void TOTPTest()
```
HaoK: Unrelated test cleanup

HaoK:
This is also a building block for the adding password expiration support, which will be in its own PR: https://github.com/aspnet/Identity/issues/99

HaoK:
Updated this PR to version the store support, boiled down to:
- `IdentityStoreOptions.Version` in `Microsoft.Extensions.Identity.Stores` can be used by stores to version themselves.
- The default implementations of the UserStores now have a `UserStoreBaseV1` and `UserStoreBaseV2` with the V2 extending V1 and additionally implementing the new `IUserActitvityStore`.
- `AddEntityFrameworkStores[V1|V2]()` add the appropriate `UserStoreV1/V2`
- Still need a way to access `IOptions<IdentityStoreOptions>` from a DbContext @ajcvickers (I hardcoded things in the tests for now)
- This should also enable us to directly support a new `IdentityClientApplication` table to eliminate the need for a separate Identity.Service.EntityFrameworkCore package if we add it to the 2.0 schema changes, and add a new `ClientApplicationManager`/`IClientApplicationStore` to `Microsoft.Extensions.Identity.Core` @javiercn @danroth27

HaoK:
@ajcvickers this is the PR with the updated store versioning support

HaoK:
This was done via be8232d3f9c7a312314ab42dfe37047ec3b16665

vladimirlapacek:
I am not sure, how to read this. Has the change ever been merged back to `dev`? Or has it been abandoned?

HaoK:
This was reverted and is waiting for the proper time to bring back

PR #1136, opened 2017-03-20
Modify UserStore and RoleStore DbSet protection level 

yuleyule66:
The DbSet object can not be used when inheriting and rewriting certain methods.


ajcvickers:
@yuleyule66 Not necessarily saying this is a bad change, but Context is public, and Context.Set gives the same thing as these private properties. Is there something you need to that requires the properties instead of using Context.Set?

HaoK:
I also believe this has been asked for in the past, and we didn't make the change since as you mention these are just private sugar methods on top of Context.Set

yuleyule66:
All right then.

PR #1140, opened 2017-03-21
Change compilation targets

pranavkm:
* Remove net451 as a compilation target
* Update tests and samples to netcoreapp2.0
* Removed unused sample


pranavkm:
@HaoK - are the tests supposed to work on AppV?

```
 System.Data.SqlClient.SqlException : A network-related or instance-specific error occurred while establishing a connection to SQL Server. The server was not found or was not accessible. Verify that the instance name is correct and that SQL Server is configured to allow remote connections. (provider: TCP Provider, error: 25 - Connection string is not valid)
```

pranavkm:
The SqlClient exception was due to https://github.com/dotnet/corefx/issues/17166. Reverting to netcoreapp1.1 for now

ajcvickers:
@divega @bricelam @Eilon No LocalDb support seems pretty bad. Seems like this may block moving EF unless we add workarounds like using SQL Express for everything.

 dotnet/corefx#17166 (Managed SNI on windows doesn't allow LocalDB connectivity)

Eilon:
@ajcvickers I added you to a thread about this.

pranavkm:
Merged. We'll continue tracking the netcoreapp2.0 conversion via https://github.com/aspnet/Coherence/issues/166

bricelam:
Did they remove the native SNI version from the package? It's odd to me that that wouldn't resolve for the closure of packages on Windows anymore.

Eilon:
@bricelam sorry I don't know the reason (the email thread doesn't explain it). But we're getting this on the right radars to get to a resolution ASAP.

PR #1163, opened 2017-04-12
Remove dependency on Microsoft.AspNetCore.Testing from test specifications

pranavkm:


src/Microsoft.AspNetCore.Identity.Specification.Tests/Microsoft.AspNetCore.Identity.Specification.Tests.csproj
```diff
@@ -14,7 +14,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.AspNetCore.Testing" Version="$(AspNetCoreVersion)" />
```
pranavkm: We don't ship this package, so referencing this from a shipping package is problematic

PR #1171, opened 2017-04-20
UseIdentity => UseAuthentication

HaoK:
@ajcvickers @Tratcher 

Less breaky for existing apps


brockallen:
Too ambiguous of a name, IMO. I work on lots of projects that are doing authentication and that have nothing to do with ASP.NET Identity :)

HaoK:
Title is misleading, we aren't renaming UseIdentity, we are just having it call UseAuthentication basically instead of obsolete error.  Since existing apps have UseIdentity today and would be broken without UseAuthentication in 2.0, so we are just doing it automatically with an obsolete warning

Issue #651, opened 2015-11-27
Revisit identity factoring and react to Auth 2.0


prasannavl:
The SignInManager currently is too coupled to cookies. It should ideally be able to switch methods for persistence. Currently, there's no way to use the `SignInManager` and enable a simple token style authentication using either the standard `Bearer` tokens, or a custom headers which are required for pure API services.

Either that, or the `SignInManager` should be moved into a namespace that emphasizes that most of its functionality only work with cookies. Shouldn't be under a very generic purview.

alexsandro-xpt:
LOL, it's fun.... at my knowledge, it's not make any sense... :full_moon_with_face:

rustd:
So far Identity is based on cookies only. We can consider this in a future release.

alexsandro-xpt:
There a substitute for `Identity` to work with `Bearer` tokens?

brockallen:
> So far Identity is based on cookies only. We can consider this in a future release.

Sorry Pranav, but this concept makes no sense. Cookies authenticate users in a browser and are unrelated to ASP.NET Identity. An identity management library (such as ASP.NET Identity) is what stores and validates users credentials. It's really independent of the consuming application. These two "things" are separate because you don't have to use them together. Conflating cookie authentication and identity management is not helpful.

I suspect it was easier to just put the SignInManager into the same assembly as the UserManager, and thus say "we are based on cookies", but it's a disingenuous answer. The part that helps with issuing cookies should be in its own assembly. The core of ASP.NET Identity should not have a dependency on HTTP.

The separation will be important once you consider other types of authentication, such as OAuth2  (and not the OAuth2 authentication middleware that's in ASP.NET 5). 

Related: the brute-force code in the SignInManager should be in the core and separate from the code that manages the cookies. You will want those features from the OAuth2 endpoints.

HaoK:
FWIW these were split apart in 2.0

HaoK:
Moving milestone for consideration

joshgarwood:
Is this still not finished?  Which is to say, are we not able to use signInManager if we're using JwtAuthentication?  I'm using .Net Core and jwt tokens, but would also like to leverage signInManager... is this possible yet?  Or is there another recommendation you could make?

brockallen:
The other recommendation is to use a token service like IdentityServer: https://blogs.msdn.microsoft.com/webdev/2016/09/19/introducing-identityserver4-for-authentication-and-access-control-in-asp-net-core/

[...]

Issue #883, opened 2016-06-25
Should IdentityUser live in Core?


atuzovic:
The way projects in this repository are coupled now, Microsoft.AspNetCore.Identity.EntityFrameworkCore package has dependency on Microsoft.AspNetCore.Identity package.

IdentityDbContext class is defined in Microsoft.AspNetCore.Identity.EntityFrameworkCore package.

If I want to create my own clean DAL project and only have ApplicationDbContext class + POCO entities there, and I decide that my ApplicationDbContext class will inherit from IdentityDbContext in order to be able to plug in to middleware later on - logically, I have to add reference to Microsoft.AspNetCore.Identity.EntityFrameworkCore package to DAL project, which results in Microsoft.AspNetCore.Identity being pulled in to DAL project as well, which gives me dependency on other packages that are related to cookies, HTTP, ... 

What kind of architecture and dependency chain is this? I have web-related stuff in my DAL project which makes absolutely no sense.

I understand that I can copy+paste stuff from projects in this repository and arrange it the way I like it since it is all open source now, but is that really way to go?

Can you please shed some light on why is all this built this way? I could be just blind to something obvious.

Thank you!

brockallen:
Perhaps you should keep your business related entities separate from the identity management related entities.

atuzovic:
@brockallen Well, even though that makes some sense when observed from DDD perspective, it still does not explain this dependency chain. 

As a matter of fact, I've taken a close look into your MembershipReboot projects structure, and that's a great example of clear separation of concerns. I don't see your MembershipReboot.Ef project referencing any of the web assemblies, which is what I was expecting to see here as well.

Thanks for the suggestion though, appreciated. :)

brockallen:
Ok, I see your point about the web dependencies. I agree with that sentiment.

divega:
Not sure it is feasible at this point, but one way we could achieve that is by moving all interfaces and abstractions to a separate abstractions package, then have store packages depend on that. Not sure what the goal is though: if you want a DAL depend on IdentityDbContext you kind of carry a lot of baggage with that anyway.

divega:
E.g. ASP.NET Core Identity is Web specific.

brockallen:
>  E.g. ASP.NET Core Identity is Web specific.

[...]

PR #1188, opened 2017-05-08
Add Microsoft.Extensions.Identity.Core + Stores

HaoK:
First cut at splitting Identity into:

This is a big change so figured we should get the ball rolling with the PR, and can go over this in person next triage.  The biggest breaking change out of this is moving to a different namespace (Microsoft.Extensions) which requires new usings.

Microsoft.Extensions.Identity.Core contains:
- All interfaces (except SecurityStampValidator)
- UserManager and RoleManager classes (minus the CancellationToken hookup to RequestAborted) [But no SignInManager]
- Most options classes (except new SecurityStampValidatorOptions, and DataProtection/Token/PasswordHasher options)

Microsoft.AspNetCore.Identity contains:
- AddIdentity()
- All the implementations of the interfaces like PasswordHasher, PasswordValidator, etc
- Stub AspNet[User|Role]Manager implementation which only hooks up the CancellationToken
- SignInManager and SecurityStamp which are tightly bound to http/Authentication right now

Microsoft.Extensions.Identity.Stores contains:
- Basic pocos: IdentityUser etc (no navigation properties)
- User/RoleStoreBase abstract base classes which implement most of the grunt get/set store implementations using the poco, yet leave the most important CRUD operations as abstract.

Microsoft.AspNetCore.Identity.EntityFrameworkStores will contain[as of next iteration]:
- POCOs are eliminated.
- Stores derive from new base stores.

Next steps: need to look into refactoring some of the sign in functionality like lockout back into UserManager.

@divega @ajcvickers @blowdart 

Fixes: https://github.com/aspnet/Identity/issues/651 and https://github.com/aspnet/Identity/issues/883


HaoK:
@ajcvickers updated PR with what we discussed (keeping the old AspNetCore namespaces).

There is still one minor breaking change with the POCOs moving from Microsoft.AspNetCore.Identity.EntityFrameworkCore => Microsoft.AspNetCore.Identity, this isn't as big of a deal since most references are already including both, but there will be a few places in the models that will need to change the using.

HaoK:
e6f3ebf5810c793edc6f3b4ca3438194786b86a2

PinpointTownes:
@HaoK is there a particular reason why the EF Core stores have to depend on `Microsoft.AspNetCore.Identity` instead of just `Microsoft.Extensions.Identity.Core`? (`IdentityBuilder` could be moved to Core).

HaoK:
@PinpointTownes no particular reason, that's a good idea, I can move the builder to Core and switch the stores to depend on only on the extensions, so effectively we can rename it to Microsoft.Extensions.Identity.EntityFrameworkCore, something we can consider post preview2

HaoK:
I'll file an issue so we don't forget

PR #1217, opened 2017-05-19
Include the developer certificate by default

javiercn:
make it no-op when environment != development and add a gesture to explicitly remove it


src/Microsoft.AspNetCore.Identity.Service.Core/IdentityServiceBuilderExtensions.cs
```diff
@@ -56,6 +58,20 @@ public static class IdentityServiceBuilderExtensions
             return builder;
         }
 
+        public static IIdentityServiceBuilder WithoutDeveloperCertificate(this IIdentityServiceBuilder builder)
```
Tratcher: Without is a bit odd. `Remove`, or `Disable`?

Issue #1211, opened 2017-05-18
Support signing certificates from key vault


javiercn:
Support retrieving the signing certificates from key vault

PR #1218, opened 2017-05-19
Add support for keyvault

javiercn:


src/Microsoft.AspNetCore.Identity.Service.KeyVault/KeyVaultSigningCredentialsSourceOptions.cs
```diff
@@ -0,0 +1,17 @@
...
+
+using System.Net.Http;
+
+namespace Microsoft.AspNetCore.Identity.Service.KeyVault
+{
+    public class KeyVaultSigningCredentialsSourceOptions
+    {
+        public string ClientId { get; set; }
+        public string ClientSecret { get; set; }
+        public string CertificateThumbprint { get; set; }
```
pakrym: Do we want to put thumbprint here or maybe certificate object?

javiercn: @pakrym I would keep the config simple (and bindable) as much as possible, and have the component load the certificate. This version only supports ClientSecret for now

pakrym: Let's remove `CertificateThumbprint` until we start supporting it.

src/Microsoft.AspNetCore.Identity.Service.KeyVault/IdentityServiceBuilderExtensions.cs
```diff
@@ -0,0 +1,51 @@
...
+
+            builder.Services.Configure(configure);
+            builder.Services.TryAddSingleton<ISigningCredentialsSource, KeyVaultSigningCredentialSource>();
+
+            return builder;
+        }
+
+        private class DefaultSetup : ConfigureOptions<KeyVaultSigningCredentialsSourceOptions>
+        {
+            public DefaultSetup(IConfiguration configuration)
```
pakrym: I think @danroth27 had an idea how to conditionally enable default options binding. We may want to use it here.

danroth27:
Fixes #1211

PR #1237, opened 2017-05-22
Switch to new default config options support

HaoK:
@ajcvickers


HaoK:
34e7a8df279b52d88bd3b78431378c8609debebe

Issue #1205, opened 2017-05-17
Proof of key for code exchange


danroth27:
https://tools.ietf.org/html/rfc7636

This is a security feature that prevents authorization code interception attacks. It's supported by MSAL.

PR #1238, opened 2017-05-23
Add support for proof of key for code exchange

javiercn:


javiercn:
The spec is here: https://tools.ietf.org/html/rfc7636
In essence:
* The client sends a code challenge, which is the sha256 of a high entropy value on the authorize request.
* The server associates that value with the issued code
* The client sends the original high entropy value to the server on the token request.
* The server hashes that value, base64url encodes it and compares it against the challenge associate with the code.
* If the computed challenge matches the original challenge we emit tokens, otherwise we fail.

Tratcher:
Should we be adding support for this in our clients?

leastprivilege:
PKCE is for desktop/mobile native clients.

danroth27:
Fixes https://github.com/aspnet/Identity/issues/1205

PR #1240, opened 2017-05-24
Fix comment of variance time for Rfc6238

Kahbazi:
Since the `timestep` in `Rfc6238AuthenticationService` is 3 minutes, variance time for validation should be 9 minutes.


HaoK:
Thanks merged via 707be3dca0e4ea363172c7cd22d22d6bf4b8c50e

Issue #1244, opened 2017-05-26
Support concept of ClientApplication in identity core


HaoK:
Should try to roll that into the 2.0 schema and eliminate the Microsoft.AspNetCore.Identity.Service.EntityFrameworkCore

This would likely look something like:
`Microsoft.Extensions.Identity`:
+ `ClientApplicationManager`, `IClientApplicationStore`

`Microsoft.Extensions.Identity.Stores`:
+ `IdentityClientApplication`, `IdentityClientApplicationClaim`, `IdentityClientApplicationScope`, `IdentityClientApplicationRedirectUri`

brockallen:
What about scenarios where customers don't want their own token service and just want an identity database? Shouldn't all this config for the onprem B2C be in a separate DB and corresponding config system? // @danroth27

HaoK:
Its no different than roles, today you can mark them as ignored in the model and just don't use the new manager, similar to what I'm sure you are doing with roles :)

HaoK:
you can also just stick to IdentityStoreVersion1 which won't have any of the new stuff

PinpointTownes:
@HaoK what's the problem with having these entities - specific to the B2C emulator - in a separate package?

HaoK:
I'd like to address the underlying issue more directly which we've struggled with for a while anyways (how do we start adding features + schema in an optional way and let people opt out of specific schema/functionality).  If we solve that, there shouldn't be as much of an issue right?  You won't get any of the conflated stuff unless you ask for it via something like `AddEntityFrameworkStoresLatest/AddIdentity<TUser, TRole, TApplication>`.

PinpointTownes:
> If we solve that, there shouldn't be as much of an issue right?

The problem is not adding new features nor modifying the DB schema (1), it's adding the "identity service" entities directly in the Identity core components, adding totally unnecessary tables for people that don't plan to use the B2C emulator.

I know the Azure folks are pushing hard to add their "Trojan horses" everywhere in ASP.NET Core to promote their stuff (e.g ApplicationInsights, that you can't avoid 'cause it's automagically added at runtime by VS in 2.0...), but at some point, you guys will have to be honest about this aggressive marketing strategy, instead of pretending you're just adding new features to make everyone's life easier (@danroth27... if you read me...) :smile: 

(1): if you do that, I'd love to see a `ClaimValueType` property for `UserClaim` :smile:

HaoK:
[...]

PR #1262, opened 2017-06-03
[Design] Add ClientManager/Store/interfaces into Core - Iteration 1

HaoK:
Issue: https://github.com/aspnet/Identity/issues/1244

## Iteration 1
- I'm starting on these PRs now to make sure we have all the support we need in core identity to do this later. 
- We can also use these PRs to review each of the interfaces/methods as we are adding them to core.
- The main change other than moving stuff was renaming Application => Client (new placeholder name, open to other ideas).
- This iteration focuses on pulling everything in from Microsoft.AspNetCore.Identity.Service.EntityFrameworkCore and also will focus on defining the surface area of ClientManager/ClientStore/IdentityClient and associated related entities. (aka the low level building blocks)
- The next iteration will flesh out registration AddIdentity overloads, and how to opt into/out of Roles/Client functionality/schema, and absorbing the IdentityServiceSpecfication tests. (aka make it all work again)

## New types
`Microsoft.Extensions.Identity.Core`:
+ `IdentityBuilder` <= `Microsoft.AspNet.Identity`
+ `ClientManager`
+ `IClientStore`
+ `IClientClaimsPrincipalFactory`
+ `IClientClaimStore`
+ `IClientRedirectUriStore`
+ `IClientScopeStore`
+ `IClientSecretStore`
+ `IClientValidator`
+ `IQueryableClientStore`

`Microsoft.Extensions.Identity.Stores`:
+ `ClientStoreBase`
+ `IdentityClient`
+ `IdentityClientClaim`
+ `IdentityClientRedirectUri`
+ `IdentityClientScope`

`Microsoft.AspNet.Identity.EntityFrameworkCore`:
+ `ClientStore`

@javiercn take a look when you have a chance, since you basically followed the existing manager/store patterns, most of the classes were moved as is with some slight renaming + refactoring.

I did have a question about Application's ClientId/UserId.  Today Application's have both a TKey id and a clientId, but clientId's must be unique since you can look them up via the Store API.  So Is there any reason we can't just use the ClientId as the key?  Its much nicer since its always a string so it eliminates a generic.  Also are all Applications tied to a single user since there's a FindByUserId as well?

cc @ajcvickers @danroth27 @blowdart

### DIFF
--- a/src/Microsoft.Extensions.Identity.Core/Microsoft.Extensions.Identity.Core.csproj
+++ b/src/Microsoft.Extensions.Identity.Core/Microsoft.Extensions.Identity.Core.csproj
@@ -10,6 +10,10 @@
     <EnableApiCheck>false</EnableApiCheck>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <NoWarn>1701;1702;1705;1591</NoWarn>
+  </PropertyGroup>
+
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Logging" Version="$(AspNetCoreVersion)" />
     <PackageReference Include="Microsoft.Extensions.Options" Version="$(AspNetCoreVersion)" />
--- a/src/Microsoft.AspNetCore.Identity.EntityFrameworkCore/ClientStore.cs
+++ b/src/Microsoft.AspNetCore.Identity.EntityFrameworkCore/ClientStore.cs
@@ -1,0 +1,518 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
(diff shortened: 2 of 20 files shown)


danroth27:
@davidfowl

src/Microsoft.Extensions.Identity.Core/Microsoft.Extensions.Identity.Core.csproj
```diff
@@ -10,6 +10,10 @@
     <EnableApiCheck>false</EnableApiCheck>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
```
davidfowl: Huh?

HaoK: These were just to disable the no doc comment warnings for all the new classes

Issue #1253, opened 2017-05-30
Cleanup: Move more classes from Microsoft.AspNetCore.Identity into Extensions


HaoK:
IdentityBuilder, and a few other classes like validators can be moved into Extensions as well

HaoK:
@ajcvickers is this still considered cleanup?  Basically moving classes between assemblies, would not have any real customer impact as currently they always get the AspNetCore assembly, we'd just be enabling more stuff working when using the new Microsoft.Extensions.Identity...

But using a strict interpretation of 'features', this would be enabling some scenarios around that (using Microsoft.Extensions.Identity directly)...

ajcvickers:
@HaoK If it doesn't impact any customer experience and won't break apps, then it's fine to do post preview2.

HaoK:
Tracking this via https://github.com/aspnet/Identity/issues/1269

PR #1277, opened 2017-06-12
[Identity] Finish moving things into Extensions

HaoK:
Fixes https://github.com/aspnet/Identity/issues/1253

I've left the more interesting question of how to expose things via AddIdentity Core to a seperate PR.  If we tackle that now, we basically end up tackle making IdentityRoles optional as well, since it'd show up at this layer.  

I'll send out that PR as well and we can decide if its a 2.0 or 2.1...

@ajcvickers @blowdart @divega


HaoK:
The net result of this PR is we have enough in Microsoft.Extensions.Identity.Core to be able to add
`services.AddIdentityCore<TUser>.AddRoles<TRoles>()` which would get you everything except SignInManager + AspNet cancellation token hookup (which I don't think anyone will even notice)

HaoK:
Will replace with a unifed PR that does this + adds roles support and cleans everything up

PR #1281, opened 2017-06-13
Rollback optional stores + activity features

HaoK:
@ajcvickers going to merge this today once appveyer/travis finishes


kinosang:
Hello, anyone here?

Will so many years passed, Asp.Net Core still does not store user registration timestamp, and events for logging in, signing out, changing password.

It there any plan or roadmap for the user activity feature?

Issue #1269, opened 2017-06-06
Make Roles optional at the Microsoft.Extensions.Identity layer


HaoK:
High level idea is to enable something akin to:

Related to #651

I think this should be safe to do in 2.1, but we might want to consider doing something like this in RTM (as this is really finishing what we started in 651)

```C#
services.AddIdentityCore<TUser>() // UserManager related servives
   .AddRoles<TRole>() // RoleManager related services
   .AddEntityFrameworkStores<TContext>(storeOptions => { }) // All stores needed for all managers, optional schema tweaking
   .AddSignInManager() // SignInManager related stuff
```

Would be equivalent to today's (which we would still support):
```
services.AddIdentity<TUser, TRole>().AddEntityFrameworkStores<TContext>()
```

HaoK:
High level idea is to enable something akin to:

Related to https://github.com/aspnet/Identity/issues/651

I think this should be safe to do in 2.1, but we might want to consider doing something like this in RTM (as this is really finishing what we started in 651)

```
services.AddIdentityCore<TUser>() // UserManager related servives
   .AddRoles<TRole>() // RoleManager related services
   .AddEntityFrameworkStores<TContext>(storeOptions => { }) // All stores needed for all managers, optional schema tweaking
   .AddSignInManager() // SignInManager related stuff
```

Would be equivalent to today's (which we would still support):

 `services.AddIdentity<TUser, TRole>().AddEntityFrameworkStores<TContext>()`

cc @blowdart @ajcvickers @divega @danroth27

divega:
@HaoK do you always need to configure in `storeOptions` in the call to `AddEntityFrameworkStores` whether the schema for roles needs to be included or can they be inferred automatically by the fact that you called `AddRoles`?

HaoK:
Its an optional thing, I just was showing it to make it obvious that was where you will can opt into things.  

AddEntityFrameworkStores() would do something by default appropriately...

HaoK:
But its likely there will be ways to mismatch things, like calling AddRoles() and then doing storeOptions.SupportsRoles = false...  We won't be able to block all of those, but it should be easy to validate

HaoK:
Basically misconfiguring storeOptions will cause userManager apis to throw if they end up calling into store functionality that they disable

HaoK:
Also related to https://github.com/aspnet/Identity/issues/581

HaoK:
[...]

PR #1282, opened 2017-06-14
Finish split to Extensions + optional roles

HaoK:
Fixes https://github.com/aspnet/Identity/issues/1269 and https://github.com/aspnet/Identity/issues/1253

Support for:
```
services.AddIdentityCore<TUser()
   .AddRoles<TRole>()
   .AddEntityFrameworkStores<TContext>()
```

@ajcvickers @blowdart @divega 

Actually landed ok, I needed to expose a few more EFCore types

UserOnlyStore (naming?) which is a IUserStore that implements everything except the roles ones. And a few IdentityDbContext class overloads that omit the role related entities.

Existing tests all pass, and I added a few variations on running with only users enabled.  I still need to add a varation that runs with users + roles via the new builder APIs, and also a variation that runs against a DbContext that doesn't derive from identityDbContext.


src/Microsoft.AspNetCore.Identity.EntityFrameworkCore/IdentityEntityFrameworkBuilderExtensions.cs
```diff
@@ -30,32 +30,68 @@ public static IdentityBuilder AddEntityFrameworkStores<TContext>(this IdentityBu
 
         private static void AddStores(IServiceCollection services, Type userType, Type roleType, Type contextType)
```
HaoK: This is the important method to review, as its doing the mega type inference we need to do for the custom poco scenario

HaoK:
f555a26b4a554f73eea70b4b34fca823fab9a643

PR #1299, opened 2017-07-04
TFA tweaks for templates

HaoK:
Add `CountRecoveryCodesAsync` API
Change default RecoveryCode generation to be the first 8 chars of a Guid()

cc @blowdart @ajcvickers


blowdart:
As we're truncating the guids now, we need to check the range for dupes ...

HaoK:
Eh, on the unlikely chance you get a dupe, that just means when you redeem it, both get removed at once: https://github.com/aspnet/Identity/blob/dev/src/Microsoft.Extensions.Identity.Stores/UserStoreBase.cs#L1049

But sure, we can ensure their unique

PR #1332, opened 2017-07-26
Update VersionPrefix

ryanbrandenburg:


JunTaoLuo:
Why are we changing this? Is this change required since we updated other version numbers from 2.0.0 to 2.1.0?

ryanbrandenburg:
Yes, I was told that when we moved to `2.1.0` we should have moved the `0.3.0` versions to `0.4.0`

JunTaoLuo:
I suppose there are more than just this one instance? https://github.com/aspnet/Diagnostics/blob/ec2faf55947940350e124829c9922771bdfc7793/src/Microsoft.AspNetCore.Diagnostics.Elm/Microsoft.AspNetCore.Diagnostics.Elm.csproj#L7?

JunTaoLuo:
Ah I see you already have a PR open there too.

Issue #1342, opened 2017-08-02
Revisit and tidy up TOTP flow


blowdart:
- [ ] Ability to back out of rolling keys
- [ ] Support for multiple tokens (Yubikey etc)
- [ ] Ability to turn off individual 2fa providers

HaoK:
Some of these are template issues.  But we'll need to change identity to have the 2fa cookies also sign out everywhere when the security stamp changes which it doesn't do today...

HaoK:
@blowdart stumbled on this old bug, are there any of these that we want to still take a look at for 2.1?  I have a PR for voiding TFA cookies when security stamp rolls, so I'll work on getting that in soon

blowdart:
Backout and forget this machine as well.

HaoK:
What do you want the UI/flow to look like for backout and forget?

blowdart:
A cancel button on the scan this image page for backout. An extra page mid logout for forget this machine?

HaoK:
I'm not sure I understand the desired flow for the back out...

We are talking about the Reset authenticator app flow right?

So today, you click Reset App: there's a button that says Reset key which if you click immediately nukes your app.  or are you referring to a different rolling of keys?

HaoK:
Or are you talking about rolling the recovery keys?

HaoK:
Forget this machine added via ae127558b55d227e3b496db52db8412f4e821e30

HaoK:
Rolling the authenticator key now also rolls the security stamp, which causes sign out everywhere.  Everything we wanted to do for 2.1 is now done.  Moving the issue to 2.2 for the tracking the nice to haves

PR #1351, opened 2017-08-08
TFA cookies now validate security stamp

HaoK:
Part of https://github.com/aspnet/Identity/issues/1342

When security stamp is updated, the TFA remember me cookie will be invalidated

@ajcvickers @Tratcher @blowdart


src/Microsoft.AspNetCore.Identity/ITwoFactorSecurityStampValidator.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Identity
+{
+    /// <summary>
+    /// Used to validate the two factor cookies security stamps.
+    /// </summary>
+    public interface ITwoFactorSecurityStampValidator : ISecurityStampValidator
```
HaoK: ICookieValidator and have both of them derive from that instead.  Assuming that's not a breaking change

src/UI/Areas/Identity/Pages/Account/Manage/TwoFactorAuthentication.cshtml.cs
```diff
@@ -24,7 +24,10 @@ public abstract class TwoFactorAuthenticationModel : PageModel
         [TempData]
         public string StatusMessage { get; set; }
 
-        public virtual Task<IActionResult> OnGet() => throw new NotImplementedException();
+        public virtual Task<IActionResult> OnGetAsync() => throw new NotImplementedException();
```
HaoK: @javiercn can you review the fix for the TFA page that was missing OnPost, is this the correct fix?

javiercn: Yep

Issue #1370, opened 2017-08-17
`UnauthorizedAccessException`  thrown when checking if a user is SignedIn while using IIS Windows Authentication and 


mxa0079:
When using EF Identity along with IIS Windows Authentication, the [SignInManager](https://github.com/aspnet/Identity/blob/eb3ff7fc32dbfff65a1ba6dfdca16487e0f6fc41/src/Microsoft.AspNetCore.Identity/SignInManager.cs) throws a `UnauthorizedAccessException` when checking if the user is signed in (line 120).

The root cause of the issue [is described here](https://github.com/aspnet/IISIntegration/issues/231) (in the comment by ph1ll) -- the call to [`i.AuthenticationType`](https://github.com/aspnet/Identity/blob/eb3ff7fc32dbfff65a1ba6dfdca16487e0f6fc41/src/Microsoft.AspNetCore.Identity/SignInManager.cs#L127) makes a win32 call to LsaGetLogonSessionData, which requires the caller to be the session owner or the local admin; something quite unlikely in a web application inside an enterprise environment.

HaoK:
If you are using windows auth, you can probably just replace the call to SignInManager with something like User.Identity.IsAuthenticated

blowdart:
@brentschmaltz do you still own WindowsIdentity? This seems to be a problem with WindowsIdentity.

brentschmaltz:
@blowdart We don't 'own' it, but we have done some work in WindowsIdentity.  From time to time we help out with issues.  It looks like a fix is needed.

blowdart:
@brentschmaltz Who should be tagged to do the work?

brentschmaltz:
@blowdart you can tag me. Ill investigate.

blowdart:
Consider yourself tagged then :)

blowdart:
Closing as external

PR #1377, opened 2017-08-19
Fixes issue #1370

mxa0079:
As described in the issue #1370 -- the method `SignInManager.IsSignedIn(ClaimsPrincipal)` throws when the request has a Windows Identity attached to it.

The root cause for this is described in Ph1ll's comment on [this issue](https://github.com/aspnet/IISIntegration/issues/231). 

I believe this proposed solution keeps the expected behavior of the method, while fixing a relevant issue for applications on the enterprise.

This issue is blocking adoption of ASP.NET Core as it is preventing the deployment to production of an application in a Fortune 500 pharmaceutical company.

### DIFF
--- a/src/Microsoft.AspNetCore.Identity/SignInManager.cs
+++ b/src/Microsoft.AspNetCore.Identity/SignInManager.cs
@@ -124,7 +124,15 @@ namespace Microsoft.AspNetCore.Identity
                 throw new ArgumentNullException(nameof(principal));
             }
             return principal?.Identities != null &&
-                principal.Identities.Any(i => i.AuthenticationType == IdentityConstants.ApplicationScheme);
+                principal.Identities.Any((i) => 
+                {
+                    try{
+                      return i.AuthenticationType == IdentityConstants.ApplicationScheme;                        
+                    }
+                    catch(UnauthorizedAccessException e){
+                        return false;
+                    }
+                });
         }
 
         /// <summary>
--- a/src/Microsoft.AspNetCore.Identity/SignInManager.cs
+++ b/src/Microsoft.AspNetCore.Identity/SignInManager.cs
@@ -129,7 +129,7 @@ namespace Microsoft.AspNetCore.Identity
                     try{
                       return i.AuthenticationType == IdentityConstants.ApplicationScheme;                        
                     }
-                    catch(UnauthorizedAccessException e){
+                    catch(UnauthorizedAccessException){
                         return false;
                     }
                 });


mxa0079:
@HaoK - we appreciate your attention and feedback on this issue. We really need this fix.

HaoK:
If you are using windows auth, you can probably just replace the call to SignInManager with something like User.Identity.IsAuthenticated instead of using this method if its causing exceptions in this case...  This method is only really supposed to be used when you are using cookie authentication, not windows auth...

mxa0079:
@HaoK  We use a combination of cooke and windows authentication, hence the need to call the method.

mxa0079:
@HaoK your comments are appreciated. The PR is very simple, I'd like to understand what is the hesitation to approve it.

Thanks

HaoK:
We can certainly consider this for 2.1, but the benefits of a fix like this are fairly limited, you could just wrap your calls to `SignInManager.IsSignedIn(ClaimsPrincipal)` in your app with this try/catch today.  That's a pretty simple workaround if this is blocking you.

HaoK:
Or you can just check if User.Identity.IsAuthenticated first, and only call SignInManager.IsSignedIn after that is false, since that would mean its safe to call right?

mxa0079:
@HaoK Trust me, we have tried. 

Why would using a try catch wont work? Because a `ClaimsPrincipal` can have multiple identities. If we wrap the call to IsSignedIn with try catch then the enumeration through the identities won't complete and hence we wont know if the user is logged in.

For the second one: I will try it this evening.

mxa0079:
@HaoK  "Or you can just check if User.Identity.IsAuthenticated first, and only call SignInManager.IsSignedIn after that is false, since that would mean its safe to call right?" This won't work because User, as a ClaimsPrincipal, has multiple Identities linked to it. When IIS sends us the Windows Authentication Identity, this is marked as the main Identity (`User.Identity`) and hence the call to `User.Identity.IsAuthenticated` will always return true (since IIS has authenticated that user) but that does not mean that he/she should access the application since we have yet to make sure that identity is registered in the EF Identity database.

HaoK:
SignInManager.IsSignedIn is mostly a sugar method meant to make the normal case easier.  Given that you have a more complex auth setup with Windows Auth involved, I'd recommend just writing custom method to do this functionality for you.

As you've seen the method is just a more compact way of writing:
```
return principal?.Identities != null &&
                 principal.Identities.Any(i => i.AuthenticationType == IdentityConstants.ApplicationScheme);
```

This method had more of a use when the ApplicationScheme was configurable, but now that's been removed in 2.0, its basically just a static extension method now...

PR #1391, opened 2017-08-23
Added RavenDB identity provider.

JudahGabriel:


HaoK:
@blowdart ok with this change?

blowdart:
Sure

PR #1398, opened 2017-08-25
Bump version to 1.0.4

natemcmaster:


src/Microsoft.AspNet.Identity.AspNetCoreCompat/project.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "0.1.3",
```
JunTaoLuo: One thing to check here is whether we want to ship this "pre-release" package. I remember I keep getting different answers when I ask this question.

JunTaoLuo: I think the last time I checked, we decided to cascade the versions but not push any new packages built to nuget, let's verify that that's still the expectation.

natemcmaster: Right. For now Microsoft.AspNet.Identity.AspNetCoreCompat is marked as "noship" in the patch system.

JunTaoLuo: Which reminds me, let's check what's in the patch packages.csv, we had issues with it in the past.

PR #1399, opened 2017-08-27
Fix for couple of obvious typos

DamirAinullin:
Fix for couple of obvious typos.


DamirAinullin:
Do you have any objections related with these changes?

Issue #1382, opened 2017-08-21
Method incorrect declaration


ajcvickers:
Moving from https://github.com/aspnet/EntityFrameworkCore/issues/9500 by @roonikz
When you want to use IdentityBuilder to setup Identity services for your application (e.g. AddUserStore<>() AddRoleStore<>() AddPasswordValidator<>() and so on). When you have looked at these methods' declarations, you will see: 

```c#
    public virtual IdentityBuilder AddUserStore<TUser>() where TUser : class
    {
      return this.AddScoped(typeof (IUserStore<>).MakeGenericType(this.UserType), typeof (TUser));
    }
```

Looking at _**TUser**_ type, you are thinking about passing User type as generic parameter. Using Autofac DI container integration I have got this ArgumentException:

```
ArgumentException: The type 'CMS.Domain.User' is not assignable to service 'Microsoft.AspNetCore.Identity.IUserStore`1[[CMS.Domain.User, CMS.Domain, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]]'.
```
Then I decompile AddScoped() method, and here what I see:

```c#
    private IdentityBuilder AddScoped(Type serviceType, Type concreteType)
    {
      this.Services.AddScoped(serviceType, concreteType);
      return this;
    }
```
AddScoped method second argument is concreteType! As I think, it would be better to replace **`Add*WhateverIdentityService*<TUser> where TUser: class`**  by **`Add*WhateverIdentityService*<TWhateverIdentityService> where TWhateverIdentityService: class`** to avoid disagreements. Thank you.

### Further technical details
EF Core version: EF Core 2.0
Database Provider: Microsoft.EntityFrameworkCore.SqlServer
Operating system: Windows 10 
IDE: Visual Studio 2017

rooonikz:
@HaoK I think it still has some stuff to be done. As i mentioned before, this renamings should be applied to all Identity services. As i see, AddPasswordValidator, AddRoleValidator and AddUserValidator generic parameters declarations are invaid. Thx!

HaoK:
Ah right, I'll rename the others today

HaoK:
Fixed the rest bc75b1a0e73d782fbe221fb6f12855d1a337a5b1

Issue #1379, opened 2017-08-21
IdentityBuilder methods not compatible with 3rd party service providers


nphmuller:
I've been scratching my head on this for a couple of hours during my migration from 1.x to 2.0.

The IdentityBuilder class uses a couple of calls that add services to the service provider of which I don't really understand how they work, but they work with the default service provider and don't with 3rd party service providers (at least Autofac).

For example: https://github.com/aspnet/Identity/blob/8cbda5a23dd594116fcd2f60c486d15fa50051af/src/Microsoft.Extensions.Identity.Core/IdentityBuilder.cs#L109 and https://github.com/aspnet/Identity/blob/8cbda5a23dd594116fcd2f60c486d15fa50051af/src/Microsoft.Extensions.Identity.Core/IdentityBuilder.cs#L200

They add a user/role store and use the user/role as the concrete type? How does the service provider know which store to actually use? TUserStore or TRoleStore seemed way more logical and convenient.

Anyway, when I try to build the service provider after calling `.AddRoles<MyRoleType>()` I get this exception (again, with Autofac):
```
System.ArgumentException: The type 'MyRoleType' is not assignable to service 'Microsoft.AspNetCore.Identity.IRoleStore`1[[MyRoleType, MyApi, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null]]'.
   at Autofac.Builder.RegistrationBuilder.CreateRegistration(Guid id, RegistrationData data, IInstanceActivator activator, IEnumerable`1 services, IComponentRegistration target)
   at Autofac.Builder.RegistrationBuilder.CreateRegistration[TLimit,TActivatorData,TSingleRegistrationStyle](IRegistrationBuilder`3 builder)
   at Autofac.Builder.RegistrationBuilder.RegisterSingleComponent[TLimit,TActivatorData,TSingleRegistrationStyle](IComponentRegistry cr, IRegistrationBuilder`3 builder)
   at Autofac.ContainerBuilder.Build(IComponentRegistry componentRegistry, Boolean excludeDefaultModules)
   at Autofac.ContainerBuilder.Build(ContainerBuildOptions options)
   at MyApi.Startup.ConfigureServices(IServiceCollection services)
```

nphmuller:
The workaround:

Instead of this:
``` c#
services.AddIdentityCore<User>(options => { ... })
    .AddRoles<Role>()
    .AddRoleStore<RoleStore>()
```
do this:
``` c#
[...]

PR #1406, opened 2017-08-31
Fix params name and AddRoles

HaoK:
AddRoles should not be adding a store, that's done by the store
implementation, it should only be registering the role services.  Also
fixed the TUser/TRole generic parameters names since they were actually
the store types

Fixes https://github.com/aspnet/Identity/issues/1382
https://github.com/aspnet/Identity/issues/1379

@ajcvickers


HaoK:
reviewed :shipit:

PR #1419, opened 2017-09-09
use WebHost.CreateDefaultBuilder in IdentitySample.Mvc

tugberkugurlu:
so that it picks up the logging config from the `appsettings.json`

### DIFF
--- a/samples/IdentitySample.Mvc/IdentitySample.Mvc.csproj
+++ b/samples/IdentitySample.Mvc/IdentitySample.Mvc.csproj
@@ -12,6 +12,7 @@
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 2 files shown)


HaoK:
Thanks!

HaoK:
Had to back this out, this introduces cycles in the build dependencies, identity can't depend on AspNetCore

tugberkugurlu:
> Had to back this out, this introduces cycles in the build dependencies, identity can't depend on AspNetCore

That's weird since it's only the sample.

HaoK:
Its a repo circular dependency, AspnetCore is basically last and depends on all the other repos, so nothing in identity can depend on that package

PR #1420, opened 2017-09-10
Rename IUserAuthenticatorInfoStore.cs to IUserAuthenticatorKeyStore.cs

tugberkugurlu:
so that it matches the name of the class and makes it easier to find


HaoK:
Good catch thanks!

PR #1431, opened 2017-09-15
[2.0.1] Update version.props and add list of dependendencies that are also being updated

natemcmaster:
CI will fail until we actually have a candidate build of the EF dependencies.

:watch: will wait until those are available to merge

### DIFF
--- a/src/Microsoft.AspNet.Identity.AspNetCoreCompat/Microsoft.AspNet.Identity.AspNetCoreCompat.csproj
+++ b/src/Microsoft.AspNet.Identity.AspNetCoreCompat/Microsoft.AspNet.Identity.AspNetCoreCompat.csproj
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <Description>A compatibility layer for sharing identity databases between Microsoft.AspNet.Identity.EntityFramework and Microsoft.AspNetCore.Identity.EntityFrameworkCore.</Description>
-    <VersionPrefix>0.3.0</VersionPrefix>
+    <VersionPrefix Condition="'$(ExperimentalProjectVersionPrefix)' != ''">$(ExperimentalProjectVersionPrefix)</VersionPrefix>
     <TargetFramework>net461</TargetFramework>
     <NoWarn>$(NoWarn);CS1591</NoWarn>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
--- a/version.props
+++ b/version.props
@@ -1,6 +1,9 @@
-<!-- This file may be overwritten by automation. Only values allowed here are VersionPrefix and VersionSuffix.  -->
 <Project>
   <PropertyGroup>
-    <VersionPrefix>2.0.0</VersionPrefix>
+    <VersionPrefix>2.0.1</VersionPrefix>
+    <VersionSuffix>rtm</VersionSuffix>
+    <PackageVersion Condition="'$(IsFinalBuild)' == 'true' AND '$(VersionSuffix)' == 'rtm' ">$(VersionPrefix)</PackageVersion>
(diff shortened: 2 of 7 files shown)


src/Microsoft.AspNet.Identity.AspNetCoreCompat/Microsoft.AspNet.Identity.AspNetCoreCompat.csproj
```diff
@@ -4,7 +4,7 @@
 
   <PropertyGroup>
     <Description>A compatibility layer for sharing identity databases between Microsoft.AspNet.Identity.EntityFramework and Microsoft.AspNetCore.Identity.EntityFrameworkCore.</Description>
-    <VersionPrefix>0.3.0</VersionPrefix>
+    <VersionPrefix Condition="'$(ExperimentalProjectVersionPrefix)' != ''">$(ExperimentalProjectVersionPrefix)</VersionPrefix>
```
JunTaoLuo: Why the experimental project version prefix? I'd assume that we would never decide during a patch update to designate a pre-release package as rtm. In fact, in the past, we do not ship patch updates for pre-release packages at all.

natemcmaster: Yeah, idk. @Tratcher ? We shipped 0.3.0, to nuget.org so I assume we should also ship 0.3.1.

version.props
```diff
@@ -1,6 +1,10 @@
-<!-- This file may be overwritten by automation. Only values allowed here are VersionPrefix and VersionSuffix.  -->
 <Project>
   <PropertyGroup>
-    <VersionPrefix>2.0.0</VersionPrefix>
+    <VersionPrefix>2.0.1</VersionPrefix>
+    <ExperimentalProjectVersionPrefix>0.3.1</ExperimentalProjectVersionPrefix>
```
natemcmaster: just FYI - I've found it's too easy to forget to update a VersionPrefix set in csproj directly. I'm hoping putting all versions in version.props will help us when we have multiple version prefixes coming from one repo.

src/Microsoft.AspNet.Identity.AspNetCoreCompat/Microsoft.AspNet.Identity.AspNetCoreCompat.csproj
JunTaoLuo: cc @Eilon would we want to ship updates for pre-release packages in 2.0.x patches? We decided not to do this in 1.x patches.

version.props
JunTaoLuo: Ah that's why you moved it here. 👍

src/Microsoft.AspNet.Identity.AspNetCoreCompat/Microsoft.AspNet.Identity.AspNetCoreCompat.csproj
Tratcher: @haok

HaoK: I don't think we need to update this package, it never was officially released and is more of a prototype/sample anyways.

Eilon: Yeah in the past we didn't ship patches of these, so no point starting now. Less stuff to deal with.

natemcmaster: Ok. I'll mark this package as "noship" in the packages.csv list. That is what ultimately determines what will ship.

PR #1432, opened 2017-09-16
XML doc comments changes

### LINKED ISSUE #1426: Move Options Defaults to Summary instead of Remarks
Just wanted to open a issue about the Defaults for things like `PasswordOptions` defaults being in the [Remarks tag of the xml doc comments instead of the Summary tag](https://github.com/scottsauber/Identity/blob/bc75b1a0e73d782fbe221fb6f12855d1a337a5b1/src/Microsoft.Extensions.Identity.Core/PasswordOptions.cs#L14).

I think for discoverability from a Developer point of view, it would be better if the defaults were in the Summary.  That way we see the default directly in the intellisense of VS while we type without having to hit F12 to dive into the remarks.

I was going to submit a PR to add the defaults to `UserOptions` which currently doesn't have any defaults in the Remarks or Summary, but then this just hit me.

Happy to submit a PR to switch this if there's interest.  Thoughts?

scottsauber:
- Move defaults from remarks to summary for better intellisense.
- Add summary to describe LockoutOptions AllowedForNewUsers.
- Added defaults to xml docs where missing.

Fixes #1426

### DIFF
--- a/src/Microsoft.AspNetCore.Identity/DataProtectionTokenProviderOptions.cs
+++ b/src/Microsoft.AspNetCore.Identity/DataProtectionTokenProviderOptions.cs
@@ -11,7 +11,7 @@ namespace Microsoft.AspNetCore.Identity
     public class DataProtectionTokenProviderOptions
     {
         /// <summary>
-        /// Gets or sets the name of the <see cref="DataProtectorTokenProvider{TUser}"/>.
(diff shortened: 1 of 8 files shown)


HaoK:
Thanks!

scottsauber:
Glad to help, thanks for accepting!

Issue #1189, opened 2017-05-09
Allowing duplicate emails results in email not being validated


CMircea:
https://github.com/aspnet/Identity/blob/dev/src/Microsoft.AspNetCore.Identity/UserValidator.cs#L51'

ValidateEmail should be called regardless of RequireUniqueEmail. There should be a separate method that checks if the email is unique and only call that if necessary, but always validate the email address.

However, the better design is to make unique email validation a requirement for the UserStore to fulfill (the same for usernames), because in this situation there is a race condition. The UserStore must be able to ensure the uniqueness of emails/usernames if requested (in SQL it's an unique index, in some document DB's you store a special document with the email as the document key).

HaoK:
Actually, we can't do this, since this would effectively require an email, when today email is only required when this flag is on.

HaoK:
Null/empty emails are allowed when this flag is off, which is intended right now

CMircea:
Well... the ValidateEmail method can check if the email is required and not throw an error if the email is missing and it's optional.

HaoK:
Oh I see, its only the valid email check that you want as opposed to that its non empty and valid

CMircea:
Yep. That is skipped, but it shouldn't be.

CMircea:
Might want to add the bug tag back, it is a bug if the email isn't validated in this case.

HaoK:
Note to triage: @blowdart @ajcvickers this is probably a breaking change since with this bug fix, existing users with invalid emails that used to be able to function, would start failing to update/save until their emails were made valid.

HaoK:
Might need a new flag to control this, an `alwaysValidateEmail` instead, so existing apps wouldn't be broken on upgrade.

HaoK:
Discussed in triage, we decided it wasn't worth doing this, there is always the workaround of plugging in your own custom UserValidator to control the validation entirely

PR #1435, opened 2017-09-18
Always validate email

HaoK:
Fixes https://github.com/aspnet/Identity/issues/1189

@ajcvickers


HaoK:
Actually I don't htink we can do this, since this would require emails

HaoK:
Tweaked, so we always validate emails that are provided, but only throw when they are missing if they are required

Issue #1409, opened 2017-09-01
Refactor AddIdentity splitting out Authentication from core identity


brockallen:
ASP.NET Identity is a library, and should not have such an opinion and side-effect on a hosting application. By setting the default signin scheme to external, you're making assumptions about the hosting application. If the hosting app wants to use external cookies, then it should do that configuration itself. 

I understand why it was added in here -- it makes the templates easier. But it's a bad library design choice (especially since the behavior is not relied upon anywhere in the ASP.NET Identity library itself). You might have an app that uses some NuGet library with a transitive dependency on ASP.NET Identity and it'd be incredibly confusing to have this behavioral change in the hosting app as a result of just adding a library.

HaoK:
Yup, the new refactored surface area won't set things up this way, but we had to preserve existing behavior for `AddIdentity<TUser, TRole>`

HaoK:
In 2.1, you should be able to switch to something like:

```
services.AddIdentityCore<TUser>()
   .AddRoles<TRole>()
   .AddEntityFrameworkStores<TContext>()
   .AddCookieAuthentication() // Brings in sign in manger cookies, and you'd be able to control the default schemes here somehow
```

davidfowl:
I agree m, Hao an I spoke about this earlier and threw around some ideas. I don't think identity should set the default scheme at all and should have extensions methods on both IServiceCollection and Authentication builder.

HaoK:
Yup, we'll use this bug to track that work

brockallen:
The name `DefaultSignInScheme` is really the wrong one given what I think the intent of this feature is. I think your intent is that this is meant for the external authentication middleware, right? This property should be renamed to be `DefaultExternalSignInScheme`. Perhaps the optional `scheme` param on all the `HttpContext` APIs should be removed and make it explicit. Given these global defaults and the implicit behavior, it seems brittle.

davidfowl:
>  I think your intent is that this is meant for the external authentication middleware, right?

No, it maps to any call to SignInAsync without parameters.

> Perhaps the optional scheme param on all the HttpContext APIs should be removed and make it explicit. Given these global defaults and the implicit behavior, it seems brittle.

In what version? 3.0?

HaoK:
It's already possible to specify schemes explicitly everywhere the only thing the defaults are mandatory for is the UseAuthentication() middleware to determine what context.User gets set to. 

[...]

PR #1436, opened 2017-09-18
[Prototype] Add CookieAuthentication

HaoK:
Work for https://github.com/aspnet/Identity/issues/1409


HaoK:
Replaced by https://github.com/aspnet/Identity/pull/1442

PR #1442, opened 2017-09-26
Split Auth into AddIdentityCookies

HaoK:
Work for https://github.com/aspnet/Identity/issues/1409

New startup code for refactored identity:

```C#            
            services.AddIdentityCore<ApplicationUser>()
                .AddRoles<IdentityRole>()
                .AddSignInManager()
                .AddDefaultTokenProviders();

            services.AddAuthentication(o =>
            {
                o.DefaultScheme = IdentityConstants.ApplicationScheme;
                o.DefaultSignInScheme = IdentityConstants.ExternalScheme;
            })
            .AddIdentityCookies(o => 
            {
                // This would be empty typically, just demonstrating the new API per cookie
                o.ApplicationCookie.Configure(c => c.Cookie.Name = "MainCookie");
                o.ExternalCookie.Configure(c => c.Cookie.Name = "SocialCookie");
                o.TwoFactorRememberMeCookie.Configure(c => c.Cookie.Name = "RememberBrowser");
                o.TwoFactorUserIdCookie.Configure(c => c.Cookie.Name = "2falogin");
            });
```

Thoughts @ajcvickers @davidfowl @blowdart ?


davidfowl:
````C#
.AddDefaultCookies() // Adds and configures default cookie auth
```

I don't like this method at all. Feels like it should be on AddAuthentication directly. Always force the user to make those 2 calls. 

Are the identity cookies usable without identity? 

```C#
services.AddAuthentication()
         .AddApplicationCookie(o => { })
         .AddExternalCookie(o => { })
         .AddTwoFactorRememberMeCookie(o => { })
         .AddTwoFactorUserIdCookie(o => { })
```

VS

```C#
services.AddAuthentication()
         .AddIdentity(o => { 
              o.ExternalCookie.Enabled = true;
              o.TwoFactorRememberMeCookie.Enabled = true;
              o.TwoFactorUserIdCookie.Enabled = true;
          })
```

HaoK:
The identity cookies are just basically like normal cookies, with the only main difference in the application cookie, is the security stamp validation that's used for sign out everywhere.  So if you tried to sign in an arbitrary principal into the Identity.ApplicationCookie, it would likely be rejected unless they put the security stamp into the right claim.

So you thinking of something like this, where `AddAuthentication` is always required:

```
services.AddIdentityCore<ApplicationUser>()
                .AddRoles<IdentityRole>()
                .AddSignIn() // Adds SignInManager (works against IdentityConstant schemes)
                .AddDefaultTokenProviders();

services.AddAuthentication().AddIdentityCookies(o => { // They all default to true?
              o.ExternalCookie.Enabled = false;
              o.TwoFactorRememberMeCookie.Enabled = false;
              o.TwoFactorUserIdCookie.Enabled = false;
          })
```

davidfowl:
> So you are thinking of something like this? Where AddAuthentication is always required:

Yep! They would all default to true as well. How do you configure each of the cookies today? Can you (and would you want to)?


I have more questions about this code block:

```C#
services.AddIdentityCore<ApplicationUser>()
                .AddRoles<IdentityRole>()
                .AddSignIn() // Adds SignInManager (works against IdentityConstant schemes)
                .AddDefaultTokenProviders();
```

- Is ApplicationUser a poco? It can be used without EF right? What does that look like?
- Why is `AddDefaultTokenProviders` an addon and not on by default? Is that because this low level API adds nothing by default?
- How do things fail if you forget one of those essential calls?

HaoK:
> How do you configure each of the cookies today? Can you (and would you want to)?

You can currently configure the cookies today via their names using named options, we have sugar for the main application/external cookies via `ConfigureApplication/ExternalCookie`, typically people tweak things like the expire duration.  Or if they want to disable the security stamp validator.

> Is ApplicationUser a poco? It can be used without EF right? What does that look like?

Yeah that's just `TUser`, There's nothing special about it at all anymore, to use the built in EF stores, they have to extend from `IdentityUser`.

To use a custom store, you just add your own store implementations to services

```C#
services.AddScoped<IUserStore<ApplicationUser>. CustomStore>()
            .AddIdentityCore<ApplicationUser>();
```


> Why is AddDefaultTokenProviders an addon and not on by default? Is that because this low level API adds nothing by default?

These add a bunch of things that the token providers for TFA/password recovery/email confirmation that the templates use.  They aren't critical to identity at all.  We can roll them into the core if we want, they were added back in Identity 2.0 which is why they weren't part of the original core stuff.

> How do things fail if you forget one of those essential calls?

The essential are `AddIdentityCore` as there will be no `UserManager` at all without that one, and it adds most of the other core identity services as well, so that would most likely surface with a controller failing to be constructed.  `AddRoles` basically just adds `RoleManager` and isn't essential unless you are using `RoleManager`.  `AddSignIn` is similar for `SignInManager`.  Missing `AddEntityFrameworkStores` would also cause all of the managers to fail due to missing Store dependencies similarly.

davidfowl:
> You can currently configure the cookies today via their names using named options, we have sugar for the main application/external cookies via ConfigureApplication/ExternalCookie, typically people tweak things like the expire duration. Or if they want to disable the security stamp validator.

That will be much cleaner with this new model if those are properties on the options. Those other top level methods are a bit gross.

Can you update the sample with the default cookie?

HaoK:
Since you can't add a scheme that's already been added, so unless they go with the existing gross top level methods, or the undiscoverable `Configure<CookieAuthenticationOptions>(IdentityConstants.ApplicationScheme, o => o.SlidingExpiration = false)`

We'd have to expose an alternate API to add + configure the cookies off AuthenticationBuilder, which would be called instead of the `AddIdentityCookies`
```C#
     services.AddAuthentication()
         .AddIdentityApplicationCookie(o => { })
         .AddIdentityExternalCookie(o => { })
         .AddIdentityTwoFactorRememberMeCookie(o => { })
         .AddIdentityTwoFactorUserIdCookie(o => { })
```

HaoK:
So hence my original naming was `AddDefaultIdentityCookies` which just calls the other 4 `AddIdentityXyzCookie()`

HaoK:
Maybe the `AddSignIn()` call should be moved from the normal `IdentityBuilder` from `AddIdentityCore` to a new `IdentityAuthenticationBuilder` which can than hold the `AddXyzCookies()` so they are also scoped to identity.

So something like:
```C#
// Non authentication identity stuff
services.AddIdentityCore<ApplicationUser>()
                .AddRoles<IdentityRole>()

// Configure auth for identity
services.AddAuthentication()
    .AddIdentitySignIn() // Adds SignInManager, sets authentication policy to "Identity"
    .AddDefaultCookies() // Adds default identity cookies
    // OR instead they can pick which cookies they want and/or configure them
    .AddApplicationCookie(o => o.CookieName = "app")
    .AddExternalCookie(o => o.CookieName = "ext")
    .AddTwoFactorUserIdCookie(o => o.CookieName = "tfa1")
    .AddTwoFactorRememberMeCookie(o => o.CookieName = "tfa2")
```

HaoK:
Moved `ISecurityStampValidator` and related classes into `SignIn.Cookies`, main Identity package now depends on that.

HaoK:
@davidfowl so there's a problem with the new plan:  We don't have TUser in AddAuthentication so we can't add `SignInManager<TUser>` at that point...  So right now this is what we end up with:

```C#
   services.AddIdentityCore<User>()
        .AddEntityFrameworkStores<DbContext>()
        .AddSignIn();

   // Default all 4 identity cookies
   services.AddAuthentication()
        .AddIdentity()
             .AddDefaultCookies();
   
   // or add only the cookies you use, where each of these returns the IOptionsBuilder for the appropriate cookie options for further tweaking if needed
   var auth = services.AddAuthentication().AddIdentity();
   auth.AddApplicationCookie(o => o.CookieName = "app").Configure(Config["AppCookie"])
   auth.AddExternalCookie(o => o.CookieName = "ext")
   auth.AddTwoFactorUserIdCookie(o => o.CookieName = "tfa1")
   auth.AddTwoFactorRememberMeCookie(o => o.CookieName = "tfa2")   
```

HaoK:
Updated the PR with the new changes as discussed with @davidfowl 

So it looks like there's no need to split SignInManager into a new package, because for things to actually work, we end up still needing to add a bunch of cookie specific services to make things like the security stamp work by default, so even if we created a SignIn and SignIn.Cookies package.  We'd still need to glue everything together in the main Identity package, so there wasn't any benefit to creating the new packages at this time...

But the new refactored AddIdentityCore looks like this:

```
            services.AddIdentityCore<ApplicationUser>()
                .AddRoles<IdentityRole>()
                .AddSignInManager()
                .AddDefaultTokenProviders();

            services.AddAuthentication(o =>
            {
                o.DefaultScheme = IdentityConstants.ApplicationScheme;
                o.DefaultSignInScheme = IdentityConstants.ExternalScheme;
            })
            .AddIdentityCookies(o => 
            {
                // This would be empty typically, just demonstrating the new API per cookie
                o.ApplicationCookie.Configure(c => c.Cookie.Name = "MainCookie");
                o.ExternalCookie.Configure(c => c.Cookie.Name = "SocialCookie");
                o.TwoFactorRememberMeCookie.Configure(c => c.Cookie.Name = "RememberBrowser");
                o.TwoFactorUserIdCookie.Configure(c => c.Cookie.Name = "2falogin");
            });
```

The main additions to what we had in 2.0 are:
- `AddSignInManager`: which adds SignInManager + security stamp services needed for cookie validation
- `AddIdentityCookies`: which hang off `AuthenticationBuilder` instead of `IdentityBuilder` which add the 4 identity cookies and take a new `IdentityCookieBuilder` which has `OptionsBuilder` instances for each cookie which is intended to be a cleaner alternative to configuring those cookies.  Also its up to the caller to configure the default schemes themselves

cc @ajcvickers @blowdart

davidfowl:
AddIdentityCookies can only be called once right?

HaoK:
Yeah, `AddIdentityCookies` can only be called once, the underlying `AddCookie` calls will throw if the same scheme is passed in.

Issue #1445, opened 2017-09-28
PasswordValidator.IsLower and .IsUpper only reconizing the english alphabet


svethors:
When validating passwords and requiring either upper or lower case characters, non-english characters are not properly recognized as being upper or lower case. For example the Norwegian character 'Æ' is the upper case of 'æ', but will not be regarded as either, and so setting PasswordOptions.RequireLowercase = true will result in an error even if the lower case 'æ' is present in a password.

The code leading to the problem is found in Microsoft.AspNetCore.Identity.PasswordValidator:
Line 96:
```
public virtual bool IsLower(char c)
{
    return c >= 'a' && c <= 'z';
}
```
And line 106:
```
public virtual bool IsUpper(char c)
{
    return c >= 'A' && c <= 'Z';
}
```

blowdart:
This is a limitation of the .NET framework. Instead you'd have to implement your own IPasswordValidator and inject it into DI to override the default one.

PR #1446, opened 2017-09-28
Recognize non-english character casing in PasswordValidator (#1445)

svethors:


src/Microsoft.Extensions.Identity.Core/PasswordValidator.cs
```diff
@@ -95,7 +95,7 @@ public virtual bool IsDigit(char c)
         /// <returns>True if the character is a lower case ASCII letter, otherwise false.</returns>
         public virtual bool IsLower(char c)
         {
-            return c >= 'a' && c <= 'z';
+            return (c >= 'a' && c <= 'z') || (c >= 'à' && c <= 'ö') || (c >= 'ø' && c <= 'ÿ');
```
rosieks: This method returns a flag indicating whether the supplied character is a lower case ASCII letter. ASCII table doesn't include non-english letters.

svethors: True enough, so I would propose that the methods should not be limited to ASCII characters. And of course the doc comments changed accordingly. (The code above should includes full UTF-8 range of chars that have upper an lower case versions).

Alternatively, if there are good reasons only ASCII characters should be evaluated for case in a password, I would propose either a new configuration extension method like ReplacePasswordValidator, or possibly a ClearPasswordValidators that can be executed before AddPasswordValidator.

The reasoning being that creating a custom password validator that handles more than ASCII is easy enough, but calling AddPasswordValidator with the custom validator will leave the default validator in place, leading to the default reporting an error, while the custom reports OK. 

To replace the default validator one has to do something a bit mode cumbersome like this:
```
IdentityBuilder identityBuilder = services.AddIdentity<AppUser, IdentityRole>();

var serviceDescriptor = services.FirstOrDefault(descriptor => descriptor.ServiceType == typeof(IPasswordValidator<AppUser>));
services.Remove(serviceDescriptor);

identityBuilder.AddPasswordValidator<CustomPasswordValidator<AppUser>>();
```

PR #1479, opened 2017-10-23
 remove roleType while make UserOnlyStore

KhuongNtrd:


HaoK:
Thanks for finding this issue, I've filed an issue to track https://github.com/aspnet/Identity/issues/1480

Vertigo093i:
I suggest that the other branch of the `if` expression also needs a similar fix. By removing `roleType` parameter and replacing `UserStore<,,,>` with `UserOnlyStore<,,>`.

HaoK:
Ah yeah, good point, missing a test for that branch

HaoK:
Fixed in dev (2.1) via https://github.com/aspnet/Identity/pull/1481, we will consider porting the fix to a 2.0.x patch if there's demand

Issue #1480, opened 2017-10-23
AddEntityFrameworkStores with only User specifies too many generic arguments


HaoK:
Reported by https://github.com/aspnet/Identity/pull/1479

slaneyrw:
Just hit this bug, using 2.0.1.  Any timeframe of when this will be released?

Is there a work around ?

HaoK:
You can just add the concrete UserOnlyStore in ConfigureServices, something like, or one of the other overloads

```
  services.AddScoped<IUserStore, UserOnlyStore<ApplicationUser, ApplicationDbContext>>();
```

PR #1481, opened 2017-10-23
Fix issue with AddEntityFrameworkStores and user only store

HaoK:
Fixes https://github.com/aspnet/Identity/issues/1480

Also patches test hole, the existing tests were using the store classes directly, rather than exercising the AddEntityFrameworkStores reflection logic so this code was never hit.

@ajcvickers


HaoK:
794bd54cdbbd952d62674c9db76e6bd6a6e6b02f

PR #1499, opened 2017-11-09
Password validation failure does not propagate

tmerkel:
If validation fails, but no errors were returned in the Errors collection, add a default so the failure result propagates.

### DIFF
--- a/src/Microsoft.AspNetCore.Identity.Specification.Tests/IdentityResultAssert.cs
+++ b/src/Microsoft.AspNetCore.Identity.Specification.Tests/IdentityResultAssert.cs
@@ -44,12 +44,16 @@ namespace Microsoft.AspNetCore.Identity.Test
         /// <summary>
         /// Asserts that the result has not Succeeded and that first error matches error's code and Description.
         /// </summary>
-        public static void IsFailure(IdentityResult result, IdentityError error)
+        public static void IsFailure(IdentityResult result, IdentityError error = null)
         {
             Assert.NotNull(result);
             Assert.False(result.Succeeded);
-            Assert.Equal(error.Description, result.Errors.First().Description);
-            Assert.Equal(error.Code, result.Errors.First().Code);
+
+            if (error != null)
+            { 
+                Assert.Equal(error.Description, result.Errors.FirstOrDefault()?.Description);
+                Assert.Equal(error.Code, result.Errors.FirstOrDefault()?.Code);
+            }
         }
 
         /// <summary>
--- a/src/Microsoft.AspNetCore.Identity.Specification.Tests/UserManagerSpecificationTests.cs
+++ b/src/Microsoft.AspNetCore.Identity.Specification.Tests/UserManagerSpecificationTests.cs
@@ -161,6 +161,20 @@ namespace Microsoft.AspNetCore.Identity.Test
             }
         }
 
+        private class EmptyBadValidator : IUserValidator<TUser>,
+            IPasswordValidator<TUser>
+        {
+            public Task<IdentityResult> ValidateAsync(UserManager<TUser> manager, TUser user, string password)
+            {
+                return Task.FromResult(IdentityResult.Failed());
+            }
+
+            public Task<IdentityResult> ValidateAsync(UserManager<TUser> manager, TUser user)
+            {
+                return Task.FromResult(IdentityResult.Failed());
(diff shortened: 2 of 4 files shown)


HaoK:
Can you add a test to verify this new behavior in the specification tests?

 https://github.com/aspnet/Identity/blob/dev/src/Microsoft.AspNetCore.Identity.Specification.Tests/IdentitySpecificationTestBase.cs#L170

tmerkel:
Ahh..nice.  Will do!

tmerkel:
Added/updated the tests in UserManagerTests.

tmerkel:
Ok.  I added/updates tests in UserManagerTests...but it's not clear to me which way you'd like the tests updated in IdentitySpecificationTestBase. 

I think I can:

1. Update AlwaysBadValidator to leave the message out of the password validator result always...but this will change the existing tests.
2. Add a parameter to the constructor to include messages, like I did in UserManagerTests...but then do you want duplicate tests?  I think the first validator CreateAsync hits is the password validator, so maybe this is the right way to test both paths?

Do you have an opinion?

HaoK:
You can just add a new test validator for your new tests if needed, but yeah you defeinitely want to be exercising both paths

tmerkel:
Ok, I added tests to IdentitySpecificationTestBase for the path where Errors is empty.

Are there other things I missed?

Thanks for your help with this, Hao!

src/Microsoft.AspNetCore.Identity.Specification.Tests/IdentitySpecificationTestBase.cs
```diff
@@ -1139,6 +1206,34 @@ private static string MakeToken(string purpose, string userId)
         /// </summary>
         /// <returns>Task</returns>
         [Fact]
+        public async Task PasswordValidatorWithEmptyErrorsCanBlockResetPasswordWithStaticTokenProvider()
```
HaoK: Does this test require the FakeDataProtectionProvider?

tmerkel: In SetupIdentityService, .AddDefaultTokenProviders is called, which eventually causes a DataProtectorTokenProvider to be created

```
public DataProtectorTokenProvider(IDataProtectionProvider dataProtectionProvider, IOptions<DataProtectionTokenProviderOptions> options)
```

This asks the IServiceCollection for an IDataProtectionProvider...which...was never registered, as far as I could find.  Most of the tests fail for me with the error:

```
Message: System.InvalidOperationException : Unable to resolve service for type 'Microsoft.AspNetCore.DataProtection.IDataProtectionProvider' while attempting to activate 'Microsoft.AspNetCore.Identity.DataProtectorTokenProvider`1[Microsoft.AspNetCore.Identity.EntityFrameworkCore.IdentityUser]'.
```

Adding a fake fixed all of these...though I still have many failures related to code I haven't touched.  I can definitely remove the fake if there's something else taking care of this that I'm just missing...?

HaoK: Hrm that's strange, have you tried running the tests without your changes does a clean dev branch run work on your box?

tmerkel:
Your last comment made me realize my fork was of a feature branch.  Wiped everything, updated from the dev branch, readded my changes.  All things work like I'd expect.  Tests pass, no need for a fake.

I updated the PR to be to dev, since that's the new base.  I hope that's right.

HaoK:
Can you rebase your branch against dev?  I'll try to get this merged this week

tmerkel:
Rebased.  Just waiting for our legal person to get back to OK my clicking the CLA link.  Sorry about the delay there!

tmerkel:
Finally signed.  Thanks, Hao!

HaoK:
The code has moved to https://github.com/aspnet/AspNetCore/tree/master/src/Identity 

I'm happy to port this over to the new repo and merge this for 3.0, but that would lose the fact that these are your changes.  If you wanted to open a new PR in the new repo for the commits to stay in your profile.  We'll wait until next year before porting this otherwise

HaoK:
never mind ported to aspnetcore repo and was able to amend adding your history so we can take it from here, thanks again for the PR!

PR #1537, opened 2017-12-22
Revert Move identity into an Area in a library (using Razor Pages)

pakrym:
Revert https://github.com/aspnet/Identity/commit/2964515d8556e90fb88bcb6dcacc9f762965bba1 as it broke Univese-Coherence


natemcmaster:
Sorry @javiercn, but this broke signing builds.

:shipit:

PR #1544, opened 2017-12-26
make Logger non virtual in UserManager

### LINKED ISSUE #1441: UserManager.Logger is virtual and set in the constructor
Pretty much as the title says.

This is considered bad practice, as ctor order is the most base class to the most derived class, so if the property is overridden in a subclass it may not be properly initialized.

Static Analysis even has a rule against this: https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2214-do-not-call-overridable-methods-in-constructors (note that properties are effectively just a pair of methods, so the rule does in fact apply).

Here's a hypothetical example of something that could go wrong:

```
public class ApplicationUserManager : UserManager<ApplicationUser>
{
    private IObserver observer;

    public ApplicationUserManager(IObserver observer)
        : base(/* ... omitted for brevity ... */)
    {
        this.observer = observer;
    }

    public override ILogger Logger
    {
        get
        {
            this.observer.RecordPropertyGet(nameof(Logger));
            return base.Logger;
        }
        set
        {
            this.observer.RecordPropertySet(nameof(Logger), value);
            base.Logger = value;
        }
    }
}
```

In that example, the base class will set the `Logger` property, which will be the derived classes's impl, which uses `this.observer`, but oops it's null at that point in time since `ApplicationUserManager`'s ctor hasn't actually executed yet.

I am curious btw why this prop is virtual to begin with. It's a bit odd to see a virtual autoprop.

reaction1989:
Fixes #1441

### DIFF
--- a/src/Microsoft.Extensions.Identity.Core/UserManager.cs
+++ b/src/Microsoft.Extensions.Identity.Core/UserManager.cs
@@ -1,6 +1,10 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Identity.Core;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 using System;
 using System.Collections.Generic;
 using System.Globalization;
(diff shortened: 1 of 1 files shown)


blowdart:
Rejecting, breaking change would need to be in new major version.

PR #1555, opened 2018-01-09
Add Store.MaxLengthForKeys

HaoK:
If set this will set max length of things we use for composite keys in
UserTokens/Logins

Needed for https://github.com/aspnet/templating/issues/62

@ajcvickers

### DIFF
--- a/samples/IdentitySample.DefaultUI/Startup.cs
+++ b/samples/IdentitySample.DefaultUI/Startup.cs
@@ -32,7 +32,7 @@ namespace IdentitySample.DefaultUI
                 options => options.UseSqlServer(Configuration.GetConnectionString("DefaultConnection"),
                 x => x.MigrationsAssembly("IdentitySample.DefaultUI")));
 
-            services.AddIdentity<IdentityUser, IdentityRole>()
+            services.AddIdentity<IdentityUser, IdentityRole>(o => o.Stores.MaxLengthForKeys = 128)
                 .AddEntityFrameworkStores<IdentityDbContext>()
                 .AddDefaultUI()
                 .AddDefaultTokenProviders();
--- a/src/Core/StoreOptions.cs
+++ b/src/Core/StoreOptions.cs
@@ -1,0 +1,17 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Identity
+{
+    /// <summary>
+    /// Used for store specific options
+    /// </summary>
+    public class StoreOptions
+    {
+        /// <summary>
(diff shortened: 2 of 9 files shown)


HaoK:
Updated to set the max length to 128 in the DefaultUI packages, so this will be set by default in new apps

cc @javiercn

src/Core/StoreOptions.cs
```diff
@@ -0,0 +1,17 @@
...
+    /// <summary>
+    /// Used for store specific options
+    /// </summary>
+    public class StoreOptions
+    {
+        /// <summary>
+        /// If set to a positive number, the default OnModelCreating will use this value as the max length for any 
+        /// properties used as keys, i.e. UserId, LoginProvider, ProviderKey.
+        /// </summary>
+        public int MaxLengthForKeys { get; set; }
```
javiercn: This is very EF specific, should this be in Core?

HaoK: This one is kinda EF specific, but this is in Core because we are going to be adding the ProtectUserData stuff here as well, and Core will need to enforce something when the new flag is true...  

I was going to have StoreOptions be stand alone in the .Stores package if not for the next PR

samples/IdentitySample.DefaultUI/Startup.cs
```diff
@@ -32,7 +32,7 @@ public void ConfigureServices(IServiceCollection services)
                 options => options.UseSqlServer(Configuration.GetConnectionString("DefaultConnection"),
                 x => x.MigrationsAssembly("IdentitySample.DefaultUI")));
 
-            services.AddIdentity<IdentityUser, IdentityRole>()
+            services.AddIdentity<IdentityUser, IdentityRole>(o => o.Stores.MaxLengthForKeys = 128)
```
javiercn: Will this require a migration for existing users?
This is kind of a breaking change, isn't it?

HaoK: Yeah, so existing apps wouldn't be using the DefaultUI, if they switch to it, they will be opting into the new 2.1 schema...

HaoK: And if they don't want the breaking change, they can set MaxLengthForKeys to 0 to turn it off and its just like before

javiercn: Ah, yes. I see now. This should be part of our template by default, shouldn't it?

ajcvickers: @javiercn Yes, this will be on for new projects. Once this is in and turned on for new projects we can then scaffold the up-to-date migration and we should be set for 2.1.

People with existing apps can choose to set this on with whatever value makes sense for them, or even explicitly define different lengths in the model, and then generate a migration. But this is an explicit step that they must take so we don't break any existing apps by default.

PR #1559, opened 2018-01-09
Add Manage/Privacy-Delete/Download functionality

HaoK:
Initial changes to identity default UI for Manage/Privacy/DownloadPersonalData and DeletePersonalData

Still need final tweaks to UI/text which we'll plan to get from @DamianEdwards review in person.

@blowdart @danroth27 @javiercn @ajcvickers @mkArtakMSFT


blowdart:
There won't be time for in person with @DamianEdwards is off having fun in London, so let's use @danroth27 and I as his proxy.

HaoK:
@danroth27 who do you want to review this PR from the perspective of proper RazorPages usage/patterns etc?  I mostly just copied what the existing pages did, I only intentionally deviated slightly using the NavPages constant instead of hardcoding the value in the ActivePages setting.

DamianEdwards:
Get Ryan and Pranav to look at it pls

src/UI/Areas/Identity/Pages/Account/Manage/DeletePersonalData.cshtml
```diff
@@ -0,0 +1,28 @@
...
+</div>
+
+<div>
+    <form method="post" class="form-group">
+        <button class="btn btn-danger" type="submit">Delete user and personal data</button>
+    </form>
+</div>
+
+@section Scripts {
+    @await Html.PartialAsync("_ValidationScriptsPartial")
```
pranavkm: `<partial name="_ValidationScriptsPartial" />`

src/UI/Areas/Identity/Pages/Account/Manage/DeletePersonalData.cshtml.cs
```diff
@@ -0,0 +1,61 @@
...
+            {
+                throw new ApplicationException($"Unexpected error occurred deleteing user with ID '{user.Id}'.");
+            }
+
+            await _signInManager.SignOutAsync();
+
+            _logger.LogInformation("User with ID '{UserId}' deleted themselves.", _userManager.GetUserId(User));
+
+            // REVIEW: should this redirect to somewhere that says user was deleted?
+            return RedirectToPage("./");
```
pranavkm: This should redirect to the website root `return Redirect("~/");` or a successfully deleted page. Redirecting to the page `./` is kinda weird.

src/UI/Areas/Identity/Pages/Account/Manage/DownloadPersonalData.cshtml
```diff
@@ -0,0 +1,12 @@
...
+@model DownloadPersonalDataModel
+@{
+    ViewData["Title"] = "Download Your Data";
+    ViewData["ActivePage"] = ManageNavPages.DownloadPersonalData;
+}
+
+<h4>@ViewData["Title"]</h4>
+
+@section Scripts {
+    @await Html.PartialAsync("_ValidationScriptsPartial")
```
pranavkm: `<partial name="_ValidationScriptsPartial" />`

src/UI/Areas/Identity/Pages/Account/Manage/DownloadPersonalData.cshtml.cs
```diff
@@ -0,0 +1,52 @@
...
+namespace Microsoft.AspNetCore.Identity.UI.Pages.Account.Manage
+{
+    public class DownloadPersonalDataModel : PageModel
+    {
+        private readonly UserManager<IdentityUser> _userManager;
+        private readonly ILogger<PrivacyModel> _logger;
+
+        public DownloadPersonalDataModel(
+            UserManager<IdentityUser> userManager,
+            ILogger<PrivacyModel> logger)
```
pranavkm: `ILogger<DownloadPersonalDataModel>`?

src/UI/Areas/Identity/Pages/Account/Manage/DownloadPersonalData.cshtml.cs
```diff
@@ -0,0 +1,52 @@
...
+            var user = await _userManager.GetUserAsync(User);
+            if (user == null)
+            {
+                return NotFound($"Unable to load user with ID '{_userManager.GetUserId(User)}'.");
+            }
+
+            _logger.LogInformation("User with ID '{UserId}' asked for their personal data.", _userManager.GetUserId(User));
+
+            // REVIEW: do we want to set a force download header?
+            return new FileContentResult(Encoding.ASCII.GetBytes(JsonConvert.SerializeObject(user)), "text/json");
```
pranavkm: You likely need Encoding.UTF8. That said, this kinda feels icky, but there really doesn't seem like a more definitive way to do this.

src/UI/Areas/Identity/Pages/Account/Manage/DeletePersonalData.cshtml.cs
```diff
@@ -0,0 +1,61 @@
...
+    public class DeletePersonalDataModel : PageModel
+    {
+        private readonly UserManager<IdentityUser> _userManager;
+        private readonly SignInManager<IdentityUser> _signInManager;
+        private readonly ILogger<PrivacyModel> _logger;
+
+        public DeletePersonalDataModel(
+            UserManager<IdentityUser> userManager,
+            SignInManager<IdentityUser> signInManager,
+            ILogger<PrivacyModel> logger)
```
pranavkm: `ILogger<DeletePersonalDataModel>`

src/UI/Areas/Identity/Pages/Account/Manage/DeletePersonalData.cshtml.cs
```diff
@@ -0,0 +1,61 @@
...
+        public async Task<IActionResult> OnPostAsync()
+        {
+            var user = await _userManager.GetUserAsync(User);
+            if (user == null)
+            {
+                return NotFound($"Unable to load user with ID '{_userManager.GetUserId(User)}'.");
+            }
+
+            var result = await _userManager.DeleteAsync(user);
+            if (!result.Succeeded)
```
pranavkm: Is it still correct for a user to be signed in if this fails?

src/UI/Areas/Identity/Pages/Account/Manage/DeletePersonalData.cshtml.cs
```diff
@@ -0,0 +1,61 @@
...
+            {
+                throw new ApplicationException($"Unexpected error occurred deleteing user with ID '{user.Id}'.");
```
pranavkm: `ApplicationException` -> `InvalidOperationException`

src/UI/Areas/Identity/Pages/Account/Manage/Privacy.cshtml
```diff
@@ -0,0 +1,23 @@
+﻿@page
+@model PrivacyModel
+@{
+    ViewData["Title"] = "Privacy";
+    ViewData["ActivePage"] = "Privacy";
```
pranavkm: `ManageNavPages.Privacy`

src/UI/Areas/Identity/Pages/Account/Manage/DownloadPersonalData.cshtml.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        public DownloadPersonalDataModel(
+            UserManager<IdentityUser> userManager,
+            ILogger<PrivacyModel> logger)
+        {
+            _userManager = userManager;
+            _logger = logger;
+        }
+
+        public async Task<IActionResult> OnGet()
```
pranavkm: When do we get to this handler? `Privacy.cshtml` posts to this action and there isn't a link to this page.

src/UI/Areas/Identity/Pages/Account/Manage/DeletePersonalData.cshtml.cs
HaoK: The default implementation basically never will fail,  it would only blow up with an exception, but this is a valid point, I guess we should only sign out if it succeeded

src/UI/Areas/Identity/Pages/Account/Manage/DeletePersonalData.cshtml
pranavkm: We generally prefer tag helpers to HtmlHelpers. It's cleaner, more idiomatic code.

src/UI/Areas/Identity/Pages/Account/Manage/DeletePersonalData.cshtml.cs
```diff
@@ -0,0 +1,61 @@
...
+            }
+
+            await _signInManager.SignOutAsync();
+
+            _logger.LogInformation("User with ID '{UserId}' deleted themselves.", _userManager.GetUserId(User));
```
mkArtakMSFT: Is it ok to log the userId as is (as this can be the email address, which is PII)

src/UI/Areas/Identity/Pages/Account/Manage/Privacy.cshtml
```diff
@@ -0,0 +1,23 @@
...
+            <button type="submit">Download your data</button>
+        </form>
+        <p>
+            <a asp-page="DeletePersonalData" class="btn btn-default">Delete your user data</a>
+        </p>
+    </div>
+</div>
+
+@section Scripts {
+    @await Html.PartialAsync("_ValidationScriptsPartial")
```
mkArtakMSFT: Similar to the rest of the comments - prefer <partial ... /> tag syntax instead

src/UI/Areas/Identity/Pages/Account/Manage/DeletePersonalData.cshtml.cs
HaoK: The UserName is the email, the user id is a guid by default, but this is the pattern the existing templates use for logging I believe.

src/UI/Areas/Identity/Pages/Account/Manage/DownloadPersonalData.cshtml.cs
HaoK: Yeah we don't, I'll remove

src/UI/Areas/Identity/Pages/Account/Manage/DeletePersonalData.cshtml
```diff
@@ -0,0 +1,28 @@
...
+
+<h4>@ViewData["Title"]</h4>
+
+<div class="alert alert-warning" role="alert">
+    <p>
+        <span class="glyphicon glyphicon-warning-sign"></span>
+        <strong>This action will delete your user.</strong>
+    </p>
+    <p>
+        This is unrecoverable.
```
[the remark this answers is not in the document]
HaoK: "This action will delete your user."
This is unrecoverable.

means you can't recover your user after you delete it.

src/UI/Areas/Identity/Pages/Account/Manage/DeletePersonalData.cshtml.cs
mkArtakMSFT: Thanks!

HaoK:
Ok updated switching all default UI over to use the new partial and consistently use ManageNavClass constants in for the active page stuff

src/UI/Areas/Identity/Pages/Account/Manage/DeletePersonalData.cshtml.cs
HaoK: So most things in manage are doing some form of RedirectToPage...  is one better than the other?

pranavkm: RedirectToPage expects to resolve a page. Since there isn't a page named `./`, it ends up resolving the empty string which ends up in the app root. But that's just a weird way of going about it.

src/UI/Areas/Identity/Pages/Account/Manage/DeletePersonalData.cshtml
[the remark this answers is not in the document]
pranavkm: Should this elaborate and say, once deleted you cannot use your login credentials and would need to re-register?

Kinda unrelated, but is there a hook that let's the app developer also clean up other stored details about the user when this is deleted? For instance, if the user id is used to store my order history or posted comments?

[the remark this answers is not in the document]
HaoK: Don't worry about the exact text yet, I'll update that shortly.

re hook, not really, ideally we'd have an easy way for them to just drop in a new DeletePersonalData.cshtml.cs to take over this work entirely to customize.  Or I guess scaffold it all and change the code if that feature doesn't make it...

src/UI/Areas/Identity/Pages/Account/Manage/DeletePersonalData.cshtml.cs
HaoK: Yeah so after they get deleted, manage will no longer be valid, so I really was just trying to redirect to home or whatever.  Unless we think its worth adding a dedicated page for a Deleted confirmation

HaoK:
Text has been updated, Privacy => Personal Data, will add one more change which will require users to reenter their password if they have a local password before being allowed to delete their account

HaoK:
Updated flow with password confirmation on Delete page if the user has a password.  Also added white list to only include Id/UserName/Email/Phone/and Confirmed properties)

HaoK:
I think all feedback has been addressed now, added `Response.Headers.Add("Content-Disposition", "attachment; filename=PersonalData.json");` to force download.

@pranavkm can you do a final review/sign off?

src/UI/Areas/Identity/Pages/Account/Manage/DeletePersonalData.cshtml.cs
```diff
@@ -0,0 +1,84 @@
...
+            {
+                return NotFound($"Unable to load user with ID '{_userManager.GetUserId(User)}'.");
+            }
+
+            RequirePassword = await _userManager.HasPasswordAsync(user);
+            if (RequirePassword)
+            {
+                if (!await _userManager.CheckPasswordAsync(user, Input.Password))
+                {
+                    ModelState.AddModelError(string.Empty, "Password not correct.");
```
pranavkm: `AddModelError("Input.Password", `) if we'd like to show error along the password field.

HaoK:
f1ed482c26df6e242b1e4eca0948dc77ba5b8fa9

grahamehorner:
If a user has two factor authentication enable would the second factor used for verification before deletion?

HaoK:
No, we just check password right now, checking a second factor is significantly more involved, but its certainly doable if an app wants that, you basically just need to duplicate the two factor login flow

PR #1561, opened 2018-01-10
Update initial migration for DefaultUI sample

HaoK:
Test run before template migration update, does this snapshot look like what you would expect now @ajcvickers ?

cc @javiercn


ajcvickers:
@HaoK This looks right to me.

PR #1562, opened 2018-01-11
Initial EncryptPersonalData support

HaoK:
- new StoreOptions.EncryptPersonalData setting
- new IPersonalDataEncryptor and IEncryptedUserStore interfaces that live in core
- new `AddPersonalDataEncryptor<MyEncryptor>()` helper method on IdentityBuilder
- The two concrete user stores in EF now implement IEncryptedUserStore
- UserManager ctor throws if EncryptPersonalData is true if store is not IEncryptedUserStore, or if IPersonalDataEncryptor service not found
- base IdentityUserContext hooks up ValueConverter using IPersonalDataEncryptor on UserName, NormalizedUserName, Email, NormalizedEmail, PhoneNumber columns when EncryptPersonalData is true.

So all of the normal store tests except one pass when turning this on with a dummy reverse encryptor. 
 There's a test that that directly uses the DbSet as an IQueryable to look for users like so: `UserManager.Users.Count(u => u.UserName.StartsWith("CanFindUsersViaUserQuerable"))` which fails because this probably bypasses the converters... @ajcvickers ?

cc @blowdart @javiercn


ajcvickers:
@HaoK For `UserManager.Users.Count(u => u.UserName.StartsWith("CanFindUsersViaUserQuerable"))` can you look in the logs and see what SQL gets generated. There's a couple of reasons why I think this might be failing.

src/EF/IdentityUserContext.cs
```diff
@@ -96,16 +96,16 @@ public abstract class IdentityUserContext<TUser, TKey, TUserClaim, TUserLogin, T
...
+        private StoreOptions GetStoreOptions() => this.GetService<IDbContextOptions>()
                             .Extensions.OfType<CoreOptionsExtension>()
                             .FirstOrDefault()?.ApplicationServiceProvider
                             ?.GetService<IOptions<IdentityOptions>>()
                             ?.Value?.Stores;
-            return options != null ? options.MaxLengthForKeys : 0;
+
+        private class PersonalDataConverter : ValueConverter<string, string>
+        {
+            public PersonalDataConverter(IPersonalDataEncryptor encryptor) : base(s => encryptor.Encrypt(s), s => encryptor.Decrypt(s), default)
```
HaoK: Change IPersonalDataEncryptor to singleton in AddPerosnalDataEncryptor

HaoK:
I don't think this needs to go into preview1 at this point, as its basically a bunch of hooks to make it easy to plug in encryption, it wouldn't do anything by default.  

@blowdart is this something you want me to try and get in this week so it makes it into preview1?

HaoK:
@ajcvickers here's the SQL generated:

```
Microsoft.EntityFrameworkCore.Query:Debug: Compiling query model: 
'(from IdentityUser u in DbSet<IdentityUser>
where bool [u].UserName.StartsWith(__userName_0)
select [u]).Count()'
Microsoft.EntityFrameworkCore.Query:Debug: Optimized query model: 
'(from IdentityUser u in DbSet<IdentityUser>
where bool [u].UserName.StartsWith(__userName_0)
select [u]).Count()'
Microsoft.EntityFrameworkCore.Query:Debug: (QueryContext queryContext) => IEnumerable<int> _InterceptExceptions(
    source: IEnumerable<int> _ToSequence(() => int GetResult(
            valueBuffers: IEnumerable<ValueBuffer> _Query(
                queryContext: queryContext, 
                shaperCommandContext: SelectExpression: 
                    SELECT COUNT(*)
                    FROM [AspNetUsers] AS [u]
                    WHERE ([u].[UserName] LIKE @__userName_0 + N'%' AND (LEFT([u].[UserName], LEN(@__userName_0)) = @__userName_0)) OR (@__userName_0 = N'')), 
            throwOnNullResult: False)), 
    contextType: Microsoft.AspNetCore.Identity.EntityFrameworkCore.Test.SqlStoreTestBase`3+TestDbContext[Microsoft.AspNetCore.Identity.IdentityUser,Microsoft.AspNetCore.Identity.IdentityRole,System.String], 
    logger: DiagnosticsLogger<Query>, 
    queryContext: queryContext)
Microsoft.EntityFrameworkCore.Database.Connection:Debug: Opening connection to database 'Scratch_af2039c9-4611-4753-a468-35c632f40812' on server '(localdb)\MSSqlLocaldb'.
Microsoft.EntityFrameworkCore.Database.Connection:Debug: Opened connection to database 'Scratch_af2039c9-4611-4753-a468-35c632f40812' on server '(localdb)\MSSqlLocaldb'.
Microsoft.EntityFrameworkCore.Database.Command:Debug: Executing DbCommand [Parameters=[@__userName_0='?' (Size = 256)], CommandType='Text', CommandTimeout='30']
SELECT COUNT(*)
FROM [AspNetUsers] AS [u]
WHERE ([u].[UserName] LIKE @__userName_0 + N'%' AND (LEFT([u].[UserName], LEN(@__userName_0)) = @__userName_0)) OR (@__userName_0 = N'')
Microsoft.EntityFrameworkCore.Database.Command:Information: Executed DbCommand (79ms) [Parameters=[@__userName_0='?' (Size = 256)], CommandType='Text', CommandTimeout='30']
SELECT COUNT(*)
FROM [AspNetUsers] AS [u]
WHERE ([u].[UserName] LIKE @__userName_0 + N'%' AND (LEFT([u].[UserName], LEN(@__userName_0)) = @__userName_0)) OR (@__userName_0 = N'')

```

HaoK:
@blowdart @ajcvickers @javiercn 

So took a stab at actually implementing support for plugging in encryption with key rotation as part of identity just to see how it looks...  

So its gets a little bit muddled, but it 'works'.  New shape of things:

```C#
    // user manager and EF converter use key ring + encryptor
    public interface IPersonalDataEncryptorKeyRing
    {
        string CurrentKeyId { get; }
        string this[string keyId] { get; }
        IEnumerable<string> GetAllKeyIds();
    }
    public interface IPersonalDataEncryptor
    {
        string Encrypt(string keyId, string data);
        string Decrypt(string keyId, string data);
    }

    // Used by EF value converter, default implementation uses encryptor + keyring
    public interface IPersonalDataProtector
    {
        string Protect(string data);
        string Unprotect(string data);
    }
```

So the good news is we already do special normalization and have methods to do FindByUserName/Email, so it was simple to just apply encryption and probe using the key ring inside of UserManager.  Even better, we already have our own dedicated NormalizedUserName/Email fields which we can manage the protected payload, so we can just have the manager manage that, and drop configuring value converters for those two fields.  We still need to use the value converter for the phone number, and this is also what we would tell customers to use to protect their own (non key data).

So to turn on PersonalDataProtection in this iteration, it looks something like this, I verified that you can rotate the key and user manager can still find the old values, it actually also reprotects the keys every time the user is modified automatically too, so the keys are always rolled forward anytime the user is updated:
```C#
   services.AddIdentity(o => o.Stores.ProtectPersonalData = true)
            .AddPersonalDataEncryption<PadEncryptor, DefaultKeyRing>();

        public class DefaultKeyRing : IPersonalDataEncryptorKeyRing
        {
            public static string Current = "Default";
            public string this[string keyId] => keyId;
            public string CurrentKeyId => Current;

            public IEnumerable<string> GetAllKeyIds()
            {
                return new string[] { "Default", "NewPad" };
            }
        }

        private class PadEncryptor : IPersonalDataEncryptor
        {
            private readonly IPersonalDataEncryptorKeyRing _keyRing;

            public SillyEncryptor(IPersonalDataEncryptorKeyRing keyRing) => _keyRing = keyRing;

            public string Decrypt(string keyId, string data)
            {
                var pad = _keyRing[keyId];
                if (!data.StartsWith(pad))
                {
                    throw new InvalidOperationException("Didn't find pad.");
                }
                return data.Substring(pad.Length);
            }

            public string Encrypt(string keyId, string data)
                => _keyRing[keyId] + data;
        }

        [Fact]
        public async Task CanRotateKeysAndStillFind()
        {
            var manager = CreateManager();
            var name = Guid.NewGuid().ToString();
            var user = CreateTestUser(name);
            IdentityResultAssert.IsSuccess(await manager.CreateAsync(user));
            IdentityResultAssert.IsSuccess(await manager.SetEmailAsync(user, "hao@hao.com"));
            var newName = Guid.NewGuid().ToString();
            Assert.Null(await manager.FindByNameAsync(newName));
            IdentityResultAssert.IsSuccess(await manager.SetPhoneNumberAsync(user, "123-456-7890"));

            Assert.Equal(user, await manager.FindByEmailAsync("hao@hao.com"));

            IdentityResultAssert.IsSuccess(await manager.SetUserNameAsync(user, newName));
            IdentityResultAssert.IsSuccess(await manager.UpdateAsync(user));
            Assert.NotNull(await manager.FindByNameAsync(newName));
            Assert.Null(await manager.FindByNameAsync(name));
            DefaultKeyRing.Current = "NewPad";
            Assert.NotNull(await manager.FindByNameAsync(newName));
            Assert.Equal(user, await manager.FindByEmailAsync("hao@hao.com"));
            Assert.Equal("123-456-7890", await manager.GetPhoneNumberAsync(user));
        }
```

HaoK:
From discussion today, the overall shape/plan looks reasonable, we'll rename things a bit:

```C#
    // UserManager uses lookup protector and keyring to probe for matching keys as part of
    // lookups when ProtectPersonalData = true
    public interface ILookupKeyRing
    {
        string CurrentKeyId { get; }
        string this[string keyId] { get; }
        IEnumerable<string> GetAllKeyIds();
    }
    public interface ILookupProtector
    {
        string Protect(string keyId, string data);
        string Unprotect(string keyId, string data);
    }

    // Used by EF value converter, default implementation uses lookup protector + keyring
    public interface IPersonalDataProtector
    {
        string Protect(string data);
        string Unprotect(string data);
    }
```

@blowdart wanted to get thoughts/feedback from @GrabYourPitchforks @leastprivilege

To summarize the intent behind these changes: we want expose a general set of hooks to allow apps to setup encryption of personal data (and potentially be able to do key rotation) while still being able to login using identity...  The app would be responsible for providing an implementation of the LookupProtector/KeyRing.  We can provide a default implementation of IPersonalDataProtector that uses {keyId}:{payload} as the format that uses the ILookupProtector/KeyRing.  Identity will take care of using the lookup interfaces to search for users using each of the keys, and also update the data to the current key whenever a user is saved.

So it would look something like in startup:
```
   services.AddIdentity(o => o.ProtectPersonalData = true)
               .AddLookupProtection<AppLookupKeyRing, AppLookupProtector>();
```

HaoK:
Add ToList for user manager.userName to work

javiercn:
@HaoK Looks good, IMO you can leave the KeyRotation part out. Having the interface doesn't really add much without an actual implementation. You can have something like the implementation of IPersonalDataProtector do the right thing by emdedding the key into the protected payload (the same way data protection does it today).

You should always protect with the current key and unprotect with whatever key was used at the time.
If a user wants to reencrypt data, they can do so by retrieving the data (unprotecting it) and protecting it with the new key.

HaoK:
The key rotation piece is actually part of core identity, without that they would have to implement that themselves on top of identity apis instead of the user manager taking care of doing lookups and automatically updating the encryption for the key in FindByName/Email.  

We don't care what the implementation is, but without this support, they'd have a lot more things to change to support key rotation since they wouldn't be able to just use FindByName/Email anymore

javiercn:
I thought we were going to do a blind index for lookups, isn't that the case?

HaoK:
That's not the plan at this point.  The default protector will use a payload format of "keyid":"data", lookups will just try every key and update the payload to use the current key if needed.

They can still plug in the entire data protector if they want, but this should be good enough to allow them to only plug in encryption algorithms.  I still need to verify everything works using something more complicated than the fake encryption tests in this PR.

HaoK:
@javiercn I added 2 new functional tests for Download/Delete user can you review those, I mostly just copied the TwoFA manage test?  I didn't add any coverage of for custom users since the test website is currently using only IdentityUser, do we need a second test website that's setup more like the DefaultUI sample which lets us test the extensiblity flows as well?

src/Core/UserManager.cs
```diff
@@ -527,15 +541,36 @@ public virtual Task<TUser> FindByIdAsync(string userId)
...
             userName = NormalizeKey(userName);
-            return Store.FindByNameAsync(userName, CancellationToken);
+
+            var user = await Store.FindByNameAsync(userName, CancellationToken);
+
+            // Need to potentially check all keys
+            if (user == null && Options.Stores.ProtectPersonalData)
+            {
+                var keyRing = _services.GetService<ILookupProtectorKeyRing>();
+                var encryptor = _services.GetService<ILookupProtector>();
```
HaoK: replace encrypt => protector, also in IEncryptedUserStore => IProtectedUserStore

BrightSoul:
Hi, I'm struggling to make this work with the Data Protection API and @HaoK I need your guidance. My goal is to protect all of the user information: email/username, cellphone and any other property I might have added to the ApplicationUser class and decorated with the `[ProtectedPersonalData]`.

First thing I did was to enable protection of personal data in ASP.NET Core Identity.
```
services.AddDefaultIdentity<ApplicationUser>(o => {
  o.Stores.ProtectPersonalData = true;
})
.AddPersonalDataProtection<LookupProtector, LookupProtectorKeyRing>();
```
My implementation of the `ILookupProtector` is the following one. I thought it would be a good idea to use the `IDataProtectionProvider` to protect and unprotect strings since it would be so easy to implement.
```
public class LookupProtector : ILookupProtector
{
  private readonly IDataProtector protector;
  public LookupProtector(IDataProtectionProvider protectionProvider)
  {
      this.protector = protectionProvider.CreateProtector("UserData");
  }
  public string Protect(string keyId, string data)
  {
      return protector.Protect(data);
  }  
  public string Unprotect(string keyId, string data)
  {
      return protector.Unprotect(data);
  }
}
```
When a users registers an account, his values are indeed encrypted in the database and shown as clear text when he visits his profile page.
But then, he *can't login*. I realized that the value returned by the `Protect` method of the `IDataProtector` is NOT deterministic. What can I do at this point? Is there *any* way to make the `IDataProtector` produce deterministic results? I don't want to leave the email address unprotected and I don't want to roll my own encryptor.
Thank you.

HaoK:
Yeah we probably need to document this at some point cc @blowdart @Rick-Anderson.

But `ILookupProtector` needs to be stable for logins to work, so you can't use data protection you will have to use something else

BrightSoul:
Thanks for your response. I'm concerned about this "something else" I might end up using. If you leave it to me, it will be for sure a sub-optimal solution, security-wise.
Instead, I wonder if the Data Protection API could be improved to allow AES with a static IV, as an additional option to the [EncryptionAlgorithm enumeration](https://github.com/aspnet/DataProtection/blob/release/2.1/src/Microsoft.AspNetCore.DataProtection/AuthenticatedEncryption/EncryptionAlgorithm.cs). This way, I could perhaps get deterministic results out of the `Protect` method. Dictionary attacks could be mitigated by transparently adding/removing some static salt to the string. I know this might sound like a naive solution but I think still better than the one I could roll on my own.

The alternative is to choose proper usernames instead of e-mail addresses. I'm more comfortable with leaving usernames unprotected rather than e-mail addresses. Logins and `FindByUsername` would work again but then I need a configuration option to tell Identity I want unprotected usernames when invoking the  `AddPersonalDataProtection` extension method.

blowdart:
I wouldn't view that as an improvement to data protection at all I'm afraid. We deliberately designed it for maximum safety, adding a static IV or making it deterministic would move away from that design philosophy.

BrightSoul:
I understand you don't want to lower the security of a Microsoft branded API but there's a missing stepping stone here. The whole functionality provided by `AddPersonalDataEncryptor` is unusable unless I, the developer, provide a working implementation of an `ILookupProtector`. I agree that having a static IV is not the most secure option but then... how should I do it? Security experts at Microsoft should provide some guidance here. With GDPR enforcement just a few days away, developers will adopt the first quick and dirty way they come up with to implement the `ILookupProtector` and show their bosses the application is compliant.
Again, the other option is to leave usernames in clear text and, in order to do that, the `Protect` method should provide some context about what it is that I'm protecting. Maybe a third parameter with the property name would suffice.

critot:
I would not believe that Microsoft's commitment to the GDPR and personal data protection in general is limited to only HTTPS by default, Cookie Consent Bar and Default Data Control actions.... and leave this sensitive implementation to the individual develop....
@blowdart 
if it's not possible to improve Data Protection Api, EncryptPersonalData should still have a default implementation of ILookupProtector, don't you think?
thank you.

blowdart:
I'm working on some samples, but you'll see when I get them done, that the real recommendation will be "Use encryption at rest on your data". MS SQL and SQL Azure provide encryption at REST with no changes needed to your code. For Linux the various databases there end up saying use full disk encryption on the volumes where databases reside, for which there are multiple options.

The protectors here are really a fallback of last resort.

And there cannot be a default implementation, because it will fall to you to manage keys, both in terms of rotation, size, even algorithms, based on your risk evaluation, which should be part of every GDPR review.

Rick-Anderson:
See [GDPR Encryption at rest](https://docs.microsoft.com/en-us/aspnet/core/security/gdpr?view=aspnetcore-2.1#encryption-at-rest)

critot:
@blowdart @Rick-Anderson 
thank you for your suggestion, this makes me feel more comfortable

egmfrs:
Is there any documentation for this yet?

PR #1577, opened 2018-01-18
Add ForgetMachine TFA button

HaoK:
Part of https://github.com/aspnet/Identity/issues/1342

Add ability to forget this machine for purposes of TFA signin.

@pranavkm @javiercn @blowdart @ajcvickers


src/UI/Areas/Identity/Pages/Account/Manage/TwoFactorAuthentication.cshtml
```diff
@@ -29,6 +29,12 @@
         </div>
     }
 
+    if (Model.IsMachineRemembered)
+    {
+        <form method="post" style="display: inline-block">
```
HaoK: REVIEW: Is this the 'right' way to get the buttons on the same row?

src/UI/Areas/Identity/Pages/Account/Manage/TwoFactorAuthentication.cshtml.cs
```diff
@@ -42,6 +45,25 @@ public class TwoFactorAuthenticationModel : PageModel
...
+            }
+
+            await _signInManager.ForgetTwoFactorClientAsync();
+
+            HasAuthenticator = await _userManager.GetAuthenticatorKeyAsync(user) != null;
+            Is2faEnabled = await _userManager.GetTwoFactorEnabledAsync(user);
+            IsMachineRemembered = false;
             RecoveryCodesLeft = await _userManager.CountRecoveryCodesAsync(user);
 
             return Page();
```
pranavkm: If you do Post-Redirect-Get, you wouldn't have to rehydrate the data in the OnPost. Would 

``OnPost()
{
   await _signInManager.ForgetTwoFactorClientAsync();
   return RedirectToPage();
}
```

File: src/UI/Areas/Identity/Pages/Account/Manage/TwoFactorAuthentication.cshtml.cs [hunk 3]
```diff
@@ -32,6 +32,9 @@ public class TwoFactorAuthenticationModel : PageModel
         [BindProperty]
         public bool Is2faEnabled { get; set; }
 
+        [BindProperty]
```
pranavkm: This shouldn't be a model bound property (same with `Is2faEnabled`).

src/UI/Areas/Identity/Pages/Account/Manage/TwoFactorAuthentication.cshtml
pranavkm: This is fine. Bootstrap 4.0 has a utility class for this, but I guess we're on an older version.

HaoK:
Ah much nicer, updated

src/UI/Areas/Identity/Pages/Account/Manage/TwoFactorAuthentication.cshtml.cs
```diff
@@ -42,9 +43,22 @@ public class TwoFactorAuthenticationModel : PageModel
...
+        public async Task<IActionResult> OnPost()
+        {
+            var user = await _userManager.GetUserAsync(User);
+            if (user == null)
+            {
+                return NotFound($"Unable to load user with ID '{_userManager.GetUserId(User)}'.");
+            }
+
+            await _signInManager.ForgetTwoFactorClientAsync();
+            return RedirectToPage();
```
pranavkm: Sorry, I should have included this in my review earlier. We should probably print a banner after the Redirect is complete that says the machine has been forgotten.

```C#
[TempData]
public string BannerMessage { get; set; }

...

await _signInManager.ForgetTwoFactorClientAsync();
BannerMessage = "The current browser has been forgotten. You will be asked to login with your credentials the next time you visit this website"
return RedirectToPage();
```

and display the banner in the page if it's set.

src/UI/Areas/Identity/Pages/Account/Manage/TwoFactorAuthentication.cshtml
```diff
@@ -29,6 +29,12 @@
...
+            <button type="submit" class="btn btn-default">Forget machine</button>
```
pranavkm: Should this be `Forget browser`? Maybe a PM could help here.

HaoK: @blowdart text help?

src/UI/Areas/Identity/Pages/Account/Manage/TwoFactorAuthentication.cshtml.cs
HaoK: This isn't forgetting the login cookie, note this is only clearing the two factor cookie, so it will be more that you will be prompted for two factor again in your next login.

blowdart: Might still want confirmation though to show people it works

src/UI/Areas/Identity/Pages/Account/Manage/TwoFactorAuthentication.cshtml
HaoK: Maybe? "Forget two-factor for this machine"

src/UI/Areas/Identity/Pages/Account/Manage/TwoFactorAuthentication.cshtml.cs
HaoK: Yeah I just meant the banner text is wrong, we can show a banner, altho we aren't consistent about this right now I don't think

blowdart: And the banner should say "The current browser has been forgotten. When you login again from this browser you will be prompted for your 2fa code."

src/UI/Areas/Identity/Pages/Account/Manage/TwoFactorAuthentication.cshtml
blowdart: Forget this browser

src/UI/Areas/Identity/Pages/Account/Manage/TwoFactorAuthentication.cshtml.cs
HaoK: @pranavkm what's the best pattern to follow for displaying the banner, do I need to go the full on Html.Partial/_statusMessage route that Manage/Index is currently doing?

HaoK: Actually I'll just use the existing statusMessage partial in here, @blowdart should we add status messages for all the buttons on the TFA page to be consistent?

pranavkm: You could just do

```
// TwoFactorAuth.cshtml
if (Model.BannerMessage != null)
{
<div class="alert alert-success" role="alert">
  @Model.BannerMessage
</div>
}
```
if you set it via TempData. Shouldn't need a partial

HaoK: Ok sounds good.

HaoK:
@blowdart looks like today there's no easy way to re-enable TFA after disabling it...  Should I add that button where you have an authenticator app but its not enabled...there's no way to get it back other than configuring another authenticator app today

blowdart:
Setting it up again would be the right way to do in my opinion, after disabling the keys should be treated as tainted imo, and you're forced to re-enable and rescan. So I'm OK with the flow as it is.

grahamehorner:
@HaoK should also think about supporting multiple authenticator apps ?

src/UI/Areas/Identity/Pages/Account/Manage/TwoFactorAuthentication.cshtml.cs
blowdart: If we can be consistent we should :)

HaoK:
@blowdart then we should probably just remove the authenticator as part of the flow instead of just disabling the bit if we aren't going to make it possible to reuse the old authenticator key

HaoK:
Note: I ask about this because testing this is super annoying having to keep re-adding new authenticators...  I would have expected this to just be a checkbox/button where I can turn on tfa or off...

blowdart:
@grahamehorner Wait, what? What is your scenario for multiple auth apps?

blowdart:
@HaoK I end up using https://gauth.apps.gbraad.nl/ when testing :)

src/UI/Areas/Identity/Pages/Account/Manage/TwoFactorAuthentication.cshtml.cs
HaoK: I'm going to switch to using the StatusMessage consistently so all the manage views are consistently using the same partial and naming rather than mixing Banner and StatusMessage.  I'll also add some messages that will get shown at the top like

"Your authenticator app key has been reset, please reenable using the new key.";
"You have successfully added an authenticator."

HaoK:
@blowdart so is there any way we can simplify the UI when you already have an authenticator...

So today it shows both Configure and Reset which as far as I can tell basically do the same thing when you click either button....  Is there a reason we can't just only show the Reset button when they have an authenticator key set?  I guess there's a slight difference in that Reset will turn off 2fa, and Configure would leave it on until its confirmed?  But that subtle difference seems like something that would be impossible for users to know the difference between...

blowdart:
Reset should generate a new key, configure should keep the old one and allow you to rescan.

HaoK:
Can we change the words to make that clearer?

Relink your authenticator vs Attach a new authenticator or something to that effect?

HaoK:
Or really if I understand what you are saying.

Configure is => Test my authenticator (to ensure that its linked and can verify codes)
Reset => Turn off 2fa, add a new authenticator

blowdart:
I'd honestly lean towards "disable" which turns it off, and blanks the key and "set up an authenticatior app" which keeps the key and redisplays the existing information if there is a key

HaoK:
So we have a [Disable] at the top which just turns off 2fa bit
[Setup an authenticator app] which is the existing enableAuthenticator
[Disable your authenticator] which rolls the key and disables 2fa, I'll change this to redirect to the main 2fa page instead of EnableAuthenticator which is why it was kinda confusing today, both buttons end up in the same place.

I'll rename the page to DisableAuthenticator as well

HaoK:
So our current flow is kinda weird then, when you are just 'verifying' an existing authenticator, it always regenerates your recovery keys

blowdart:
Hence the tidy up :)

HaoK:
So do we want Enable + Setup instead?  Where setup just lets them check that its working?

blowdart:
And setup allows you to scan the image in a new phone, if, for example, you're replacing a phone.

But yes. 

Enable if it's not enabled, which generates a new key and then ends up in Setup, and then when you confirm it's working actually enables the feature (this would also be where that cancel feature I wanted comes into play - in the best possible flow that authenticator key should not be saved and made active until confirmation)

Setup a new app if a key already exists should display the information, allow for rescanning and confirmation.

Reset burns the key, disables 2fa, and enables it again once they've confirmed a working TOTP code.

Disable would disable 2fa and burn the key.

HaoK:
Actually so I think i see an easier fix.  We really only want to generate recovery codes the first time an authenticator was added.  So we can just check for that in EnableAuthenticator....  If RecoveryCode count == 0, generate new ones, otherwise skip that step...

blowdart:
Couldn't we just have clearer flags in the database? When did recovery codes enter into this?!

HaoK:
Exactly, there's no reason to reset anyones 2fa recovery codes, they are completely separate from authenticators...  its super weird that those would get reset when you reset your authenticator key.  I'm fixing that as part of this :)

HaoK:
I don't think disable works, we probably need to go back to Reset authenticator.  The reason being, there's no way to set AuthenticatorKey back to null, so even if you click Disable, it still shows up... which is super confusing...  Hence I think the existing Reset Authenticator Key is fine.

HaoK:
So it's going to be:

[Add new authenticator] the first time

Then forever will be:
[Setup an authenticator app] [Reset your authenticator key]

HaoK:
Recovery codes are generate the first time you add an authenticator (assuming you have 0 recovery keys), otherwise it will never generate you new codes....  they can do that themselves via the action anyways.

blowdart:
OK that makes sense, that's how it is now :)

Knock it up and then schedule 30 minutes to show me.

(And I'm sure you could make a way to set the code back to null)

HaoK:
I can't do it via identity, i can cheat and do it via the entity directly...not really what we should be doing :)

blowdart:
Add it to identity then :p

HaoK:
The whole point is that we shouldn't care what the authenticator key is...  The operations should be (reset the key), or configure an authenticator using that key.  There should be no other operations

blowdart:
It'd just be nice to tidy up if a user disables 2fa.

HaoK:
Ok this PR has grown from just adding a new button, I'll keep iterating and we can review this next week in triage and this can go in for preview2 rather than trying to cram it in today :)

blowdart:
Ok :)

grahamehorner:
@blowdart We are looking at a user base that has multiple devices, the user however when visiting some of their client have to hand-in devices that can make phone calls; so under this situation they would switch from a (protected geo-fenced & IMEI) mobile with authenticator to use a tablet device that has restricted operations with authenticator installed.

HaoK:
@pranavkm ok updated using status message everywhere, and now we set the status message in various places as well.

@blowdart feel free to start reviewing the new strings, we can review the updated flow in person next week.  The only major change is that we now no longer generate recovery codes every time through EnableAuthenticator

HaoK:
@grahamehorner I don't think our current system cares which authenticator app you use, so long as any of the apps are registered using the authenticator key for the user, they would generate the same code no?

blowdart:
@grahamehorner That's really, really specialized. At that point I'd honestly be thinking you need to move away from Identity. It's really aimed at the "Get me started and going and safe" scenarios, which cover the basic uses for 95% of our customers. We're always going to prioritize features for the majority here. You can rescan and use multiple devices using the same keys, but we're not going to add multiple seed support.

grahamehorner:
@blowdart totally understand, was just thinking multiple device support with different keys may be useful for others and if it was a relatively minor change in the flow it could be an option; as I understood that the TOTP has a modifier which can be used to indicate the device that is used to generate the TOTP codes? and it was worth a try ;) as my dads favourite saying 'shy kids get no cake!'

src/UI/Areas/Identity/Pages/Account/Manage/EnableAuthenticator.cshtml.cs
```diff
@@ -93,9 +95,18 @@ public class InputModel
...
             _logger.LogInformation("User with ID '{UserId}' has enabled 2FA with an authenticator app.", user.Id);
 
-            var recoveryCodes = await _userManager.GenerateNewTwoFactorRecoveryCodesAsync(user, 10);
-            RecoveryCodes = recoveryCodes.ToArray();
-            return RedirectToPage("./ShowRecoveryCodes");
+            StatusMessage = "Your authenticator app has been verified.";
+
+            // Only generate recovery codes the first time
+            if (await _userManager.CountRecoveryCodesAsync(user) == 0) {
+                var recoveryCodes = await _userManager.GenerateNewTwoFactorRecoveryCodesAsync(user, 10);
```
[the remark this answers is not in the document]
HaoK: Yeah this is intended, I'll just nuke the comment

src/UI/Areas/Identity/Pages/Account/Manage/Disable2fa.cshtml
```diff
@@ -5,6 +5,7 @@
     ViewData["ActivePage"] = ManageNavPages.TwoFactorAuthentication;
 }
 
+@Html.Partial("_StatusMessage", Model.StatusMessage)
```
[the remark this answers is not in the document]
HaoK: `<partial name="_StatusMessage" asp-for="StatusMessage" />` without the "Model.": right?  I'll update all of the views to use this pattern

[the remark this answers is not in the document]
HaoK: So interesting, when I make this change to Manage/Index we get an exception, while the old HtmlPartial works...  Is this indicative of a bug somewhere else?

```
The model item passed into the ViewDataDictionary is of type 'Microsoft.AspNetCore.Identity.UI.Pages.Account.Manage.IndexModel', but this ViewDataDictionary instance requires a model item of type 'System.String'.
```

HaoK:
ae127558b55d227e3b496db52db8412f4e821e30

PR #1578, opened 2018-01-20
Moving identity UI to use precompiled views

javiercn:


javiercn:
@rynowak I am not able to trigger precompilation here

src/UI/Microsoft.AspNetCore.Identity.UI.csproj
```diff
@@ -7,16 +7,25 @@
     <PackageTags>aspnetcore;identity;membership;razorpages</PackageTags>
     <EnableApiCheck>false</EnableApiCheck>
     <GenerateEmbeddedFilesManifest>true</GenerateEmbeddedFilesManifest>
+    <RazorCompileOnBuild>true</RazorCompileOnBuild>
```
rynowak: Add `<ResolvedRazorCompileToolset>RazorSdk</ResolvedRazorCompileToolset>`

This is a feature that the Web SDK has added for us, but it hasn't made it into our builds yet.

javiercn:
@rynowak Got the attributes in the assembly, will we have to do anything to support signing the precompiled views assembly?

rynowak:
We pass the same information to CSC as the main compile target does.

javiercn:
@rynowak Fair enough, everything went through correctly through our build pipeline, so I think it must be fine

PR #1579, opened 2018-01-20
Add authorization for identity UI area

javiercn:


src/UI/IdentityDefaultUIConfigureOptions.cs
```diff
@@ -30,6 +33,8 @@ public void PostConfigure(string name, RazorPagesOptions options)
             options = options ?? throw new ArgumentNullException(nameof(options));
 
             options.AllowAreas = true;
+            options.Conventions.AuthorizeAreaFolder(IdentityUIDefaultAreaName, "/Account/Manage");
+            options.Conventions.AuthorizeAreaPage(IdentityUIDefaultAreaName, "/Account/Logout");
```
HaoK: Why does logout need authorization, what happens if someone tries to logout when they aren't logged in?

PR #1586, opened 2018-01-22
Fix manage views layout

javiercn:


src/UI/Areas/Identity/Pages/Account/Manage/_ViewStart.cshtml
```diff
@@ -0,0 +1,7 @@
+﻿@{
+    if (Layout != "_Layout")
```
pranavkm: Can you give me some back ground on what this code is trying to do? It's not very clear.

javiercn: It checks if the layout has been overriden and saves the overriden layout in viewdata.
Then on the Manage/_Layout.cshtml view, it checks if an overriden layout was specified in viewdata, and if thats the case uses that as the parent layout instead of statically pointing to the one in Identity.

PR #1587, opened 2018-01-23
Update jQuery

jbagga:


jbagga:
@javiercn Could you take a look when you have time?

PR #1594, opened 2018-01-25
Fix IUserRoleStore docs

robertmiles3:
Fix some documentation inside `IUserRoleStore` that I recently noticed while working with it.


HaoK:
Thanks @robertmiles3

PR #1603, opened 2018-01-30
Add functional tests for the default identity UI

javiercn:
Added initial functional tests for the identity UI.
* Created a website using the default UI.
* Created a functional test project.
* Created the required infrastructure to write tests in a reasonable way.
* Created an initial set of tests for:
  * Authorization.
  * Registering a new user.
  * Login with a user name and a password.
  * Enabling two factor authentication.
  * Login with user name and password + two factor authentication.
  * Login with user name and password + recovery code.
  * Login with a user and an unconfirmed email (Email confirmation required)
  * Login with a user and a confirmed email (Email confirmation required)


Identity.sln
```diff
@@ -86,6 +87,10 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "IdentitySample.DefaultUI",
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.AspNetCore.Identity.UI", "src\UI\Microsoft.AspNetCore.Identity.UI.csproj", "{894E102D-56D4-4B02-8F13-8781F4324C3E}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Identity.DefaultUI.WebSite", "test\WebSites\Identity.DefaultUI.WebSite\Identity.DefaultUI.WebSite.csproj", "{CAE02AD2-F941-4ACB-B469-13EFF551BB74}"
```
HaoK: Should we just use the ui sample for the functional test?  kills two birds with one stone as we don't have to worry about the sample not working :)

test/Identity.FunctionalTests/AuthorizationTests.cs
```diff
@@ -0,0 +1,112 @@
...
+        public async Task AnonymousUserCantAccessAuthorizedPages(string url)
+        {
+            // Arrange
+            var client = ServerFactory.CreateDefaultClient();
+
+            // Act
+            var response = await client.GetAsync(url);
+
+            // Assert
+            ResponseAssert.IsRedirect(response);
```
HaoK: Should this check the redirect location is the login page?

test/Identity.FunctionalTests/LoginTests.cs
```diff
@@ -0,0 +1,151 @@
...
+            var password = $"!Test.Password1$";
+
+            var loggedIn = await UserStories.RegisterNewUserAsync(client, userName, password);
+            var showRecoveryCodes = await UserStories.EnableTwoFactorAuthentication(loggedIn, twoFactorEnabled: false);
+
+            var twoFactorKey = showRecoveryCodes.Context[EnableAuthenticator.AuthenticatorKey];
+
+            // Act & Assert
+            // Use a new client to simulate a new browser session.
+            var loginWith2fa = await UserStories.LoginExistingUser2FaAsync(newClient, userName, password, twoFactorKey);
```
HaoK: nit: don't need to remember the result here

Identity.sln
HaoK: Oh never mind, the test website is doing most of the test/asserts

HaoK: As we discussed, I think long term it would be ideal if we could point the functional tests against the sample ui site, maybe something to track as a future enhancement/issue

src/UI/Areas/Identity/Pages/Account/LoginWith2fa.cshtml
```diff
@@ -33,7 +33,7 @@
 </div>
 <p>
     Don't have access to your authenticator device? You can
-    <a asp-page="./LoginWithRecoveryCode" asp-route-returnUrl="@Model.ReturnUrl">log in with a recovery code</a>.
+    <a id="recovery-code-login" asp-page="./LoginWithRecoveryCode" asp-route-returnUrl="@Model.ReturnUrl">log in with a recovery code</a>.
```
Eilon: Interesting - I assume all these id's are added for testability so you can find the elements? Seems a'ight to me.

test/Identity.FunctionalTests/Extensions/HtmlAssert.cs
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        internal static IEnumerable<IHtmlElement> HasElements(string selector, IHtmlDocument document)
+        {
+            var elements = document
+                .QuerySelectorAll(selector)
+                .OfType<IHtmlElement>()
+                .ToArray();
+
+            Assert.True(elements.Length > 0);
```
Eilon: Is there like an Assert.NotEmpty or something?

test/Identity.FunctionalTests/AuthorizationTests.cs
```diff
@@ -0,0 +1,112 @@
...
+                //"/Identity/Account/Manage/Disable2fa",
+                "/Identity/Account/Manage/DownloadPersonalData",
+                "/Identity/Account/Manage/EnableAuthenticator",
+                "/Identity/Account/Manage/ExternalLogins",
+                //"/Identity/Account/Manage/GenerateRecoveryCodes",
+                "/Identity/Account/Manage/Index",
+                "/Identity/Account/Manage/PersonalData",
+                "/Identity/Account/Manage/ResetAuthenticator",
+                //"/Identity/Account/Manage/SetPassword",
+                //"/Identity/Account/Manage/ShowRecoveryCodes",
```
pranavkm: Commented out

test/Identity.FunctionalTests/AuthorizationTests.cs
```diff
@@ -0,0 +1,112 @@
...
+
+        // The routes commented below are not directly accessible by
+        // typing the URL in the browser. They have to be accessed as
+        // part of a more complex interation. (like login with 2fa).
+        public static TheoryData<string> RouteableAuthorizedPages =>
+            new TheoryData<string>
+            {
+                "/Identity/Account/Manage/ChangePassword",
+                "/Identity/Account/Manage/DeletePersonalData",
+                //"/Identity/Account/Manage/Disable2fa",
```
pranavkm: Why is this commented out?

test/Identity.FunctionalTests/AuthorizationTests.cs
```diff
@@ -0,0 +1,112 @@
...
+                "/Identity/Account/Manage/DownloadPersonalData",
+                "/Identity/Account/Manage/EnableAuthenticator",
+                "/Identity/Account/Manage/ExternalLogins",
+                //"/Identity/Account/Manage/GenerateRecoveryCodes",
```
pranavkm: commented out

test/Identity.FunctionalTests/AuthorizationTests.cs
```diff
@@ -0,0 +1,112 @@
...
+            ResponseAssert.IsHtmlDocument(response);
+        }
+
+        // The routes commented below are not directly accessible by
+        // typing the URL in the browser. They have to be accessed as
+        // part of a more complex interation. (like login with 2fa).
+        public static TheoryData<string> UnauthorizedPages =>
+            new TheoryData<string> {
+                "/Identity/Account/ResetPassword",
+                //"/Identity/Account/LoginWithRecoveryCode",
```
pranavkm: Bunch of more commented out urls. Can you uncomment them or remove them with a work item to add them back? Let's not leave commented out tests

test/Identity.FunctionalTests/Extensions/HtmlAssert.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using AngleSharp.Dom;
+using AngleSharp.Dom.Html;
+using AngleSharp.Parser.Html;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Identity.FunctionalTests
+{
+    public class HtmlAssert
```
pranavkm: Did you consider just subclassing from `Xunit.Assert`? https://github.com/aspnet/Razor/blob/dev/test/Microsoft.AspNetCore.Razor.Design.Test/IntegrationTests/Assert.cs#L12

test/Identity.FunctionalTests/Extensions/HtmlAssert.cs
pranavkm: There is.

test/Identity.FunctionalTests/Extensions/HtmlAssert.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            var form = Assert.Single(document.QuerySelectorAll(selector));
+            return Assert.IsAssignableFrom<IHtmlFormElement>(form);
+        }
+
+        internal static IHtmlHtmlElement IsHtmlFragment(string htmlMessage)
+        {
+            var synteticNode = $"<div>{htmlMessage}</div>";
+            var fragment = Assert.Single(new HtmlParser().ParseFragment(htmlMessage, context: null));
+            return Assert.IsAssignableFrom<IHtmlHtmlElement>(fragment);
```
pranavkm: Is the fragment an `<html>` element? Why wrap it in a div?

test/Identity.FunctionalTests/Extensions/HttpClientExtensions.cs
```diff
@@ -0,0 +1,45 @@
...
+                Content = new StreamContent(submit.Body)
+            };
+
+            foreach (var header in submit.Headers)
+            {
+                submision.Headers.TryAddWithoutValidation(header.Key, header.Value);
+                submision.Content.Headers.TryAddWithoutValidation(header.Key, header.Value);
+            }
+
+            return await client.SendAsync(submision);
```
pranavkm: Don't need the `await`

test/Identity.FunctionalTests/Extensions/ResponseAssert.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            Assert.Equal(HttpStatusCode.Redirect, responseMessage.StatusCode);
+            return responseMessage.Headers.Location;
+        }
+
+        public static IHtmlDocument IsHtmlDocument(HttpResponseMessage response)
+        {
+            Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+            Assert.Equal("text/html", response.Content.Headers.ContentType.MediaType);
+            var content = response.Content.ReadAsStringAsync().Result;
```
pranavkm: Why is this synchronous?

test/Identity.FunctionalTests/Extensions/ResponseAssert.cs
```diff
@@ -0,0 +1,56 @@
...
+            var document = BrowsingContext.New()
```
pranavkm: Why can't you do `new HtmlParser().Parse` here? The Response headers don't change the result from what I can tell.

test/Identity.FunctionalTests/Infrastructure/HtmlPageContext.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.AspNetCore.Identity.FunctionalTests
+{
+    public class HtmlPageContext
+    {
+        private readonly IDictionary<string, string> _properties = 
+            new Dictionary<string, string>();
+
+        public string this[string key]
```
pranavkm:
```C#
public string this[string key]
{
   get => _properties[key];
   set => _properties[key] = value;
}
```

test/Identity.FunctionalTests/Infrastructure/ServerFactory.cs
```diff
@@ -0,0 +1,50 @@
...
+using Identity.DefaultUI.WebSite;
+using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.TestHost;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.AspNetCore.Identity.FunctionalTests
+{
+    public class ServerFactory
+    {
+        public static TestServer CreateServer(Action<IWebHostBuilder> configureBuilder, [CallerMemberName]string isolationKey = "")
```
pranavkm: Formatting - space after `[CallerMemberName]`

test/Identity.FunctionalTests/Pages/Account/Register.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        private IHtmlFormElement _registerForm;
+
+        public Register(HttpClient client, IHtmlDocument register, HtmlPageContext context)
+            : base(client, register, context)
+        {
+            _registerForm = HtmlAssert.HasForm(register);
+        }
+
+        public async Task<Index> CreateValidUser(string userName, string password)
```
pranavkm: Why not call this `SubmitRegisterFormAsync` to keep the follow the convention established earlier e.g. `ClickEnableAuthenticatorLinkAsync`? The action is self-descriptive in this case rather than describing the side effect

test/Identity.FunctionalTests/Pages/Index.cs
```diff
@@ -0,0 +1,75 @@
...
+        private readonly IHtmlAnchorElement _manageLink;
+        public static readonly string Path = "/";
+
+        public Index(HttpClient client, IHtmlDocument index, HtmlPageContext context, bool authenticated)
+            : base(client, index, context)
+        {
+            _authenticated = authenticated;
+            if (!_authenticated)
+            {
+                _registerLink = HtmlAssert.HasLink("#register", Document);
```
pranavkm: This should ensure the opposite is true when authenticated i,e, `#register` does not appear for authenticated users

test/Identity.FunctionalTests/xunit.runner.json
```diff
@@ -0,0 +1,3 @@
+{
+  "shadowCopy": false
```
pranavkm: Btw, do you actually need this? Presumably we only need this for the deps file but we're have precompiled views.

test/Identity.FunctionalTests/AuthorizationTests.cs
javiercn: It's explained above. These urls are not directly hittable from the browser. The user gets there as part of redirect flows.

test/Identity.FunctionalTests/AuthorizationTests.cs
javiercn: It's explained above. These urls are not directly hittable from the browser. The user gets there as part of redirect flows.

test/Identity.FunctionalTests/AuthorizationTests.cs
javiercn: It's explained above. These urls are not directly hittable from the browser. The user gets there as part of redirect flows.

test/Identity.FunctionalTests/AuthorizationTests.cs
javiercn: It's explained above. These urls are not directly hittable from the browser. The user gets there as part of redirect flows.

test/Identity.FunctionalTests/Extensions/HtmlAssert.cs
javiercn: It's a fragment. It can contain multiple elements. `<span>a</span>some text<a href="blah">text</a>some other stuff`

test/Identity.FunctionalTests/Extensions/ResponseAssert.cs
javiercn: Because the browsing context sets things like the base Url and things like that. Without it, things like href values will not be correct.

test/Identity.FunctionalTests/Pages/Account/Register.cs
javiercn: The pages model a state machine. The methods are the transitions, so I want to have different methods for different cases, like registering an invalid user (returning back the register view)

test/Identity.FunctionalTests/Pages/Index.cs
javiercn: That increases the complexity of the tests a lot without providing too much value. In the case where it's worth it, we can have it, but I think it's ok to focus on the elements that should be present on the page rather than on the ones that are not present.

test/Identity.FunctionalTests/xunit.runner.json
javiercn: MVC still needs to load deps, so I rather keep it here.

test/Identity.FunctionalTests/AuthorizationTests.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        // The routes commented below are not directly accessible by
+        // typing the URL in the browser. They have to be accessed as
+        // part of a more complex interation. (like login with 2fa).
+        // /Identity/Account/LoginWithRecoveryCode
+        // /Identity/Account/LoginWith2fa
+        // /Identity/Account/ExternalLogin
+        // /Identity/Account/ConfirmEmail
+        public static TheoryData<string> UnauthorizedPages =>
+            new TheoryData<string> {
```
pranavkm: Formatting

test/Identity.FunctionalTests/Microsoft.AspNetCore.Identity.FunctionalTests.csproj
```diff
@@ -0,0 +1,28 @@
...
+
+  <ItemGroup>
+    <None Include="xunit.runner.json" CopyToOutputDirectory="PreserveNewest" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="$(MicrosoftAspNetCoreMvcTestingPackageVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Logging.Testing" Version="$(MicrosoftExtensionsLoggingTestingPackageVersion)" />
+    <PackageReference Include="Microsoft.EntityFrameworkCore.InMemory" Version="$(MicrosoftEntityFrameworkCoreInMemoryPackageVersion)" />
+    <PackageReference Include="AngleSharp" Version="$(AngleSharpPackageVersion)" />
```
pranavkm: You'll need to send a PR to universe to mirror this package. See https://github.com/aspnet/Universe/blob/dev/build/external-dependencies.props#L220

javiercn:
ac3bea16fdf6bb44f4811d327482c49d94d84cd2

PR #1615, opened 2018-02-05
Refresh user signin after updating their profile

jbagga:
See https://github.com/aspnet/templating/pull/199

Identity code has moved to this repo from templating

cc @Eilon

### DIFF
--- a/samples/IdentitySample.DefaultUI/Areas/Identity/Pages/Account/Manage/Index.cshtml.cs
+++ b/samples/IdentitySample.DefaultUI/Areas/Identity/Pages/Account/Manage/Index.cshtml.cs
@@ -129,6 +129,7 @@ namespace IdentitySample.DefaultUI
                 }
             }
 
+            await _signInManager.RefreshSignInAsync(user);
             StatusMessage = "Your profile has been updated";
             return RedirectToPage();
         }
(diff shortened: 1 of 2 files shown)


HaoK:
One small question, is there any reason we prefer RefreshSignIn over just SignIn here?

Eilon:
@HaoK you're the expert, you tell us 😄

HaoK:
Yeah so looks like we added this feature specifically for the templates to use, but never updated them :)

So we should update the rest of the pages in the default UI that are currently using SignIn to use RefreshSignIn

See: https://github.com/aspnet/Identity/blob/dev/src/UI/Areas/Identity/Pages/Account/Manage/SetPassword.cshtml.cs#L94

Today they are hardcoding false after you change the password, which downgrades any persistent remember me cookies to a sesion one.

jbagga:
@HaoK @Eilon I think I should file a new work item for that. It will need some testing

HaoK:
@jbagga sounds fine, the general issue can be summed up as Manage actions lose cookie info (persistent/external login claim) when they call SignInAsync instead of RefreshSignIn

jbagga:
Filed [here](https://github.com/aspnet/Identity/issues/1616)

PR #1623, opened 2018-02-12
Release/2.0

oferns:
This PR removes the unneeded dependency on TUser from the IPasswordValidator interface.


HaoK:
This dependency was intentional, so the validator can access any properties on TUser.  The default implementation doesn't use this, but custom implementations would be able to access user data this way.

oferns:
Thanks for getting back to me. I suppose I can see the reasoning behind that

PR #1631, opened 2018-02-15
Use Razor Sdk from CLI

pranavkm:


src/UI/Microsoft.AspNetCore.Identity.UI.csproj
```diff
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk;Microsoft.NET.Sdk.Razor">
```
rynowak: You shouldn't need both of these right? Or did we forget to include the net sdk 😢

src/UI/Microsoft.AspNetCore.Identity.UI.csproj
```diff
@@ -10,6 +10,7 @@
     <GenerateEmbeddedFilesManifest>true</GenerateEmbeddedFilesManifest>
     <RazorCompileOnBuild>true</RazorCompileOnBuild>
     <ResolvedRazorCompileToolset>RazorSdk</ResolvedRazorCompileToolset>
```
rynowak: Is this still required? We might already have a bug tracking that.

src/UI/Microsoft.AspNetCore.Identity.UI.csproj
pranavkm: Yeah, I guess we missed covering this in our take-over from Web.Sdk thing. We'd have to effectively move this from line from https://github.com/aspnet/websdk/blob/master/src/Web/Microsoft.NET.Sdk.Web.Targets/Sdk.props#L20 to Razor.Sdk

src/UI/Microsoft.AspNetCore.Identity.UI.csproj
pranavkm: It's hidden behind the flag that needs to be cleaned up. Need one more round of stuff to get in to the CLI to sort those out.

rynowak:
Consider me signed off on this then. We shouldn't block on these issues, but we do need to fix them.

PR #1640, opened 2018-02-16
BuildNumber on ExperimentalVersionSuffix

ryanbrandenburg:
Merging to fix build.


natemcmaster:
Same feed back here as on aspnet/Hosting. This isn't correct

Issue #1639, opened 2018-02-16
Reset password doesn't work


javiercn:
* The ResetPassword view doesn't send the code as a hidden field of the reset password form, so it doesn't get modelbound.
* The reset password handlers are incorrectly defined on the ResetPasswordModel and the behavior is as if they weren't there.

PR #1642, opened 2018-02-18
[Fixes #1638, #1639] UI fixes + tests for social login

### LINKED ISSUE #1638: Send confirmation email doesn't generate the callback URI correctly
* On the manage/index page. The send email link sends an email confirmation message with an incorrect callback URI of id=<<user-id>> instead of userId=<<userId>> causing email confirmation to fail.
* The email confirmation handlers redirects the user using RedirectToPage("/Index") which is not correct for MVC based pages.

javiercn:
* Added a "fake social login provider"
* Refactored test infrastructure.
* Added tests for social login.
* Fixed bug in email confirmation through the management UI.
* Fixed bug in password reset.


src/UI/Areas/Identity/Pages/Account/ResetPassword.cshtml.cs
```diff
@@ -31,12 +31,13 @@ public class InputModel
...
             public string ConfirmPassword { get; set; }
 
+            [Required]
             public string Code { get; set; }
 
-            public virtual IActionResult OnGet(string code = null) => throw new NotImplementedException();
-
-            public virtual Task<IActionResult> OnPostAsync() => throw new NotImplementedException();
         }
+        public virtual IActionResult OnGet(string code = null) => throw new NotImplementedException();
```
HaoK: Is there any way to detect this defect via some kind of reflection test?  Is it safe to say that every page in the default UI should have these two as public virtuals?

src/UI/Areas/Identity/Pages/Account/Manage/Index.cshtml.cs
```diff
@@ -132,7 +132,7 @@ internal class IndexModel<TUser> : IndexModel where TUser : IdentityUser
             var callbackUrl = Url.Page(
                 "/Account/ConfirmEmail",
                 pageHandler: null,
-                values: new { user.Id, code },
+                values: new { userId = user.Id, code = code },
```
HaoK: Was this a root cause of one of the bugs?

src/UI/Areas/Identity/Pages/Account/ResetPassword.cshtml.cs
javiercn: I'm trying to find a way of detecting this at run time, but I haven't come up with something that is not "flaky". I'll chat with Ryan about it.

src/UI/Areas/Identity/Pages/Account/Manage/Index.cshtml.cs
javiercn: Yes, the parameter being sent was not correct.

Issue #1617, opened 2018-02-07
Razor Pages individual auth template page LoginModel missing    [AllowAnonymous]


jbagga:
_From @Rick-Anderson on January 18, 2018 4:51_

To Repo:

1.  Create a RP (Razor Pages) app with individual user accounts.
1.  Add  [RequireAuthenticatedUser](https://github.com/Rick-Anderson/QS2_long2/blob/master/QS2long/Startup.cs#L46-L59) to StartUp.
1.  Add [AllowAnonymous] to *Pages/Index.cshtml.cs*

Run the app. Tap Register/Login/About and it trys to redirect you to the Login page, but the LoginModel doesn't allow anon, so it redirects recursively and results in the following error:

HTTP Error 404.15 - Not Found
The request filtering module is configured to deny a request where the query string is too long.
Most likely causes:
•         Request filtering is configured on the Web server to deny the request because the query string is too long.

[Repo app here.](https://github.com/Rick-Anderson/QS2_long2)

Fix is to add `[AllowAnonymous]` to the Login/Register `PageModel`.

```
[AllowAnonymous]
 public class LoginModel : PageModel
```

_Copied from original issue: aspnet/templating#238_

jbagga:
_From @Tratcher on January 24, 2018 16:15_

Infinite redirect


Module | RequestFilteringModule
-- | --
Notification | BeginRequest
Handler | aspNetCore
Error   Code | 0x00000000
[...]

PR #1644, opened 2018-02-21
Add AllowAnonymous for some Identity razor pages

jbagga:
Addresses #1617


jbagga:
Fixing tests. Wanted a quick OK for the files the attribute was added to

HaoK:
Looks fine, we are adding/updating some of the functional tests to verify that these new attributes work right?

jbagga:
@javiercn @HaoK Updated with functional tests

Issue #1616, opened 2018-02-05
Replace SignIn with RefreshSignIn in UI 


jbagga:
> Manage actions lose cookie info (persistent/external login claim) when they call SignInAsync instead of RefreshSignIn

See https://github.com/aspnet/Identity/pull/1615#issuecomment-363236876

cc @HaoK

HaoK:
Cool, are you planning on fixing this @jbagga ?

jbagga:
I'll let it be triaged first. @Eilon

Eilon:
Yeah seems like a good idea.

jbagga:
Assigning cost: S as I expect to wrap this up within a day or two.

PR #1655, opened 2018-03-02
Use RefreshSignInAsync

jbagga:
Addresses #1616

### DIFF
--- a/test/Identity.FunctionalTests/ManagementTests.cs
+++ b/test/Identity.FunctionalTests/ManagementTests.cs
@@ -92,7 +92,7 @@ namespace Microsoft.AspNetCore.Identity.FunctionalTests
 
         private string GetClaimValue(ClaimsPrincipal principal)
         {
-            return principal.Identities.First().Claims.Last().Value;
+            return principal.Identities.Single().Claims.Single(c => c.Type == "AspNet.Identity.SecurityStamp").Value;
         }
 
         [Fact]
--- a/test/Identity.FunctionalTests/ManagementTests.cs
+++ b/test/Identity.FunctionalTests/ManagementTests.cs
@@ -63,36 +63,70 @@ namespace Microsoft.AspNetCore.Identity.FunctionalTests
             // Arrange
             var principals = new List<ClaimsPrincipal>();
             var server = ServerFactory.CreateServer(builder =>
-                builder.ConfigureTestServices(s => s.SetupGetUserClaimsPrincipal(user => principals.Add(user))));
+                builder.ConfigureTestServices(s => s.SetupGetUserClaimsPrincipal(user => principals.Add(user), IdentityConstants.ApplicationScheme)));
 
             var client = ServerFactory.CreateDefaultClient(server);
             var newClient = ServerFactory.CreateDefaultClient(server);
 
             var userName = $"{Guid.NewGuid()}@example.com";
-            var password = $"!Test.Password1$";
+            var password = "!Test.Password1";
 
             var index = await UserStories.RegisterNewUserAsync(client, userName, password);
-            var manage = await index.ClickManageLinkAsync();
-            var changePassword = await manage.ClickChangePasswordLinkAsync();
 
             // Act 1
-            var changedPassword = await UserStories.ChangePasswordAsync(changePassword);
+            var changedPassword = await UserStories.ChangePasswordAsync(index, "!Test.Password1", "!Test.Password2");
 
             // Assert 1
-            // RefreshSignIn regenerates a new ClaimsPrincipal
-            Assert.NotEqual(GetClaimValue(principals[0]), GetClaimValue(principals[1]));
+            // RefreshSignIn generates a new security stamp claim
+            Assert.NotEqual(GetSecurityStampClaimValue(principals[0]), GetSecurityStampClaimValue(principals[1]));
 
             // Act 2
-            await UserStories.LoginExistingUserAsync(newClient, userName, $"!Test.Password2$");
+            await UserStories.LoginExistingUserAsync(newClient, userName, "!Test.Password2");
 
             // Assert 2
-            // Signing in again with a different client uses the same ClaimsPrincipal
-            Assert.Equal(GetClaimValue(principals[1]), GetClaimValue(principals[2]));
+            // Signing in again with a different client uses the same security stamp claim
+            Assert.Equal(GetSecurityStampClaimValue(principals[1]), GetSecurityStampClaimValue(principals[2]));
         }
 
-        private string GetClaimValue(ClaimsPrincipal principal)
+        [Fact]
+        public async Task CanChangePasswordWithExternalLogin()
+        {
+            // Arrange
+            var principals = new List<ClaimsPrincipal>();
+            var server = ServerFactory.CreateServer(builder =>
+                builder.ConfigureTestServices(s => s.SetupTestThirdPartyLogin().SetupGetUserClaimsPrincipal(user => principals.Add(user), IdentityConstants.ExternalScheme)));
+
+            var client = ServerFactory.CreateDefaultClient(server);
+            var newClient = ServerFactory.CreateDefaultClient(server);
+
+            var guid = Guid.NewGuid();
+            var userName = $"{guid}";
+            var email = $"{guid}@example.com";
+
+            // Act 1
+            var index = await UserStories.RegisterNewUserWithSocialLoginAsync(client, userName, email);
+            var claim = GetNameIdentifierClaimValue(principals[0]);
+
+            // Assert 1
+            Assert.NotNull(claim);
+
+            // Act 2
+            var changedPassword = await UserStories.ChangePasswordExternalLoginAsync(index, "!Test.Password2");
+            await UserStories.LoginExistingUserAsync(newClient, email, "!Test.Password2");
+
+            // Assert 2
+            // Signing in again with a different client uses the same name identifier claim
+            Assert.Equal(claim, GetNameIdentifierClaimValue(principals[0]));
+        }
+
+        private string GetSecurityStampClaimValue(ClaimsPrincipal principal)
         {
             return principal.Identities.Single().Claims.Single(c => c.Type == "AspNet.Identity.SecurityStamp").Value;
+        }
+
+        private string GetNameIdentifierClaimValue(ClaimsPrincipal principal)
+        {
+            return principal.Identities.Single().Claims.Single(c => c.Type == ClaimTypes.NameIdentifier).Value;
         }
 
         [Fact]
--- a/test/Identity.FunctionalTests/ManagementTests.cs
+++ b/test/Identity.FunctionalTests/ManagementTests.cs
@@ -4,6 +4,10 @@
 using System;
 using System.Threading.Tasks;
 using Identity.DefaultUI.WebSite;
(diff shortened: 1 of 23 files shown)


test/Identity.FunctionalTests/ManagementTests.cs
```diff
@@ -52,5 +56,25 @@ public ManagementTests(ServerFactory serverFactory)
...
+            var userName = $"{Guid.NewGuid()}@example.com";
+            var password = $"!Test.Password1$";
+
+            var index = await UserStories.RegisterNewUserAsync(client, userName, password);
+            var manage = await index.ClickManageLinkAsync();
+            var changePassword = await manage.ClickChangePasswordLinkAsync();
+
+            var changedPassword = await UserStories.ChangePasswordAsync(changePassword);
+            await UserStories.LoginExistingUserAsync(newClient, userName, $"!Test.Password2$");
+        }
```
jbagga: @HaoK Any suggestions on how to verify the user claim? Or how to check for a new cookie? (any other way to test using RefreshSignInAsync)

HaoK: I think the easiest thing to do is add a new page to https://github.com/aspnet/Identity/tree/dev/test/WebSites/Identity.DefaultUI.WebSite/Pages which just prints out the claim so you can confirm that its still there...

jbagga:
@HaoK Updated. 

cc @javiercn

jbagga:
@HaoK Updated

HaoK:
Cool I think the new test looks good now, can you see if its easy to add one more test flow which should now be fixed by the change to refresh token?

Can you login via an external login, i.e. this test https://github.com/aspnet/Identity/blob/dev/test/Identity.FunctionalTests/LoginTests.cs#L180

Assert that this identity has an External claim that lists the external provider scheme in the authenticatio method: https://github.com/aspnet/Identity/blob/dev/src/Identity/SignInManager.cs#L192

Change the password and verify that the authentication method claim remains.

That was the original bug that was motivating the switch to refresh sign in.

jbagga:
@HaoK Logging in with an external provider as in the test you linked to https://github.com/aspnet/Identity/blob/dev/test/Identity.FunctionalTests/LoginTests.cs#L180 adds ClaimTypes.NameIdentifier to Claims for the Principal (not AuthenticationMethod).

test/Identity.FunctionalTests/ManagementTests.cs
```diff
@@ -52,5 +56,43 @@ public ManagementTests(ServerFactory serverFactory)
...
+
+            var client = ServerFactory.CreateDefaultClient(server);
+            var newClient = ServerFactory.CreateDefaultClient(server);
+
+            var userName = $"{Guid.NewGuid()}@example.com";
+            var password = $"!Test.Password1$";
+
+            var index = await UserStories.RegisterNewUserAsync(client, userName, password);
+            var manage = await index.ClickManageLinkAsync();
+            var changePassword = await manage.ClickChangePasswordLinkAsync();
```
javiercn: This belongs inside UserStories.ChangePasswordAsync

test/Identity.FunctionalTests/UserStories.cs
```diff
@@ -135,5 +135,10 @@ public class UserStories
             var resetPassword = await ResetPassword.CreateAsync(link, client, new DefaultUIContext().WithExistingUser());
             return await resetPassword.SendNewPasswordAsync(email, newPassword);
         }
+
+        internal static async Task<ChangePassword> ChangePasswordAsync(ChangePassword changePassword)
+        {
+            return await changePassword.ChangePasswordAsync($"!Test.Password1$", $"!Test.Password2$");
```
javiercn: A few small comments: 
* Pass in the old password and the new password as parameters.

test/Identity.FunctionalTests/ManagementTests.cs
```diff
@@ -52,5 +56,43 @@ public ManagementTests(ServerFactory serverFactory)
...
+
+            // Assert 1
+            // RefreshSignIn regenerates a new ClaimsPrincipal
+            Assert.NotEqual(GetClaimValue(principals[0]), GetClaimValue(principals[1]));
+
+            // Act 2
+            await UserStories.LoginExistingUserAsync(newClient, userName, $"!Test.Password2$");
+
+            // Assert 2
+            // Signing in again with a different client uses the same ClaimsPrincipal
```
javiercn: nit: The comment is a bit inaccurate. The Security timestamp claim is the only thing that is equal.

test/Identity.FunctionalTests/ManagementTests.cs
```diff
@@ -52,5 +56,43 @@ public ManagementTests(ServerFactory serverFactory)
...
+
+            var index = await UserStories.RegisterNewUserAsync(client, userName, password);
+            var manage = await index.ClickManageLinkAsync();
+            var changePassword = await manage.ClickChangePasswordLinkAsync();
+
+            // Act 1
+            var changedPassword = await UserStories.ChangePasswordAsync(changePassword);
+
+            // Assert 1
+            // RefreshSignIn regenerates a new ClaimsPrincipal
```
javiercn: nit: RefreshSignIn generates a new security stamp

jbagga:
Added a test for SetPassword and the scenario @HaoK described.


WIP adding tests for removing external login and resetting authenticator.

test/Identity.FunctionalTests/UserStories.cs
```diff
@@ -136,9 +136,20 @@ public class UserStories
...
         {
-            return await changePassword.ChangePasswordAsync($"!Test.Password1$", $"!Test.Password2$");
+            var manage = await index.ClickManageLinkAsync();
+            var changePassword = await manage.ClickChangePasswordLinkAsync();
+
+            return await changePassword.ChangePasswordAsync(oldPassword, newPassword);
+        }
+
+        internal static async Task<SetPassword> ChangePasswordExternalLoginAsync(Index index, string newPassword)
+        {
```
javiercn: This would be more like set account pasword async or something similar?

test/Identity.FunctionalTests/ManagementTests.cs
```diff
@@ -52,5 +56,77 @@ public ManagementTests(ServerFactory serverFactory)
...
+            // Act 2
+            await UserStories.LoginExistingUserAsync(newClient, userName, "!Test.Password2");
+
+            // Assert 2
+            // Signing in again with a different client uses the same security stamp claim
+            Assert.Equal(GetSecurityStampClaimValue(principals[1]), GetSecurityStampClaimValue(principals[2]));
+        }
+
+        [Fact]
+        public async Task CanChangePasswordWithExternalLogin()
```
javiercn: CanSetPasswordWithExternalLogin

test/Identity.FunctionalTests/ManagementTests.cs
```diff
@@ -52,5 +56,77 @@ public ManagementTests(ServerFactory serverFactory)
...
+            // Assert 1
+            Assert.NotNull(claim);
+
+            // Act 2
+            var changedPassword = await UserStories.ChangePasswordExternalLoginAsync(index, "!Test.Password2");
+            await UserStories.LoginExistingUserAsync(newClient, email, "!Test.Password2");
+
+            // Assert 2
+            // Signing in again with a different client uses the same name identifier claim
+            Assert.Equal(claim, GetNameIdentifierClaimValue(principals[0]));
```
javiercn: What does the name has to be with these changes? Don't you have to be checking the security timestamp?

test/Identity.FunctionalTests/ManagementTests.cs
```diff
@@ -52,5 +56,77 @@ public ManagementTests(ServerFactory serverFactory)
...
+            Assert.Equal(GetSecurityStampClaimValue(principals[1]), GetSecurityStampClaimValue(principals[2]));
+        }
+
+        [Fact]
+        public async Task CanChangePasswordWithExternalLogin()
+        {
+            // Arrange
+            var principals = new List<ClaimsPrincipal>();
+            var server = ServerFactory.CreateServer(builder =>
+                builder.ConfigureTestServices(s => s.SetupTestThirdPartyLogin().SetupGetUserClaimsPrincipal(user => principals.Add(user), IdentityConstants.ExternalScheme)));
```
javiercn: Can you break this into multiple lines?

test/Identity.FunctionalTests/ManagementTests.cs
javiercn: Or just CanSetPassword

test/Identity.FunctionalTests/ManagementTests.cs
HaoK: I think this is because the test is tracking the external cookie due to .SetupGetUserClaimsPrincipal(user => principals.Add(user), IdentityConstants.ExternalScheme) isntead of the normal application cookie

HaoK:
So the behavior we should be seeing in my scenario is, after login with an external provider, you should get an additional AuthenticationMethod claim set to the scheme of the external provider.

That will be lost when SignInAsync is called, with the switch to Refresh, that claim should be preserved in the main application cookie.

Its possible that the test I linked isn't properly setup for this behavior to be tested so you might need to mock something else.

HaoK:
If its too hard to mock, you can just add a new page that checks for that claim and write out success or not based on whether HttpContext.User.Claims.Where(c => c.Type == "AuthenticationMethod") is found, and just hit that page

javiercn:
@haok The recommended approach is not to create new pages. Observing the principals created on every signin should be easy enough by just plugging an event handler for the cookie signin method of each of the cookies that captures the principal at the time of sign in and then do the checks for the appropiate values on each of them.

HaoK:
That is cleaner I guess, downside is it took me a lot longer to figure out what these tests were trying to doing since it reads like a bunch of funcs checking claims are equal...  but that could just be a naming thing or comments...

test/Identity.FunctionalTests/ManagementTests.cs
```diff
@@ -53,6 +57,78 @@ public ManagementTests(ServerFactory serverFactory)
...
+            // Assert 1
+            Assert.NotNull(GetAuthenticationMethodIdentifierClaimValue(principals[1]));
+
+            // Act 2
+            var changedPassword = await UserStories.SetPasswordAsync(index, "!Test.Password2");
+
+            // Assert 2
+            // RefreshSignIn uses the same AuthenticationMethod claim value
+            Assert.Equal(GetAuthenticationMethodIdentifierClaimValue(principals[1]), GetAuthenticationMethodIdentifierClaimValue(principals[2]));
+        }
```
jbagga: I had to sign in after registering to get this claim. @HaoK Is this sufficient? I will add tests for other scenarios tomorrow.

HaoK: Yeah that looks fine

minor nit: maybe cleaner if you refactor the helper method to a single one like this: `AssertClaimsEqual(principals[1], principals[2], "AspNet.Identity.SecurityStamp"`)

jbagga: There will be another helper method `AssertClaimsNotEqual` but it is a lot cleaner. Thanks

jbagga:
@HaoK @javiercn Updated

PR #1656, opened 2018-03-02
Add link to Identity.Firebase

aguacongas:
In response to @Elion comment in this discussion https://github.com/aspnet/Home/issues/2916#issuecomment-370019674


HaoK:
Cool thanks!

PR #1680, opened 2018-03-14
Use ApplicationPartFactory and Sdk features in IdentityUI

pranavkm:


src/UI/IdentityBuilderUIExtensions.cs
```diff
@@ -44,63 +40,18 @@ public static IdentityBuilder AddDefaultUI(this IdentityBuilder builder)
 
         private static void AddAdditionalApplicationParts(IdentityBuilder builder)
```
pranavkm: Ignore the version changes \ new package references etc that were made. The intent of the PR is to ensure the design of the Razor.Sdk, ApplicationPartFactory is as we need. I verified that the functional tests in the app work and we code gen the same set of attributes that we previously did.

src/UI/Microsoft.AspNetCore.Identity.UI.csproj
```diff
@@ -8,9 +8,7 @@
     <PackageTags>aspnetcore;identity;membership;razorpages</PackageTags>
     <EnableApiCheck>false</EnableApiCheck>
     <GenerateEmbeddedFilesManifest>true</GenerateEmbeddedFilesManifest>
-    <RazorCompileOnBuild>true</RazorCompileOnBuild>
-    <PrecompiledAssemblyTitle Condition=" '$(PrecompiledAssemblyTitle)' == '' ">$(RazorTargetName).dll</PrecompiledAssemblyTitle>
-    <PreserveCompilationContext>false</PreserveCompilationContext>
+    <EnableDefaultCompiledViewAssemblyLoadBehavior>false</EnableDefaultCompiledViewAssemblyLoadBehavior>
```
pranavkm: Property name

PR #1681, opened 2018-03-14
Create or update baselines to match 2.0

ryanbrandenburg:


javiercn:
@ryanbrandenburg I think there might be libraries that are missing baselines in here. @HaoK

src/AspNetCoreCompat/baseline.netframework.json
```diff
@@ -1,5 +1,5 @@
 {
-  "AssemblyIdentity": "Microsoft.AspNet.Identity.AspNetCoreCompat, Version=0.1.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60",
+  "AssemblyIdentity": "Microsoft.AspNet.Identity.AspNetCoreCompat, Version=0.3.2.0, Culture=neutral, PublicKeyToken=adb9793829ddae60",
```
HaoK: Do we need to baseline prerelease non shipping alphas?

HaoK:
src/Core and src/Stores need baselines as well I think those are the main two that are missing baselines that shipped and are public in 2.0

ryanbrandenburg:
:up::date:

Issue #1625, opened 2018-02-13
Add: AddDefaultIdentity()


HaoK:
Based on discussions with @davidfowl 

Consider something like:

```
   services.AddIdentity<User, Role, DbContext>()
```

For auth consider adding a new options class with 4 `IOptionsBuilder` properties that allow configuration of each cookie.

cc @ajcvickers @blowdart

HaoK:
Met with @davidfowl @Eilon @DamianEdwards and we decided to add a new

`AddDefaultIdentity<TUser>()` method which will collapse today's identity methods down except for the stores, roles will no longer be enabled by default as well.

Startup in the templates will now look like:
```
   services.AddDefaultIdentity<IdentityUser>.AddEntityFrameworkStores<ApplicationDbContext>();
```

Its will be functionally the same as:
```
            services.AddIdentityCore<IdentityUser>(o => o.Stores.MaxLengthForKeys = 128)
                 .AddEntityFrameworkStores<ApplicationDbContext>()
                 .AddDefaultUI()
                 .AddDefaultTokenProviders();

            services.AddAuthentication(o =>
            {
                o.DefaultScheme = IdentityConstants.ApplicationScheme;
                o.DefaultSignInScheme = IdentityConstants.ExternalScheme;
            })
            .AddIdentityCookies(o => { });
```

The rest of auth changes will be looked at in 2.2

raffaeler:
Could you please confirm that RTM implementation does not require the following anymore?
```
services.AddAuthentication(o =>
            {
                o.DefaultScheme = IdentityConstants.ApplicationScheme;
                o.DefaultSignInScheme = IdentityConstants.ExternalScheme;
            })
            .AddIdentityCookies(o => { });
```
I get an error of scheme being already registered...

HaoK:
You need to add roles back, it isn't enabled by default anymore `.AddRoles<IdentityRole>`

Rick-Anderson:
>You need to add roles back, it isn't enabled by default anymore `.AddRoles<IdentityRole>`

This was fixed in #1832 so roles are now added by by `AddDefaultIdentity`
See https://github.com/aspnet/Identity/issues/1813

ntimmerman:
This is a joke.

PR #1694, opened 2018-03-19
Add AddDefaultIdentity() method

HaoK:
- Fixes https://github.com/aspnet/Identity/issues/1625
- Adds new template sugar method (AddIdentityV2) basically.
- Roles are now not included by default, can be easily added back via `.AddRoles<IdentityRole>()`
- Note this has to live in the UI package since that's the leaf package that has everything.
- Still clobbers the authentication schemes to point to identity cookies, we'll revisit that in 2.2
- Rolls the MaxKeyLength into a hidden default in the sugar method (should we set it only in the args less overload?)

@ajcvickers @blowdart @javiercn

### DIFF
--- a/samples/IdentitySample.DefaultUI/Startup.cs
+++ b/samples/IdentitySample.DefaultUI/Startup.cs
@@ -34,18 +34,9 @@ namespace IdentitySample.DefaultUI
 
             services.AddMvc();
 
-            services.AddIdentityCore<ApplicationUser>(o => o.Stores.MaxLengthForKeys = 128)
+            services.AddDefaultIdentity<ApplicationUser>()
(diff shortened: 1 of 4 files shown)


javiercn:
+@DamianEdwards, @prafullbhosale for awareness.

I'm fine with it so long as you guys are.

javiercn:
How do you go from DefaultUI to Scaffolded UI?

HaoK:
You would use AddIdentity or AddIdentityCore just like before, this is just a online liner that combines up the 3 AddDefaults today.

HaoK:
Merged to release/2.1 and merged back to dev 167a108f4ef5d95adc1f4264ee22ce0754f7de0c

PR #1701, opened 2018-03-21
React to application part changes

pranavkm:


src/UI/IdentityBuilderUIExtensions.cs
```diff
@@ -51,56 +47,19 @@ private static void AddAdditionalApplicationParts(IdentityBuilder builder)
...
-                        var fileName = Path.GetFileName(reference);
-                        var filePath = Path.Combine(Path.GetDirectoryName(thisAssembly.Location), fileName);
-                        var additionalAssembly = LoadAssembly(filePath);
-                        // This needs to change to additional assembly part.
-                        var additionalPart = new AssemblyPart(additionalAssembly);
-                        if (!apm.ApplicationParts.Any(ap => HasSameName(ap.Name, additionalPart.Name)))
-                        {
-                            apm.ApplicationParts.Add(additionalPart);
-                        }
+                        partManager.ApplicationParts.Add(new CompiledRazorAssemblyPart(assembly));
```
javiercn: This should be just using the part factory on the related assembly with a different context to light it up 😕. Something along the lines of:
```csharp
var partFactory = ProvidePartFactory.GetPartFactory(assembly);
var parts = partFactory.CreateApplicationParts(assembly,"someCtx");
apm.ApplicationParts.Add(parts)
```

pranavkm: Ok. So don't use the `NullPartFactory` and write a custom part factory?

javiercn: Yes.

Orthogonal to this, maybe NullPartFactory should be -> OptionalPartFactory and support this scenario through the ctx parameter.

javiercn: The important bit is that the parts for an assembly always have to be instantiated through the factory in the assembly so that if something changes it only changes in one place.

pranavkm: > The important bit is that the parts for an assembly always have to be instantiated through the factory 

Why is this important?

javiercn: Because if not, when something changes in the parts implementation you need to change it in all the consuming clients. For example, if for some reason precompilation changes and starts creating an additional part for something else, then all consuming clients (identity, etc) need to change to create that new part themselves.

pranavkm:
Will make changes in a follow up.

javiercn:
@pranavkm Can you confirm that you are changing this as we discussed?

Issue #1168, opened 2017-04-20
Reenable API check


HaoK:
disabled to unblock build to get Auth 2.0 build unblocked

HaoK:
Revert 2b343726897b2cece524537d56aea4c80ac265bd and updated baseline

Eilon:
@HaoK / @ajcvickers - needed for preview2, or revisit in 2.1.0-rc1? @ryanbrandenburg - does your work with ApiCheck cover everything here anyway?

HaoK:
@ryanbrandenburg already updated the baselines, so I think we just need to turn EnableApiCheck on again...I'll do that today and hopefully nothing fails :)

HaoK:
@ryanbrandenburg @javiercn what do we want to do about the ui package, should we wait until closer to rtm to generate baselines + enable api check for this since its new in 2.1?

Eilon:
I think the new packages, such as the UI ones, should get *empty* baselines.

Eilon:
(That way we know it needs to be checked, but there will be no breaking changes detected, because everything is an 'add'.)

HaoK:
So getting one breaking change with constants moving from protected to public...

     public class Microsoft.AspNetCore.Identity.UserManager<T0> : System.IDisposable where T0 : class
          protected const System.String ChangePhoneNumberTokenPurpose = "ChangePhoneNumber"
          protected const System.String ConfirmEmailTokenPurpose = "EmailConfirmation"
          protected const System.String ResetPasswordTokenPurpose = "ResetPassword"



  Following is the list of exclusions that either need to be added to the list of breaking changes, or the breaking changes themselves need to be reverted:
  [
    {
      "TypeId": "public class Microsoft.AspNetCore.Identity.UserManager<T0> : System.IDisposable where T0 : class",
      "MemberId": "protected const System.String ChangePhoneNumberTokenPurpose = \"ChangePhoneNumber\"",
      "Kind": "Removal"
    },
    {
      "TypeId": "public class Microsoft.AspNetCore.Identity.UserManager<T0> : System.IDisposable where T0 : class",
      "MemberId": "protected const System.String ConfirmEmailTokenPurpose = \"EmailConfirmation\"",
      "Kind": "Removal"
    },
    {
      "TypeId": "public class Microsoft.AspNetCore.Identity.UserManager<T0> : System.IDisposable where T0 : class",
[...]

PR #1713, opened 2018-03-27
Reenable APICheck

HaoK:
Fixes https://github.com/aspnet/Identity/issues/1168


src/Core/breakingchanges.netcore.json
```diff
@@ -0,0 +1,17 @@
+  [
+    {
```
Eilon: FYI to others: this is due to a limitation in ApiCheck; these are not actual breaking changes.

HaoK:
@ryanbrandenburg what causes those errors?  I see those locally too for release/2.1 (but not dev)

ryanbrandenburg:
Fix the version names to match something that both works and is available on one of your sources.

HaoK:
@ryanbrandenburg will these failures also happen on the CI or is this just local/pr verification failures?

ryanbrandenburg:
Possibly the referenced version fell off the nuget retention. Anyone except Universe that builds this will be broken, please update the deps by running ".\run.ps1 -update upgrade deps" and commit that change so people can use the repo.

HaoK:
Fair enough, updating deps and will merge that with this PR

Issue #626, opened 2015-11-06
GUIDs for security stamp


brockallen:
So a GUID is being used for the security stamp. On windows that has ~96 bits of entropy (according to @GrabYourPitchforks), but on linux there's no guarantee (unless that's changed and something in CoreCLR addresses this). So I worry that somehow the lack of entropy could be used to abuse ASP.NET Identity if/when running on linux, and perhaps crypto RNG is better.

If this is still an issue, then it will be true of any DNX code (perhaps in ~/security repo as well -- i've not checked if there is any Guid.NewGuid code being used there).

GrabYourPitchforks:
They actually have 122 bits of entropy. Regardless, as of the last time I looked at the code base, the security stamp doesn't need to be secret or unpredictable. It's used simply to version existing payloads like refresh tokens or reset tokens. If the security stamp in the incoming refresh token doesn't match the security stamp in the DB, then the request should be rejected. The refresh token in the incoming request should be integrity-protected (the DataProtection APIs take care of this automatically) so that the client can't mess with the embedded security stamp.

brockallen:
Just raising it as an issue to be considered given how the guid is being used.

As for:

> the security stamp doesn't need to be secret or unpredictable

IIRC, the security stamp is the main input to the TOTP (as well as time, but that's well known-ish for 2FA given the steps) and thus if an attacker has just read-only access to the DB they can predict the SMS code for 2FA. I guess in theory if the security stamp is also predictable then it's also potentially a vector.

I'm sure it's a remote vector and I'm certainly not motivated enough to work at cracking it, but someone else might be :)

Again, just mentioning it so someone smarter than me can think about it.

GrabYourPitchforks:
Ah, I thought there was an additional source of entropy for the TOTP code paths. This is something @HaoK and @blowdart should look at.

brockallen:
Same issue exists in AspId v2 :)

blowdart:
Darnit Linux :) Yea, we should switch to an CNG based random array of bytes, of the maximum size the TOTP algorithm uses. @HaoK

HaoK:
Are we backloging this @blowdart ?

blowdart:
I guess so, it's too big a change to make now unfortunately.

PinpointTownes:
A big change? Heh, why?

[...]

PR #1714, opened 2018-03-27
Update new security stamp logic

HaoK:
Fixes https://github.com/aspnet/Identity/issues/626

This doesn't really have any visible impact since we treated the security stamp as an opaque string anywyas.


HaoK:
@blowdart can you review this one since its mostly a security review that this looks like a 'legit' thing to do

steve-warren:
@HaoK, Would it make more sense to use the already implemented Rfc6238AuthenticationService.GenerateRandomKey() to generate the key? I noticed it is dead code and no longer referenced.

To me it would make sense to have key generation tied to the implementation of the algorithm as the length of the key depends on the length of the HMAC output (in this case, HMACSHA1, 160 bits).

PR #1719, opened 2018-03-28
[Fixes #1690] AddDefaultUI() throws when using IdentityUser<T>

### LINKED ISSUE #1690: AddDefaultUI() throws when using IdentityUser<T>
If your user and role inherit from IdentityUser\<T\> / IdentityRole\<T\> instead of IdentityUser / IdentityRole, then AddDefaultUI() throws.

Project to reproduce here: https://github.com/jeffhube/aspnet-identity-test

```csharp
public class ApplicationDbContext : IdentityDbContext<ApplicationUser, ApplicationRole, Guid>
{
}

public class ApplicationUser : IdentityUser<Guid>
{
}

public class ApplicationRole : IdentityRole<Guid>
{
}

services.AddIdentity<ApplicationUser, ApplicationRole>()
    .AddDefaultTokenProviders()
    .AddDefaultUI();
```

```
Unhandled Exception: System.ArgumentException: GenericArguments[0], 'IdentityTest.ApplicationUser', on 'Microsoft.AspNetCore.Identity.UI.IdentityDefaultUIConfigureOptions`1[TUser]' violates the constraint of type 'TUser'. ---> System.TypeLoadException: GenericArguments[0], 'IdentityTest.ApplicationUser', on 'Microsoft.AspNetCore.Identity.UI.IdentityDefaultUIConfigureOptions`1[TUser]' violates the constraint of type parameter 'TUser'.
   at System.RuntimeTypeHandle.Instantiate(RuntimeTypeHandle handle, IntPtr* pInst, Int32 numGenericArgs, ObjectHandleOnStack type)
   at System.RuntimeTypeHandle.Instantiate(Type[] inst)
   at System.RuntimeType.MakeGenericType(Type[] instantiation)
   --- End of inner exception stack trace ---
   at System.RuntimeType.ValidateGenericArguments(MemberInfo definition, RuntimeType[] genericArguments, Exception e)
   at System.RuntimeType.MakeGenericType(Type[] instantiation)
   at Microsoft.AspNetCore.Identity.IdentityBuilderUIExtensions.AddDefaultUI(IdentityBuilder builder)
   at IdentityTest.Startup.ConfigureServices(IServiceCollection services) in C:\Users\Jeff\Desktop\IdentityTest\Startup.cs:line 32
--- End of stack trace from previous location where exception was thrown ---
   at Microsoft.AspNetCore.Hosting.ConventionBasedStartup.ConfigureServices(IServiceCollection services)
   at Microsoft.AspNetCore.Hosting.Internal.WebHost.EnsureApplicationServices()
   at Microsoft.AspNetCore.Hosting.Internal.WebHost.Initialize()
   at Microsoft.AspNetCore.Hosting.WebHostBuilder.Build()
   at IdentityTest.Program.BuildWebHost(String[] args) in C:\Users\Jeff\Desktop\IdentityTest\Program.cs:line 21
   at IdentityTest.Program.Main(String[] args) in C:\Users\Jeff\Desktop\IdentityTest\Program.cs:line 17
```

kichalla:


test/UI.Test/IdentityPageModelConventionTest.cs
```diff
@@ -0,0 +1,120 @@
+﻿using Microsoft.AspNetCore.Mvc.ApplicationModels;
+using Microsoft.AspNetCore.Mvc.RazorPages;
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using Xunit;
+
+namespace Microsoft.AspNetCore.Identity.UI.Test
+{
+    public class IdentityPageModelConventionTest
```
kichalla: I would like to also add some functional tests with the custom identity user model but the currently test website that we have has a razor page which seems is for testing the overriding scenario (?) and uses the model as `IdentityUser` which doesn't work for my scenario (even if i have a difference Startup file having a custom identityuse). Any suggestions? 
Is it fine to create a separate website for this purpose? It feels like a overkill but we ideally want an end-to-end test.

javiercn: There is a separate website in Testing/WebSites

src/UI/IdentityPageModelConvention.cs
```diff
@@ -7,7 +7,8 @@
 
 namespace Microsoft.AspNetCore.Identity.UI
 {
-    internal class IdentityPageModelConvention<TUser> : IPageApplicationModelConvention where TUser : IdentityUser
+    internal class IdentityPageModelConvention<TUser> : IPageApplicationModelConvention
```
javiercn: Make this TUser,TKey and have the layer above pass it to you

src/UI/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
```diff
@@ -40,7 +40,9 @@ public class InputModel
         public virtual Task<IActionResult> OnPostConfirmationAsync(string returnUrl = null) => throw new NotImplementedException();
     }
 
-    internal class ExternalLoginModel<TUser> : ExternalLoginModel where TUser : IdentityUser, new()
+    internal class ExternalLoginModel<TUser, TKey> : ExternalLoginModel
```
HaoK: So we actually need to get rid of the IdentityUser constraint entirely...and we certainly don't want to bring in the TKey either...

I believe the new TUser calls are what's tying you IdentityUser since its using the UserName/Email properties to initialize.

I think if you introduce a UI service like `TUser IUIUserInstanceFactory<TUser>.Create(string userName, string email`, and we can add some logic to AddDefaultUI to automatically register a concrete `UserInstanceFactory<TUser>` that does what we do today whenever TUser is an IdentityUser, otherwise we don't, and the app will need to register the factory

src/UI/IdentityPageModelConvention.cs
HaoK: The good news is this will be a lot simpler since it will probably only require that TUser is a class

src/UI/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
javiercn: I would not introduce a new abstraction. Simply force TUser to be a class and have a parameterless constructor

javiercn: I would suggest to start by replacing any access to user data with method calls from user manager and then apply the minimum required set of constraints

HaoK: Yeah unfortunately that won't work, CreateAsync validates the user which requires a user name to be set at a minimum, and potentially a non null email if that was required as well, otherwise all usermanager operations will be rejected.  The idea is that it was up to the callee to create TUser however they wish and then call CreateAsync when ready...  Its an abstraction that isn't needed for IdentityCore, but its now a new abstraction that is needed for the UI :(

src/UI/Areas/Identity/Pages/Account/Manage/ExternalLogins.cshtml.cs
```diff
@@ -57,7 +58,9 @@ internal class ExternalLoginsModel<TUser> : ExternalLoginsModel where TUser : Id
             OtherLogins = (await _signInManager.GetExternalAuthenticationSchemesAsync())
                 .Where(auth => CurrentLogins.All(ul => auth.Name != ul.LoginProvider))
                 .ToList();
-            ShowRemoveButton = user.PasswordHash != null || CurrentLogins.Count > 1;
+
+            string passwordHash = null;
```
kichalla: I couldn't find a way to get this value. I tried IUserPasswordStore but that service seems to be not necessarily registered. @HaoK any ideas?

src/UI/Areas/Identity/Pages/Account/Manage/EnableAuthenticator.cshtml.cs
```diff
@@ -102,7 +102,8 @@ internal class EnableAuthenticatorModel<TUser> : EnableAuthenticatorModel where
             }
 
             await _userManager.SetTwoFactorEnabledAsync(user, true);
-            _logger.LogInformation("User with ID '{UserId}' has enabled 2FA with an authenticator app.", user.Id);
+            var userId = await _userManager.GetUserIdAsync(user);
```
kichalla: Doing these extra calls for logging seems overkill. We can do the `LoggerMessage.Define` pattern for all the logging here.

src/UI/Areas/Identity/Pages/Account/Manage/ExternalLogins.cshtml.cs
HaoK: You can use the IUserStore service and cast it to IUserPasswordStore

src/UI/IUIUserInstanceFactory.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace Microsoft.AspNetCore.Identity.UI
+{
+    public interface IUIUserInstanceFactory<TUser> where TUser : class
```
HaoK: This name is pretty meh, maybe we should just call it IUserFactory with a Create method

src/UI/IdentityBuilderUIExtensions.cs
```diff
@@ -36,6 +36,11 @@ public static IdentityBuilder AddDefaultUI(this IdentityBuilder builder)
                     .MakeGenericType(builder.UserType));
             builder.Services.TryAddTransient<IEmailSender, EmailSender>();
 
+            if (builder.UserType == typeof(IdentityUser))
+            {
+                builder.Services.AddSingleton<IUIUserInstanceFactory<IdentityUser>, DefaultUIUserInstanceFactory>();
```
HaoK: We can be a bit smarter than this too I think, with `IdentityUserFactory<TUser, TKey> : IUserFactory<TUser> where TUser : IdentityUser<TKey>` and use reflection on builder.UserType to see if its a derived IdentityUser and then register the appropriate closed IdentityUserFactory.

src/UI/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
```diff
@@ -40,19 +41,22 @@ public class InputModel
         public virtual Task<IActionResult> OnPostConfirmationAsync(string returnUrl = null) => throw new NotImplementedException();
     }
 
-    internal class ExternalLoginModel<TUser> : ExternalLoginModel where TUser : IdentityUser, new()
+    internal class ExternalLoginModel<TUser> : ExternalLoginModel where TUser : class
     {
         private readonly SignInManager<TUser> _signInManager;
         private readonly UserManager<TUser> _userManager;
+        private readonly IUIUserInstanceFactory<TUser> _userInstanceFactory;
```
javiercn: Please, do not introduce a new abstraction. Keep the new() constraint around instead

HaoK: How would a new constraint help when to be able to save a user both a UserName / and or Email must be set as well.

javiercn: Use https://github.com/aspnet/Identity/blob/85f8a49aef68bf9763cd9854ce1dd4a26a7c5d3c/src/Stores/UserStoreBase.cs#L152 instead

HaoK: Nope: https://github.com/aspnet/Identity/blob/85f8a49aef68bf9763cd9854ce1dd4a26a7c5d3c/src/Stores/UserStoreBase.cs#L35

Those are tied to identity user as well.

src/UI/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
```diff
@@ -124,7 +128,7 @@ public override IActionResult OnPost(string provider, string returnUrl = null)
 
             if (ModelState.IsValid)
             {
-                var user = new TUser { UserName = Input.Email, Email = Input.Email };
+                var user = _userInstanceFactory.CreateUser(email: Input.Email, userName: Input.Email);
```
javiercn:
```csharp
var user = new TUser();
await _userStore.SetEmailAsync(input.Email);
await _userStore.SetUserNameAsync(input.Email);
_userManager.CreateAsync(user);
```

src/UI/Areas/Identity/Pages/Account/Manage/EnableAuthenticator.cshtml.cs
javiercn: Your call. But file a bug and do it on a separate PR.

src/UI/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
HaoK: That will work, with the tradeoff of requiring overriding multiple views required whenever they want to add required data, but if that's already required I guess that's fine.

src/UI/IUIUserInstanceFactory.cs
javiercn: As mentioned above. Let's not introduce an additional abstraction for this and just rely on the parameter-less constructor constraint.

src/UI/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
HaoK: This is sorta breaking the abstractions though, the whole point is you aren't supposed to use the stores directly, only the manager...

src/UI/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
javiercn: https://github.com/aspnet/Identity/blob/85f8a49aef68bf9763cd9854ce1dd4a26a7c5d3c/src/Core/IUserStore.cs#L39

src/UI/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
HaoK: The password hash null check is unavoidable since the only workaround would be calling AddPassword with a null password and checking if that fails, which is super gross.

src/UI/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
javiercn: I meant to use them through the interface, just pointed to the implementation so that Kiran could see what they were doing

src/UI/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
HaoK: You'd have to do a cast/check to make sure the user store even supports Email as well going this route...

HaoK: I think in the end, the core abstraction that the UI needs to introduce around custom users, is around the registration model and creating that initial instance.

Basically what's needed is a way to plug in a user data schema (property name + data type), and generate some UI around that data by default, and then create a TUser instance with that data, and then pass that into Identity Core.

We can file an issue to track this separately, I guess for preview 2 we can just hack something in that works for now and revisit this in rc if needed

javiercn: The user manager doesn't have the functionality. If you are OK adding it, we can just add it there. This can take multiple forms:
```csharp
_userManager.Create(userName, email)
```
```
_userManager.SetUserNameAsync(user, userName, updateUser:false)
_userManager.SetEmailAsync(user, email, updateUser:false)
```

src/UI/IdentityBuilderUIExtensions.cs
```diff
@@ -36,6 +36,11 @@ public static IdentityBuilder AddDefaultUI(this IdentityBuilder builder)
                     .MakeGenericType(builder.UserType));
             builder.Services.TryAddTransient<IEmailSender, EmailSender>();
 
+            if (builder.UserType == typeof(IdentityUser))
+            {
+                builder.Services.TryAddSingleton<IUIUserInstanceFactory<IdentityUser>, DefaultUIUserInstanceFactory>();
```
javiercn: This will not work, you need to register it as TUser, if you get a custom user this will not be registered and we don't want people to have to register a custom factory for creating users. This will not even work in the default templates as they use ApplicationUser

HaoK: Yeah I commented earlier, maybe on an older version, but this should be smarter and user reflection and set things up properly for any derived `IdentityUser<TKey>` derived `TUser` type

src/UI/UserFactory.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+
+namespace Microsoft.AspNetCore.Identity.UI
+{
+    internal class UserFactory<TUser, TKey> : IUserFactory<TUser>
```
HaoK: Nit: what do you think about making this public instead of internal?  Not a big deal either way but I could see this being somewhat useful potentially.

kichalla: Yeah, should be ok, but will leave it as it is for now as we are going to revisit this api for rc1

src/UI/Areas/Identity/Pages/Account/Manage/ExternalLogins.cshtml.cs
```diff
@@ -57,7 +61,14 @@ internal class ExternalLoginsModel<TUser> : ExternalLoginsModel where TUser : Id
...
+            if (_userStore is IUserPasswordStore<TUser> userPasswordStore)
+            {
+                passwordHash = await userPasswordStore.GetPasswordHashAsync(user, CancellationToken.None);
```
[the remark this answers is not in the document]
kichalla: Ah, `RequestAborted` was the one I was looking for but didn't realize about the name. Thanks, will update.

src/UI/Areas/Identity/Pages/Account/Manage/EnableAuthenticator.cshtml.cs
kichalla: I will file a separate issue for this.

PR #1734, opened 2018-04-03
React to future Razor Sdk package breaking changes

pranavkm:


pranavkm:
@Eilon we have a change in the Razor Sdk that would cause Identity UI and AADIntegration UI to break when consumed. Preempting the issue by referencing the package.

Issue #1722, opened 2018-03-29
Simplify using a custom TUser with DefaultUI


HaoK:
This came up as a part of investigating the fix to https://github.com/aspnet/Identity/issues/1679

Today to add additional basic data to a custom user requires:
- overriding the Register/Manage pages, see https://github.com/aspnet/Identity/tree/dev/samples/IdentitySample.DefaultUI/Areas/Identity/Pages/Account, to hand craft the InputModel and update the actions.

```
   public class MyUser : IdentityUser {
       public string Name { get; set; }
       public int Age { get; set; }
   }
```

I think we can simplify this to only needing to define only the additional properties + the corresponding input model if we add a basic mechanism for generating the appropriate inputs and introduce a new abstraction in the UI package around updating/creating the actual TUser:

```C#
   public class MyUser : IdentityUser {
       public string Name { get; set; }
       public int Age { get; set; }
   }

   public class MyUserInputModel : RegisterInputModel, ManageInputModel {
            [Required]
            [DataType(DataType.Text)]
            [Display(Name = "Full name")]
            public string Name { get; set; }

            [Required]
            [Range(0,199, ErrorMessage = "Age must be between 0 and 199 years")]
            [Display(Name = "Age")]
            public int Age { get; set; }
   }

   public interface IUserFactory<TUser> {
       TUser Create(IUserData data); // Register calls this to create a new user
       void Update(TUser user, IUserData data); // Manage uses this to update the user
   }

   // We would provide a default implementation of this that uses reflection to work against properties, only users with truly custom POCOs would need to implement this.

    // We can add a way to generate the Register/Manage.cshtml inputs from the input models, like IInputGenerator.Generate(InputModel) to generate the form groups and give the ability to customize:
            <div class="form-group">
                <label asp-for="Input.Name"></label>
                <input asp-for="Input.Name" class="form-control" />
                <span asp-validation-for="Input.Name" class="text-danger"></span>
            </div>
            <div class="form-group">
                <label asp-for="Input.Age"></label>
                <input asp-for="Input.Age" class="form-control" />
                <span asp-validation-for="Input.Age" class="text-danger"></span>
            </div>
```

HaoK:
Thoughts @DamianEdwards @danroth27 @javiercn @ajcvickers @blowdart @kichalla  @mkArtakMSFT ?

mkArtakMSFT:
I've assigned this to rc1 milestone independent on the actual design decision here. We just know that there is some work to be done here.

[...]

PR #1743, opened 2018-04-09
Switch to abstract user factory

HaoK:
Part of https://github.com/aspnet/Identity/issues/1722


javiercn:
@HaoK Can we press pause on this? I don't want to rush https://github.com/aspnet/Identity/issues/1722 in 2.1 as the scope is much bigger and we need time to properly bake it in,

The requirement for 2.1 is that you can use a custom TUser with the default Identity UI. I agreed to get this in for preview2, but I wanted us to revisit it on RC and discuss alternatives.

Can we setup a meeting to discuss this?

HaoK:
Sure we can discuss, but going with an abstract base class should mean that we can fix #1722 however we want later since it won't be a breaking change to add new methods to the base class

HaoK:
I don't think we should try and actually do the full on thing in 2.1 at all

javiercn:
Close? @HaoK

Issue #1736, opened 2018-04-05
/Identity/Account/ResetPassword page is not reachable when auth required globally


dougbu:
Looks like at least one page was missed when fixing #1617 
## Repro steps
1. `dotnet new mvc --name netcoreIndividualAuth --auth Individual --use-local-db` (doubt `--use-local-db` is mandatory)
2. `cd netcoreIndividualAuth`
3. Open project in Visual Studio
4. Edit `Startup.ConfigureServices(...)` to require an authenticated user globally i.e. change `services.AddMvc().SetCompatibilityVersion(CompatibilityVersion.Version_2_1);` to
  ``` c#
  services
      .AddMvc(configuration =>
      {
          var policy = new AuthorizationPolicyBuilder()
              .RequireAuthenticatedUser()
              .Build();
          configuration.Filters.Add(new AuthorizeFilter(policy));
      })
      .SetCompatibilityVersion(CompatibilityVersion.Version_2_1);
  ```

5. Start app in the debugger
6. Ensure debug option "Enable Just My Code" is disabled and "Microsoft Symbol Servers" are enabled under "Symbols"
7. Set C# function breakpoint at `Microsoft.AspNetCore.Identity.UI.Services.EmailSender.SendEmailAsync` 
8. In browser, register a new user
9. In debugger, copy the URL in the `htmlMessage` parameter
10. In browser, head to the copied URL but change `&amp;code` to `&code`
11. In browser, log out, click on "Forgot your password", then enter the new user's email and click Submit
12. In debugger, copy the URL in the `htmlMessage` parameter
13. In browser, head to the copied URL

### Expected
- Land on /Identity/Account/ForgotPasswordConfirmation page after step 11
- Land on /Identity/Account/ResetPassword page after step 13
### Actual
Redirected to the login page again in both cases i.e. both page models lack `[AllowAnonymous]` though the lack in /Identity/Account/ResetPassword is more troublesome.

----
This is a small part of #1617 which does not work end-to-end. Found while verifying (aspnet/Release#263) with the 2.1.300-preview2-008523 SDK and Microsoft.AspNetCore.Identity.UI 2.1.0-preview2-30552.

Eilon:
So... is the suggestion to just add AllowAnonymous to the two pages that should allow anonymous?

dougbu:
Yes, exactly. Sorry for leaving that out.

dougbu:
[...]

PR #1744, opened 2018-04-10
Add AllowAnonymous for resetting password navigation flow

jbagga:
Addresses #1736 and aspnet/templating#448

### DIFF
--- a/src/UI/Areas/Identity/Pages/Account/ResetPassword.cshtml.cs
+++ b/src/UI/Areas/Identity/Pages/Account/ResetPassword.cshtml.cs
@@ -4,11 +4,13 @@
 using System;
 using System.ComponentModel.DataAnnotations;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.RazorPages;
 
 namespace Microsoft.AspNetCore.Identity.UI.Pages.Account.Internal
 {
+    [AllowAnonymous]
     [IdentityDefaultUI(typeof(ResetPasswordModel<>))]
     public abstract class ResetPasswordModel : PageModel
     {
--- a/src/UI/Areas/Identity/Pages/Account/ResetPasswordConfirmation.cshtml.cs
+++ b/src/UI/Areas/Identity/Pages/Account/ResetPasswordConfirmation.cshtml.cs
@@ -1,10 +1,12 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc.RazorPages;
 
 namespace Microsoft.AspNetCore.Identity.UI.Pages.Account.Internal
 {
+    [AllowAnonymous]
     public class ResetPasswordConfirmationModel : PageModel
     {
         public void OnGet()
--- a/src/UI/Microsoft.AspNetCore.Identity.UI.csproj
+++ b/src/UI/Microsoft.AspNetCore.Identity.UI.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk.Razor">
+﻿<Project Sdk="Microsoft.NET.Sdk.Razor">
   
   <PropertyGroup>
     <Description>ASP.NET Core Identity UI is the default Razor Pages built-in UI for the ASP.NET Core Identity framework.</Description>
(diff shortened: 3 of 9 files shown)


src/UI/Areas/Identity/Pages/Account/ResetPassword.cshtml.cs
```diff
@@ -4,11 +4,13 @@
 using System;
 using System.ComponentModel.DataAnnotations;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.RazorPages;
 
 namespace Microsoft.AspNetCore.Identity.UI.Pages.Account.Internal
 {
+    [AllowAnonymous]
```
kichalla: This seems incorrect as only a logged in user should be able to reset the password

src/UI/Areas/Identity/Pages/Account/ResetPasswordConfirmation.cshtml.cs
```diff
@@ -1,10 +1,12 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Mvc.RazorPages;
 
 namespace Microsoft.AspNetCore.Identity.UI.Pages.Account.Internal
 {
+    [AllowAnonymous]
```
kichalla: Users typically reach this after they login | reset password | reset password confirmation...so i think this should not be allowanonymous

src/UI/Areas/Identity/Pages/Account/ResetPassword.cshtml.cs
kichalla: This should probably explicitly have Authorize attribute.

kichalla: We probably need a consistent experience of using these attributes. We currently do like the following to authorize some areas. @javiercn ?
https://github.com/aspnet/Identity/blob/release/2.1/src/UI/IdentityDefaultUIConfigureOptions.cs#L37-L38

src/UI/Areas/Identity/Pages/Account/ResetPasswordConfirmation.cshtml.cs
jbagga: This is when the user has forgotten the password and clicks the `ForgotPassword` link. The flow you're talking about is through Manage action `ChangePassword`. In the `ForgotPassword` case the user is not logged in and gets an email to reset the password

1. User has to access `ResetPassword` to set a new password (from the link in the email) and is not logged in automatically (has to sign in again)
2. Being redirected to `ResetPasswordConfirmation` after the previous step makes it clear that a new password was set and the user may log in. 
3. If `ResetPasswordConfirmation` is not allowed to be accessed anonymously, the user will be redirected to the Login page without any confirmation.

kichalla: As yes, you are right. I confused myself with Chnage Password flow. Looks good to me then!

src/UI/Microsoft.AspNetCore.Identity.UI.csproj
```diff
@@ -16,7 +16,6 @@
     <EmbeddedResource Remove="wwwroot/**/LICENSE*" />
     <None Remove="wwwroot/**/.bower.json" />
     <None Remove="wwwroot/**/LICENSE*" />
-    <None Include="THIRD-PARTY-NOTICES" Pack="true" />
```
jbagga: @javiercn This was added to include the THIRD-PARTY-NOTICES to the Identity UI package as part of https://github.com/aspnet/Identity/commit/4105ee67e907301539e0e2392e7ae50321b93fb5
But as a result it appears in all templates. 

Do we just want the notices in the GitHub repo or also as a part of the package? 
cc @Eilon @kichalla

Eilon: The notices file needs to be within the NUPKGs. But not in the generated project template or otherwise visible when this NUPKG is installed in a project.

jbagga: I have made a change to reflect that. It is in the nupkg but does not appear in the generated project

jbagga:
@javiercn Please let me know when your infrastructure changes PR is merged (tag me or link it here) so I can go ahead with changing/merging this!

jbagga:
Merging now

PR #1745, opened 2018-04-10
Include social logins and authenticator key as part of personal download

HaoK:
We now display the linked login providers and the authenticator key if it exists.

Apps will be responsible for displaying any additional custom tokens they might store (like access/refresh tokens) since we don't have any way of enumerating these today.

@blowdart @ajcvickers


HaoK:
There will be another PR that adds encryption/protection to these data as well

HaoK:
@ajcvickers updated the PR adding [ProtectedPersonalData] to the token values as well as hooking up converters for anything on TUserToken as well

HaoK:
I left in a test in the key rotation to verify that logins work before and after ( even tho I backed out the changes that protect the external login provider...  this test will be needed in 2.2 anyways)

Issue #1725, opened 2018-03-30
Add functional tests for different user types in DefaultUI


kichalla:
Some user types to think of:
- `class IdentityUser`
- `class TestUser : IdentityUser`
- `class TestUser : IdentityUser<Guid>`
- `class TestUser`

javiercn:
Moving this to RC1 as I've already done it as part of making the tests more generally applicable

Issue #1706, opened 2018-03-23
Unable to disable the defaultUI in asp.net core 2.1


baseless:
Hi!

I am uncertain if im missing soemthing but removing the default UI should be as simple as removing `AddDefaultUI()` but it odes not seem to be enough. 
If i remove this line from the default template and  try to access /Identity/Account/Register i recieve "InvalidOperationException: A suitable constructor for type 'Microsoft.AspNetCore.Identity.UI.Pages.Account.Internal.RegisterModel' could not be located" when i should logically recieve a 404.

Anyone happen to know if im missing ssomething here or if this is an active issue?

Eilon:
@javiercn can you help answer this?

Eilon:
@baseless if you can show the stack trace of the exception that would also be helpful. Thanks!

javiercn:
And the startup code you are using

baseless:
@Eilon the only steps i did was:

1. Create new 2.1 Web application (installed vs2017 preview 2 update earlier) with individual user accounts auth.
2. Remove the 'AddDefaultUI()' and deleted the 'Areas' folder.
3. Run project and visit /Identity/Account/Register in browser.

The raw exception details are:

```
System.InvalidOperationException: A suitable constructor for type 'Microsoft.AspNetCore.Identity.UI.Pages.Account.Internal.RegisterModel' could not be located. Ensure the type is concrete and services are registered for all parameters of a public constructor.
   at Microsoft.Extensions.DependencyInjection.ActivatorUtilities.FindApplicableConstructor(Type instanceType, Type[] argumentTypes, ConstructorInfo& matchingConstructor, Nullable`1[]& parameterMap)
   at Microsoft.Extensions.DependencyInjection.ActivatorUtilities.CreateFactory(Type instanceType, Type[] argumentTypes)
   at Microsoft.AspNetCore.Mvc.RazorPages.Infrastructure.DefaultPageModelActivatorProvider.CreateActivator(CompiledPageActionDescriptor actionDescriptor)
   at Microsoft.AspNetCore.Mvc.RazorPages.Infrastructure.DefaultPageModelFactoryProvider.CreateModelFactory(CompiledPageActionDescriptor descriptor)
   at Microsoft.AspNetCore.Mvc.RazorPages.Internal.PageActionInvokerProvider.CreateCacheEntry(ActionInvokerProviderContext context, FilterItem[] cachedFilters)
[...]

PR #1749, opened 2018-04-11
[Fixes #1742, #1722, #1706, #1725] Fixing multiple issues

### LINKED ISSUE #1742: Logout must redirect instead of just rendering the page so that the identity in the request gets properly updated
Logout.cshtml and Logout.cshtml.cs are used before _and after_ the user is logged out. If user posts to the Logout page without a `returnUrl` value, the page displays a message: `You have successfully logged out of the application.` Unfortunately, the page is unreachable after logging out (i.e. requires authentication) due to the following line in the generated `IdentityDefaultUIConfigureOptions.PostConfigure(...)`: https://github.com/aspnet/Identity/blob/394e4746b89485e6f6d0736b363d947361e7a378/src/UI/IdentityDefaultUIConfigureOptions.cs#L38

Because this setting is done in an `IPostConfigureOptions<CookieAuthenticationOptions>` implementation, it is extra hard for users to override.

See also aspnet/Scaffolding#734

javiercn:
* Add a test to validate that the Identity endpoints are not available when the default UI is not plugged in.
* Remove IUserFactory<TUser> and UserFactory<TUser>
* Update the testing infrastructure to newer bits from MVC.
* Parameterize the testing infrastructure to support testing different user types.
* Make EmailSender internal.


samples/IdentitySample.DefaultUI/IdentitySample.DefaultUI.csproj
```diff
@@ -32,4 +32,13 @@
...
 
+  <ItemGroup>
+    <Content Update="Pages\About.cshtml">
+      <Pack>$(IncludeRazorContentInPack)</Pack>
+    </Content>
+    <Content Update="Pages\_ViewStart.cshtml">
+      <Pack>$(IncludeRazorContentInPack)</Pack>
+    </Content>
+  </ItemGroup>
+
```
javiercn: Revert

samples/IdentitySample.DefaultUI/Pages/About.cshtml
```diff
@@ -0,0 +1,3 @@
+@page
```
javiercn: Revert

samples/IdentitySample.DefaultUI/Views/Shared/_Layout.cshtml
```diff
@@ -22,6 +22,7 @@
                 <div class="navbar-collapse collapse">
                     <ul class="nav navbar-nav">
                         <li><a asp-controller="Home" asp-action="Index">Home</a></li>
+                        <li><a asp-page="/About">About</a></li>
```
javiercn: Revert

test/WebSites/Identity.DefaultUI.WebSite/PocoUserStartup.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        public override void ConfigureServices(IServiceCollection services)
+        {
+            services.Configure<CookiePolicyOptions>(options =>
+            {
+                // This lambda determines whether user consent for non-essential cookies is needed for a given request.
+                options.CheckConsentNeeded = context => true;
+            });
+
+            services.AddDefaultIdentity<TestUser>()
```
HaoK: Isn't this supposed to be using PocoUser?

javiercn: TestUser = PocoUser, I saw there was already a store implementation for it on shared and reused it.

javiercn: I can rename the tests to TestUser but I think its better to convey all over the place that TestUser is a poco

HaoK: Just rename TestUser => PocoUser then

src/UI/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
```diff
@@ -44,18 +45,20 @@ internal class ExternalLoginModel<TUser> : ExternalLoginModel where TUser : clas
...
-        private readonly IUserFactory<TUser> _userFactory;
+        private readonly IUserStore<TUser> _userStore;
+        private readonly IUserEmailStore<TUser> _emailStore;
         private readonly ILogger<ExternalLoginModel> _logger;
 
         public ExternalLoginModel(
             SignInManager<TUser> signInManager,
             UserManager<TUser> userManager,
-            IUserFactory<TUser> userFactory,
+            IUserStore<TUser> userStore,
```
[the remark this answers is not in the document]
HaoK: code smell doesn't go away just because its internal :)  

I thought we agreed that there was no supported scenario where they would be using the built in pages with a custom user, that the expectation is that they MUST override these with a custom user.  If that's the case, can't we just constrain these pages to be IdentityUser?

test/WebSites/Identity.DefaultUI.WebSite/PocoUserStartup.cs
javiercn: Sounds good

src/UI/Areas/Identity/Pages/Account/ExternalLogin.cshtml.cs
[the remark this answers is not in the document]
HaoK: So I think we can just check in the ctor and throw if TUser isn't derived from `IdentityUser<TKey>` and you can say they need to override the model in this case.

Then we can just set the properties via reflection

src/UI/Areas/Identity/Pages/Account/Manage/Index.cshtml.cs
```diff
@@ -104,6 +104,15 @@ internal class IndexModel<TUser> : IndexModel where TUser : class
                     var userId = await _userManager.GetUserIdAsync(user);
                     throw new InvalidOperationException($"Unexpected error occurred setting email for user with ID '{userId}'.");
                 }
+
+                // In our UI email and user name are one and the same, so when we update the email
+                // we need to update the user name.
+                var setUserNameResult = await _userManager.SetUserNameAsync(user, Input.Email);
+                if (!setUserNameResult.Succeeded)
+                {
+                    var userId = _userManager.GetUserIdAsync(user);
```
kichalla: await

Issue #1669, opened 2018-03-08
Misleading exception message in Identity 2.0.1


marcwittke:
```
System.NotSupportedException : No IUserTokenProvider named 'Default' is registered.
   at Microsoft.AspNetCore.Identity.UserManager`1.GenerateUserTokenAsync(TUser user, String tokenProvider, String purpose)
   at Microsoft.AspNetCore.Identity.UserManager`1.GeneratePasswordResetTokenAsync(TUser user)
```

The framework in reality requires an implementation of `IUserTwoFactorTokenProvider<TUser>`

PR #1752, opened 2018-04-12
Fix error message

HaoK:
Fixes https://github.com/aspnet/Identity/issues/1669

### DIFF
--- a/src/Core/Resources.resx
+++ b/src/Core/Resources.resx
@@ -174,8 +174,8 @@
     <comment>Error when AddIdentity is not called</comment>
   </data>
   <data name="NoTokenProvider" xml:space="preserve">
-    <value>No IUserTokenProvider named '{0}' is registered.</value>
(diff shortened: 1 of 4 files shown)


src/Core/Resources.resx
```diff
@@ -174,8 +174,8 @@
     <comment>Error when AddIdentity is not called</comment>
   </data>
   <data name="NoTokenProvider" xml:space="preserve">
-    <value>No IUserTokenProvider named '{0}' is registered.</value>
-    <comment>Error when there is no IUserTokenProvider</comment>
+    <value>No IUserTwoFactorTokenProvider named '{0}' is registered.</value>
```
Eilon: Consider using `nameof(T)` by passing in a `{1}` into here.

HaoK: Sure I'll update this instance since I'm editing this, but I don't want to update all the resources that need this change in this PR

Eilon: Works for me.

PR #1761, opened 2018-04-17
Fix build break

pranavkm:
* Use Razor.Sdk package to not perform runtime compilation
* Use Mvc with PageResult fix (https://github.com/aspnet/Mvc/issues/7665)


pranavkm:
cc @mikeharder \ @mkArtakMSFT - I have a fix for one of the issues causing Identity tests to fail. There's a second issue that's currently pending ship room approval. https://github.com/aspnet/Mvc/pull/7672. I used a feature branch package to verify the issue's resolved when the package is consumed. IMO, we should treat this as a build break and check in this + the Mvc PR.

PR #1777, opened 2018-04-30
Added SignInManager interface

zahasoft:
It is terrible to mock SignInManager without an interface.

### DIFF
--- a/src/Identity/ISignInManager.cs
+++ b/src/Identity/ISignInManager.cs
@@ -1,0 +1,278 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Collections.Generic;
+using System.Security.Claims;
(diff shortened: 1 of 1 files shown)


HaoK:
This shouldn't really affect mocking, you can already Mock `SignInManager<TUser>` already, existing code is already asking for SignInManager from DI, irrespective of this being a breaking change, adding an interface isn't going to change anything unless all the consuming code was updated to depend on this new interface.

See https://github.com/aspnet/Identity/blob/f51af820a5ccf2268b785c62140b29bbb5a24e83/test/Identity.Test/SecurityStampValidatorTest.cs#L105 for an example of how we mock sign in manager today

zahasoft:
I know that. It was just the act of my annoyance. Actually I think that it is better to separate this huge classes to smaller with less injected dependencies or to remove null checks for some dependencies in their constructors so to eliminate the need to be mocked 5-6 classes just to simulate return value for one method only.

PR #1794, opened 2018-05-16
Fixed sha-integrity content of script tags in IdentityUI's cshtml files

kichalla:
Related to https://github.com/aspnet/templating/issues/520


test/Identity.Test/CdnScriptTaghelperTests.cs
```diff
@@ -0,0 +1,111 @@
...
+        public CdnScriptTagTests(ITestOutputHelper output)
+        {
+            _output = output;
+        }
+
+        [Fact]
+        public async Task IdentityUI_ScriptTags_SubresourceIntegrityCheck()
+        {
+            var slnDir = GetSolutionDir();
+            var sourceDir = Path.Combine(slnDir, "src", "UI", "Areas", "Identity", "Pages");
```
kichalla: @natemcmaster I am looking into 'src' here because IdentityUI's .cshtml files are embedded in its assembly and its a bit more involved in getting their content (even though that is the ideal thing to do in tests). We can file a follow-up issue to fix this, if needed.

natemcmaster: I think checking the original source is good enough for now. Would it be better to have this test glob over every 
.cshtml file under src/UI/ ?

PR #1805, opened 2018-05-27
Allow to access internal methods of SignInManager from a derived class

hikalkan:
For a multi-tenant application, I need to override these methods.


HaoK:
Please file an issue if this is still something important to you

HaoK:
The code has moved to https://github.com/aspnet/AspNetCore/

PR #1827, opened 2018-06-11
Fix /// comments

tstivers1990:
The returns tag incorrectly states that AddDefaultIdentity<T>
returns IServiceCollection when it actually returns IdentityBuilder.


HaoK:
Thanks!

PR #1830, opened 2018-06-13
Set 2.1 baselines

ryanbrandenburg:


HaoK:
The service stuff hasn't shipped yet, not sure if that's something we should baseline... @javiercn ?

ryanbrandenburg:
If something isn't shipping yet we should set `<EnableApiCheck>false</EnableApiCheck>` in its csproj to prevent baselines from being generated. By "service stuff" I assume you mean any project which starts with `Service.`? I can remove them from this PR.

ryanbrandenburg:
Talked to @javiercn offline, he doesn't mind that it's included here.

PR #1838, opened 2018-06-18
Fix ActivePage declarations

tstivers1990:
* Some of the `ViewData["ActivePage"]` declarations were missing.
* `DeletePersonalData` and `DownloadPersonalData` should set `ActivePage` to `PersonalData` as there is no navigation element for `DeletePersonalData` or `DownloadPersonalData`.
* Remove `DeletePersonalData` and `DownloadPersonalData` variables and methods as they're no longer used by anthing.
* Make `PageNavClass` private as it is never used outside of its class.

### DIFF
--- a/src/UI/Areas/Identity/Pages/Account/Manage/ChangePassword.cshtml
+++ b/src/UI/Areas/Identity/Pages/Account/Manage/ChangePassword.cshtml
@@ -2,6 +2,7 @@
 @model ChangePasswordModel
 @{
     ViewData["Title"] = "Change password";
+    ViewData["ActivePage"] = ManageNavPages.ChangePassword;
 }
 
 <h4>@ViewData["Title"]</h4>
--- a/src/UI/Areas/Identity/Pages/Account/Manage/DeletePersonalData.cshtml
+++ b/src/UI/Areas/Identity/Pages/Account/Manage/DeletePersonalData.cshtml
@@ -2,7 +2,7 @@
 @model DeletePersonalDataModel
 @{
     ViewData["Title"] = "Delete Personal Data";
-    ViewData["ActivePage"] = ManageNavPages.DeletePersonalData;
+    ViewData["ActivePage"] = ManageNavPages.PersonalData;
 }
 
 <h4>@ViewData["Title"]</h4>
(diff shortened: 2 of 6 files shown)


tstivers1990:
The `TwoFactorAuthentication` page was also missing the `ActivePage` assignment. I've added that in the latest commit.

HaoK:
@javiercn do these changes look reasonable to you?

javiercn:
Thanks for contacting us @tstivers1990 
Could you please file or point the PR to an issue describing the problem and what the proposed changes do (What's the visual before/after)

tstivers1990:
See #1866

src/UI/Areas/Identity/Pages/Account/Manage/ManageNavPages.cs
```diff
@@ -12,10 +12,6 @@ public static class ManageNavPages
 
         public static string ChangePassword => "ChangePassword";
 
-        public static string DownloadPersonalData => "DownloadPersonalData";
```
javiercn: The changes look good. The only thing  we need to change are the removed members and the visibility changes. Is there any issue if you remove those changes? Those are breaking changes and we would ideally like to take this into one of our patch releases, so we need to revert those changes even if the methods are not used.

tstivers1990: @javiercn Yeah, my thought process at the time was that these changes shouldn't break anything, as it's scaffolded code. But now that you mention it, if someone were to have scaffolded one of the pages that was changed, and then updates Identity UI with these changes, it's likely to break something.

I'm not sure how you want me to go about this. The breaking changes are in two commits, so they can easily be reverted. Should the breaking changes be reverted and put into a separate pull request?

javiercn: @tstivers1990 I think scraping them all together is fine. While it's not super clean, its in a .Internal namespace so it doesn't really matter. I don't want to trouble you into sending another PR as there's not much value on fixing it. Scraping the commits with the breaking changes will work fine.

Thanks for your contribution

tstivers1990:
The breaking changes should be removed.

javiercn:
Merged! Thanks for your contribution!

PR #1899, opened 2018-07-30
Fix provider display name

### LINKED ISSUE #1892: ExternalLoginInfo.ProviderDisplayName doesn't use the authentication scheme display name
I use ASP.NET Identity with Azure AD as an external identity provider. I configured Azure AD with a display name:

```csharp
services.AddAuthentication()
    .AddOpenIdConnect("AzureAD", "Azure Active Directory", options =>
            {
                ...
            });
```

When I call `SignInManager.GetExternalAuthenticationSchemesAsync()`, I would expect `ProviderDisplayName` in the result to be "Azure Active Directory", but it's "AzureAD" instead.

thomaslevesque:
Fixes #1892

### DIFF
--- a/test/Identity.FunctionalTests/ManagementTests.cs
+++ b/test/Identity.FunctionalTests/ManagementTests.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Net;
 using System.Security.Claims;
 using System.Threading.Tasks;
+using AngleSharp.Dom.Html;
 using Identity.DefaultUI.WebSite;
 using Microsoft.AspNetCore.Identity.EntityFrameworkCore;
 using Microsoft.AspNetCore.TestHost;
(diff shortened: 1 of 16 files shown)


HaoK:
Looks good overall, do you want to try and take a stab at adding a functional test for the UI changes?

Would probably go in here https://github.com/aspnet/Identity/blob/master/test/Identity.FunctionalTests/ManagementTests.cs#L169 with a slightly different flow that just verifies that the Display name is different, you would have to change the display name as well for Contoso here:

https://github.com/aspnet/Identity/blob/master/test/WebSites/Identity.DefaultUI.WebSite/Services/ContosoAuthenticationConstants.cs#L9

thomaslevesque:
@HaoK sure, I'll give it a try.

thomaslevesque:
@HaoK I tried something, but I'm not very happy with it. It seems a bit brittle to examine the HTML like this. Is this the kind of test you were expecting?

test/Identity.FunctionalTests/ManagementTests.cs
```diff
@@ -193,6 +194,31 @@ public async Task CanRemoveExternalLogin()
...
+            var email = $"{userName}@example.com";
+            var index = await UserStories.RegisterNewUserWithSocialLoginAsync(client, userName, email);
+            var manage = await index.ClickManageLinkWithExternalLoginAsync();
+            var externalLogins = await manage.ClickExternalLoginsAsync();
+
+            // Assert
+            var title = externalLogins.Document.GetElementsByTagName("h4").FirstOrDefault(e => e.TextContent == "Registered Logins");
+            var table = title?.NextElementSibling as IHtmlTableElement;
+            var firstCell = table?.Bodies?.FirstOrDefault()?.Rows.FirstOrDefault()?.Cells?.FirstOrDefault();
+            Assert.Equal("Contoso auth", firstCell?.TextContent);
```
HaoK: @javiercn is this the recommended way to verify this change in the UI tests?

HaoK:
Thanks merged via 6b257f0abbbe6938c0712b9a8e00b85131ed51d5

thomaslevesque:
Thanks @HaoK

test/Identity.FunctionalTests/ManagementTests.cs
javiercn: Good enough

PR #1914, opened 2018-08-07
Update Login.cshtml.cs

Rick-Anderson:


HaoK:
@blowdart this is actually a (maybe accidental change to the UI) which enabled password failures locking out normal lockouts by default

HaoK:
Replaced by https://github.com/aspnet/Identity/pull/2007

PR #1921, opened 2018-08-16
Merge release 2.2 -> master

HaoK:
Manual merge replacing automated PR

cc @natemcmaster


HaoK:
Yeah those failures are not new, vsts is the EF time out issue.  I'm not sure exactly why travis is failing in one instance, but it was failing in the other PRs as well.

natemcmaster:
We should really get to the bottom of this. From @smitpatel's last comment on https://github.com/aspnet/EntityFrameworkCore/issues/12842, it appears that EF's async tests can deadlock. Is Identity doing a bunch of async queries in EF?

HaoK:
Yeah tons of async queries

Issue #1926, opened 2018-08-17
Test failure: CannotLogInWithoutRequiredEmailConfirmation_WithGlobalAuthorizeFilter


ryanbrandenburg:
This test [fails](http://aspnetci/viewLog.html?buildId=533918&buildTypeId=XPlat_HighSierraOsxUniverse) occasionally with the following error:
```
System.OutOfMemoryException : Exception of type 'System.OutOfMemoryException' was thrown.
   at System.Threading.Thread.StartInternal()
   at System.Threading.Thread.Start(Object parameter)
   at System.IO.FileSystemWatcher.RunningInstance.Start()
   at System.IO.FileSystemWatcher.StartRaisingEvents()
   at System.IO.FileSystemWatcher.StartRaisingEventsIfNotDisposed()
   at System.IO.FileSystemWatcher.set_EnableRaisingEvents(Boolean value)
   at Microsoft.Extensions.FileProviders.Physical.PhysicalFilesWatcher.TryEnableFileSystemWatcher()
   at Microsoft.Extensions.FileProviders.Physical.PhysicalFilesWatcher.CreateFileChangeToken(String filter)
   at Microsoft.AspNetCore.Mvc.TagHelpers.Internal.FileVersionProvider.AddFileVersionToPath(String path)
   at Microsoft.AspNetCore.Mvc.TagHelpers.ScriptTagHelper.Process(TagHelperContext context, TagHelperOutput output)
   at Microsoft.AspNetCore.Razor.TagHelpers.TagHelper.ProcessAsync(TagHelperContext context, TagHelperOutput output)
   at Microsoft.AspNetCore.Razor.Runtime.TagHelpers.TagHelperRunner.RunAsync(TagHelperExecutionContext executionContext)
   at Identity.DefaultUI.WebSite.Pages.Pages__Layout.<ExecuteAsync>b__46_14()
   at Microsoft.AspNetCore.Razor.Runtime.TagHelpers.TagHelperExecutionContext.SetOutputContentAsync()
   at Identity.DefaultUI.WebSite.Pages.Pages__Layout.<ExecuteAsync>b__46_1()
   at Microsoft.AspNetCore.Razor.Runtime.TagHelpers.TagHelperExecutionContext.SetOutputContentAsync()
   at Identity.DefaultUI.WebSite.Pages.Pages__Layout.ExecuteAsync()
[...]

PR #1944, opened 2018-08-27
Disable parallelization

HaoK:
For https://github.com/aspnet/Identity/issues/1926

Test running time go from 17s -> 29s locally

@Eilon @ajcvickers

### DIFF
--- a/test/Identity.FunctionalTests/Properties/AssemblyInfo.cs
+++ b/test/Identity.FunctionalTests/Properties/AssemblyInfo.cs
@@ -1,0 +1,7 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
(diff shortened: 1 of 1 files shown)


test/Identity.FunctionalTests/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using Microsoft.AspNetCore.Testing.xunit;
+using Xunit;
+
+[assembly: CollectionBehavior(DisableTestParallelization = true)]
```
Eilon: Actually, can we add a comment here that explains why we did it, and links to the test failure issue?

HaoK: Updated with link and some words

PR #1966, opened 2018-09-14
Titles in h1

ryanbrandenburg:
Part of aspnet/templating#67.


ryanbrandenburg:
@javiercn the failure doesn't look related to my changes at first glance, are there preexisting conditions in Identity that would cause this kind of behavior?

HaoK:
@ryanbrandenburg yeah the VSTS builds for identity aren't working yet, there's another bug tracking it, appveyor is green so you should be good

ryanbrandenburg:
Need to have @danroth27 check out how this looks in the browser and approve.

PR #1976, opened 2018-09-22
[Design] Add UI user factory

HaoK:
Exploring fixes for https://github.com/aspnet/Identity/issues/1722

Not sure if this might be more appropriate for 3.0 as its will be hard to do this in a non breaking change way in 2.2, basically moves away from Activator.CreateInstance infavor of a dedicated DefaultUIUserFactory which will be used to create new user instances. 

We provide a default one that works for identity user based classes that have a default constructor, custom users can add their own factory implementation that implements:
 `TUser Create(string userName, string email)`

or they can just register it inline like so with a sugar method:
```
            services.AddDefaultIdentity<PocoUser>()
                .AddUserManager<UserManager<PocoUser>>()
                .AddDefaultUIUserFactory<PocoUser>((userName,email) =>
                    new PocoUser()
                    {
                        UserName = userName,
                        Email = email
                    });
```

Thoughts @ajcvickers @javiercn @blowdart

### DIFF
--- a/src/UI/Areas/Identity/Pages/V4/Account/Register.cshtml.cs
+++ b/src/UI/Areas/Identity/Pages/V4/Account/Register.cshtml.cs
@@ -52,22 +52,20 @@ namespace Microsoft.AspNetCore.Identity.UI.V4.Pages.Account.Internal
     {
         private readonly SignInManager<TUser> _signInManager;
         private readonly UserManager<TUser> _userManager;
-        private readonly IUserStore<TUser> _userStore;
-        private readonly IUserEmailStore<TUser> _emailStore;
+        private readonly IdentityDefaultUIUserFactory<TUser> _userFactory;
         private readonly ILogger<LoginModel> _logger;
         private readonly IEmailSender _emailSender;
 
         public RegisterModel(
             UserManager<TUser> userManager,
-            IUserStore<TUser> userStore,
             SignInManager<TUser> signInManager,
+            IdentityDefaultUIUserFactory<TUser> userFactory,
             ILogger<LoginModel> logger,
             IEmailSender emailSender)
         {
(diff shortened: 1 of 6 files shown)


src/UI/Areas/Identity/Pages/V4/Account/Register.cshtml.cs
```diff
@@ -82,10 +80,8 @@ public override async Task<IActionResult> OnPostAsync(string returnUrl = null)
             returnUrl = returnUrl ?? Url.Content("~/");
             if (ModelState.IsValid)
             {
-                var user = CreateUser();
-
-                await _userStore.SetUserNameAsync(user, Input.Email, CancellationToken.None);
```
DamianEdwards: Is the idea that any custom data that a custom `IUserStore` is pushing onto a a `TUser` instance needs to be explicitly added to the app's pages? i.e. we can't make it "automatic"?

HaoK: Yeah basically its possible they would want to set arbitrary data from the register page into the initial user before creating it.

Age, Gender, SSN, etc...  

The factory in this PR is an abstract class instead of an interface so we can add something like
`Create(IDictionary<string, string> registerData)` in the future if we wanted without a breaking change

HaoK:
Per triage we will revisit this in for 3.0

gulbanana:
I’ll look forward to it. At the moment our TUser is really a “user accessor” which has to be contextualised to retrieve the real orm object, which is fine in IUserStore but kind of a pain for app code to do.

natemcmaster:
This repo is being archived. If this is still something you want to get merged, open a new PR against https://github.com/aspnet/AspNetCore

PR #1980, opened 2018-09-27
 Enable MicroBuild code signing

natemcmaster:
Prepare for an update to buildtools. https://github.com/aspnet/BuildTools/pull/766


src/UI/Microsoft.AspNetCore.Identity.UI.csproj
```diff
@@ -20,7 +20,7 @@
     <EmbeddedResource Remove="wwwroot/**/LICENSE*" />
     <None Remove="wwwroot/**/.bower.json" />
     <None Remove="wwwroot/**/LICENSE*" />
-    <None Include="THIRD-PARTY-NOTICES" Pack="true" PackagePath="/" />
+    <None Include="THIRD-PARTY-NOTICES.txt" Pack="true" PackagePath="/" />
```
natemcmaster: There is some kind of weird bug in NuGet that causes malformed zip entries for extension-less files in the package root. It causes issues for System.IO.Packaging, but not for most other zip implementations including System.IO.Compression. So, (shrug) add a .txt?

HaoK: Weird yeah seems fine

PR #1990, opened 2018-10-03
Add missing "private readonly" modifiers to _userManager and _logger fields

miroslavpopovic:
Two small fixes for missing "private readonly" modifiers in Identity UI.


javiercn:
This needs to target the release/2.2 branch I believe

HaoK:

Issue #1868, opened 2018-07-09
Why are the `DbSet` in `IdentityUserContext` non-virtual?


benm-eras:
In ASP.NET Core 2.x the `DbSet`s defined in the abstract classes `IdentityDbContext` and `IdentityUserContext` are all non-virtual making it much harder to effectively mock my database context that is built on them. It looks like they were virtual in ASP.NET Core 1.x, although the library was arranged differently then, so I am not 100% sure on that. 

Is there a reason that they cannot just be made virtual to facilitate unit testing? I really don't like having to resort to something drastic like Fakes when I could just use Moq etc.

If there is a reason they can't be made virtual I would really appreciate it if any one has any ideas on how  I can proceed with mocking my db context that implements `IdentityDbContext`!

HaoK:
I don't believe there's anything preventing us from making them virtual, @ajcvickers can confirm.

This is something cheap we can do for 2.2

benm-eras:
Thanks @HaoK! In case anyone else is interested I found a work around that feels a little hacky but works fine. Simply declare a new `DbSet` foreach of the identity classes in your context along side any others you may have:

```csharp
public class MyContext 
    : IdentityDbContext<User, Role, int, UserClaim, UserRole, UserLogin, RoleClaim, UserToken>
{
    // Identity entities
    public new virtual DbSet<User> Users { get; set; }
    public new virtual DbSet<Role> Roles { get; set; }
    public new virtual DbSet<RoleClaim> RoleClaims { get; set; }
    public new virtual DbSet<UserClaim> UserClaims { get; set; }
    public new virtual DbSet<UserLogin> UserLogins { get; set; }
    public new virtual DbSet<UserRole> UserRoles { get; set; }
    public new virtual DbSet<UserToken> UserTokens { get; set; }

    // Your other entities
    public virtual DbSet<Foo> Foos { get; set; }
    public virtual DbSet<Bar> Bars{ get; set; }
}
```

HaoK:
Yep that should work in the mean time, glad you aren't blocked!

ajcvickers:
No real reason not to do this, but keep in mind that mocking DbSet and in particular IQueryable can be painful. We usually recommend creating a repository abstraction if you are going to create test doubles.

benm-eras:
Just goes to show how little us end users know about everything that goes in making the tools we use! Just glad I was able to work around it! Thanks for letting me know @HoaK

PR #1991, opened 2018-10-04
Make DbSet's virtual (#1868)

S3bt3r:
This PR sets the virtual modifier for the DbSets in IdentityDbContext and IdentityUserContext.
Issue reference: #1868


ajcvickers (approved):
@HaoK This looks good given that it is going into master for 3.0. However, it should be failing the API baseline check given that this is a breaking change. Can you look into why the baseline check is not failing on AppVeyor?

HaoK:

Issue #1891, opened 2018-07-25
Why isn't MemoryStream in DataProtectionTokenProvider disposed?


Ztarbox:
This question is a duplicate from https://github.com/aspnet/AspNetIdentity/issues/7

But since core has more contributors, I reissue it over here:
The `MemoryStream `in `GenerateAsync `and `ValidateAsync `are both not disposed.
https://github.com/aspnet/Identity/blob/c7276ce2f76312ddd7fccad6e399da96b9f6fae1/src/Identity/DataProtectionTokenProvider.cs#L72
Also the disposing by the `BinaryReader `and `BinaryWriter `is "disabled" through setting the `leaveOpen `parameter to `true`.
https://github.com/aspnet/Identity/blob/c7276ce2f76312ddd7fccad6e399da96b9f6fae1/src/Identity/DataProtectionTokenProvider.cs#L174

Shouldn't the `MemoryStream `be disposed, cause we don't need him anymore?

blowdart:
Feel free to submit a PR, we're not generally worried about MemoryStreams as it's not holding onto native resources which dispose would get rid of sooner.

zachesposito:
Hello, I made a PR to dispose the streams so that this issue can get closed out.

PR #2001, opened 2018-10-11
Dispose MemoryStream in DataProtectionTokenProvider for consistency

zachesposito:
Addresses #1891


HaoK:

PR #2023, opened 2018-10-25
[Fixes #2002] UI package has path error

### LINKED ISSUE #2002: UI package has path error
I was trying to import the Microsoft.AspNetCore.Identity.UI NuGet package into a local package server (using NuGet.Server).  This one package refused to import.  After opening an issue with the NuGet team (https://github.com/NuGet/NuGetGallery/issues/6543) the problem was traced to a bad path in the embedded nuspec file.

```
<Override ContentType="application/octet" PartName="//THIRD-PARTY-NOTICES"/>
```

The server did not like the double slash in front of the name.

It appears this path is coming from the csproj file (https://github.com/aspnet/Identity/blob/master/src/UI/Microsoft.AspNetCore.Identity.UI.csproj).

```
<None Include="THIRD-PARTY-NOTICES.txt" Pack="true" PackagePath="/" />
```

My guess is the "PackagePath" shouldn't start with "/" or it cannot be just a single "/".

javiercn:


javiercn:
Wrong branch

PR #2032, opened 2018-10-27
Update README.md

vyaspranjal33:
removed text from badge


natemcmaster:
This repo is being archived and will be made read-only. I'm removing these badges entirely in https://github.com/aspnet/Identity/pull/2092

PR #2035, opened 2018-10-29
Block enabled 2fa in the UI without cookie consent

HaoK:


HaoK:
I tried out the code via scaffolding the fix into a 2.1 app and it seems fine:

image

This broke a bunch of functional tests so we actually do have coverage, the tests now have some minimal coverage for enabling 2fa behavior with and without cookie consent now.

@Eilon @blowdart @ajcvickers

blowdart:
What happens if the dev removed all the consent code?

HaoK:
They would need to remove this as well I assume?    Or does CanTrack cover that case?

blowdart:
I don't know, that's what I'm asking. I would have hoped the check defaulted to true if there was no consent feature, but ...

HaoK:
Are you telling me CanTrack returns false by default?  @Tratcher any insight into the behavior?

Tratcher:
"Indicates either if consent has been given or if consent is not required." _should_ be true by default if you removed the features.

src/UI/Areas/Identity/Pages/V3/Account/Manage/TwoFactorAuthentication.cshtml
```diff
@@ -7,51 +8,62 @@
 
 <partial name="_StatusMessage" model="Model.StatusMessage" />
 <h4>@ViewData["Title"]</h4>
-@if (Model.Is2faEnabled)
+@if (HttpContext.Features.Get<ITrackingConsentFeature>().CanTrack)
```
Tratcher: Null ref?

src/UI/Areas/Identity/Pages/V3/Account/Manage/TwoFactorAuthentication.cshtml
```diff
@@ -7,51 +8,63 @@
...
+        <a id="enable-authenticator" asp-page="./EnableAuthenticator" class="btn btn-default">Add authenticator app</a>
     }
-
-    if (Model.IsMachineRemembered)
+    else
     {
-        <form method="post" style="display: inline-block">
-            <button type="submit" class="btn btn-default">Forget this browser</button>
-        </form>
+        <a id="enable-authenticator" asp-page="./EnableAuthenticator" class="btn btn-default">Setup authenticator app</a>
```
Eilon: Since we're changing things, can you change this to `Set up` (verb)?

PR #2060, opened 2018-11-08
Update Rfc6238AuthenticationService.cs

grahamehorner:
20bytes is 160bits


grahamehorner:
? changing a comment ? breaks build ?? hmm

PR #2088, opened 2018-11-30
Remove `SetCompatibiltyVersion(...)` call in sample

dougbu:
- aspnet/Mvc#7156 2 of ?


dougbu:
@rynowak this is really part 1: must go in before aspnet/Mvc#4328

Please have a look @ both PRs soon. More to come 😺

